PR #213, opened 2012-04-24
SERVER-5319 $strlen for aggregation framework

afchin:
initial implementation and tests


Timokasse:
So how do we find the length of a string in MongoDB? Is there a solution now?

PR #400, opened 2013-03-15
SERVER-8990 Cross-Cluster replication marker

vasmik:
Append a special marker to document to get an ability for cross-cluster replication using an op-log.

https://jira.mongodb.org/browse/SERVER-8990


benety:
Hi @vasmik,

Apologies for taking so long to get back to you.

We revisited this pull request recently and, on further discussion with @ramonfm and @scotthernandez, decided that we are unable to move forward with the patch as proposed. 

Please refer to Jan 17 comments in [SERVER-8990](https://jira.mongodb.org/browse/SERVER-8990) for a fuller explanation and an potential alternative solution that is currently implemented in MongoDB 2.6.

Ben

PR #416, opened 2013-04-23
SERVER-2114 Don't use select() timeouts for fast coarse timing

jcalvinowens:
https://jira.mongodb.org/browse/SERVER-2114

I have tested this on Linux. I do not have access to machines running 64-bit OSX or Solaris, and I don't have Visual Studio, so those platforms are not even compile tested.

### DIFF
--- a/src/mongo/util/net/listen.cpp
+++ b/src/mongo/util/net/listen.cpp
@@ -39,6 +39,7 @@
 #include <arpa/inet.h>
 #include <errno.h>
 #include <netdb.h>
+#include <time.h>
 #ifdef __openbsd__
 # include <sys/uio.h>
 #endif
@@ -49,6 +50,23 @@
 #undef errno
 #define errno WSAGetLastError()
 
+#endif
+
+// Make sure our Linux has the timer we need
+#if defined(__linux__)
+    #ifndef CLOCK_MONOTONIC_COARSE
+        #error "Your ancient kernel doesn't support CLOCK_MONOTONIC_COARSE: upgrade to v2.6.31 or newer"
+    #endif
+#endif
+
+// For absolute_time monotonic counters
+#if defined(__MACH__)
+    #include <mach/mach_time.h>
+#endif
+
+// For monotonic performance counters
+#if defined(_WIN32)
+    #include <windows.h>
 #endif
 
 namespace mongo {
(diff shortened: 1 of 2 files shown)


src/mongo/util/net/listen.cpp
```diff
@@ -51,6 +52,23 @@
 
 #endif
 
+// Make sure our Linux has the timer we need
+#if defined(__linux__)
+    #ifndef CLOCK_MONOTONIC_COARSE
+        #error "Your ancient kernel doesn't support CLOCK_MONOTONIC_COARSE: upgrade to v2.6.31 or newer"
```
benmccann: Now that it's almost two years later, I'm wondering are you still targeting the same kernel versions? I'd love to see this PR make it's way in so that Mongo stops killing my laptop battery

benmccann:
@jcalvinowensthere's a merge conflict. perhaps you could update this PR?

src/mongo/util/net/listen.cpp
benmccann: @erh what kernel versions need to be supported today? I'd like to try to resurrect this PR, but need to know what to support

erh: We still support RHEL 5, which is still at 2.6.18, but with a lot of
patches.

benmccann: Thanks @erh! Really appreciate it. It's very helpful to know what we're targeting.

@jcalvinowens it looks like since we have to target 2.6.18 that CLOCK_PROCESS_CPUTIME_ID, CLOCK_THREAD_CPUTIME_ID, CLOCK_REALTIME, and CLOCK_MONOTONIC are available according to [the man page](http://man7.org/linux/man-pages/man2/clock_gettime.2.html). Is knowing that we have to target 2.6.18 enough to unblock you? There's a chance I might be able to look at this, but I'm very unexperienced with it, so if you would be able to comment here as to an approach that you think would work, that would be really helpful.

benmccann: @jcalvinowens any pointers? i'd love to help get this in

benmccann:
@jcalvinowens think you'll be able to update this PR?

monkey101:
this was closed by accident, but the source repo appears to have been deleted, so can't be reopened.

PR #486, opened 2013-08-30
SERVER-10651 - MySQL's DATE_ADD DATE_SUB like functions for $project

dribba:
It would be nice to count with something like DATE_ADD and DATE_SUB of MySQL where you can add or subtract intervals, not only milliseconds like the $add expression is doing now.
Right now all the add/subtract operations with date are handled like numbers, and some cases this is not enough, like adding months or years.
I propose to create new expressions to handle date addition/subtraction supporting units and also milliseconds. Like:
db.test.aggregate( { $project: { expiryDate: { $dateAdd: [ "$date", [5, "months"] ] } } });
db.test.aggregate( { $project: { lastYear: { $dateSub: [ "$date", [1, "years"] ] } } });

The code might need revision, we tried to complete all the tests(C++ and JS)


arg20:
This was closed but never again revisited. It's important for reporting purposes using the aggregation framework. Say you want to aggregate by month and need to have a `from` & `to` date fields. you can't just add 30 days to the `from` date because months have varying numbers of days.

PR #580, opened 2013-12-12
SERVER-12128 Change Earth radius from 6371 to 6378.1 to reflect constant used by MongoDB

rps:
constant kRadiusOfEarthInMeters is defined as 6378.1 \* 1000 in mongo/src/mongo/db/geo/geoconstants.h. 

Changed tests that referenced Earth radius in km as 6371 to reflect the constant value.


benety:
Related to https://github.com/mongodb/docs/pull/1470

PR #588, opened 2013-12-14
SERVER-12128 Inconsistent Earth radius used in tests / documentation

rps:
MongoDB tests and documentation utilize mean Earth radius (6371km) rather than equatorial radius (6378.1km). Because Earth is not a perfect sphere, mean radius may be a more appropriate constant.

Ref: https://github.com/mongodb/mongo/pull/580#issuecomment-30535697

### DIFF
--- a/src/mongo/db/geo/geoconstants.h
+++ b/src/mongo/db/geo/geoconstants.h
@@ -31,6 +31,6 @@
 namespace mongo {
 
     // Thanks, Wikipedia.
-    const double kRadiusOfEarthInMeters = (6378.1 * 1000);
+    const double kRadiusOfEarthInMeters = (6371 * 1000);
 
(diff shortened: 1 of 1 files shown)


rps:
@benety The net effect of having documentation inconsistent with the constants used by Mongo is that end users will conduct inaccurate conversions from kilometers to radians. For geospatial applications reliant on Mongo, this results in inaccurate and incomplete proximity search. While I understand that changing the constant used by Mongo breaks existing indexes, I still don't understand why the documentation does not reflect the variable used.

benety:
@rps, I've filed a DOCS ticket to fix this in the documentation:

https://jira.mongodb.org/browse/DOCS-5038

benety:
Related to https://github.com/mongodb/docs/pull/1470

PR #622, opened 2014-01-31
SERVER-9580 Changed spinning code that waits for more data on tailable cursors for event based awakening

jsbattig:
Current implementation in tailable cursor imposes on the reader a wait of at least 2ms when cursor is exhausted even if new fresh data is inserted from another session.
When using MongoDB to fuel a service-bus implementation, this imposes a very expensive 2ms penalty on the majority of the requests going thru the bus.

A more scalable model implies using event notification while waiting for more data to avoid this unnecesary wait.

This wait has to be designed in a such a way to avoid imposing a too onerous cost on insertion of documents on collection object but yet achieving the goal of quick awakening of tailing cursors when waiting for more data.

On the collection side important to be noted that I have used an rwlock which is in the list of no-no for kernel, but I think it's justified in this case as it can allow concurrent sessions inserting documents to reduce contention when they notify subscribers.
An exclusive lock is acquired when new cursors register subscribers to wait for new documents when tailing.
There's an optimization on collection object to avoid creating the list<> that contains the subscribers unless it's necesary (first subscriber subscribing to collection).


benmccann:
@jsbattig looks like the PR needs to be rebased in order to be mergable

jsbattig:
Ben, I would love to do it, but we are buried with work at Convey so I
doubt I will be able to get to this potentially for a couple of weeks.

I do recommended the whole package is reviewed by your team, as I think
it's a critical enhancement to a big issue with tailable cursors.

jsbattig:
PR is mergable now

jsbattig:
@benety , I keep doing rebase and refreshing this PR with merge-able code? What's the plan with getting this issue addressed?
We can't use official build of MongoDB until this is fixed, and I imagine that if there's other people out there leveraging capped collection they probably have the issue we have.

jsbattig:
@benmccann @benety , what's the plan with the PR? I'm having trouble keeping the code mergeable, not to mention compiling.
My latest attempt to compile failed with some scons/msbuild issue, and I find it hard to keep up to date with compiler changes and so on on the project...
It will be ideal if this is moved forward, tested and made part of the standard Mongo build for everybody to enjoy.

PR #782, opened 2014-09-19
SERVER-15322 mongosniff is erroneously displaying the same string for both source and destination IP addresses

89berner:
I used to have a lot of this type of messages:

10.32.164.103:6612  <<--  10.32.164.103:49383   85 bytes  id:10ad72f    17487663 - 38764427
10.32.164.94:49383  -->> 10.32.164.94:6612 admin.$cmd  58 bytes  id:24f7f8b     38764427

Where the srcip and dstip are the same, it's solved with this minor change.


benety:
Fixed in https://github.com/mongodb/mongo/commit/ab47afeb030b42097acede562760c3a2ca785086

PR #834, opened 2014-10-22
SERVER-15789 Add a new set of KVEngine implementation files to...

esmet:
...storage/kv/dictionary to give engine authors another option for
code reuse.

The KVEngineImpl uses two classes, KVRecordStore and KVSortedDataImpl,
to implement the standard KVEngine API while delegating actual storage
responsibily to a new virtual class, KVDictionary. This means a
minimal engine can be written by implementing just a dictionary and a
transaction API.

The main motivation here is that many storage engines expose a sorted KV
API. This means that most of the code to implement a record store vs an
index would end up being the same.

This commit includes:
- A test-only heap implementation of the new code framework
  (storage/kv_heap). This implementation passes core js tests (except
  those that perform a renameCollection, see below).
- Unit tests for the new code frame work hooked up to the record store,
  sorted data interface, and kv engine test harnesses.

Rename in the kv engine closes the source collection before re-opening
it with a new name, which means kv_heap will drop all of its data when
it gets renamed, so tests fail. This is not a terribly hard problem to
fix but it involves some change to KVEngine so we chose not to solve
that problem right away.


benety:
Hi @esmet,

Sorry for the delay in our response.Unfortunately, this doesn't rebase cleanly on master (c3da2fc6642f143111d104c871f420d523f949b5) anymore. We can discuss further in [SERVER-15789](https://jira.mongodb.org/browse/SERVER-15789) on the next steps.

Ben

PR #837, opened 2014-10-24
 SERVER-15793 Pass the old record to the storage engine's updateRecord function.

esmet:
Now, any updateRecord implementation that needs the old record
no longer needs to do a read-before-write. One possible implementation is
a RecordStore that maintains the data size of all records and needs to
know the size delta of every update.

The upstream rocks engine will immediately benefit from this. I have not included any changes to rocks, however, as that is outside of my realm.

https://jira.mongodb.org/browse/SERVER-15793

All unit tests pass and so do the core js tests that start with 'update' (ie: jstests/core/update*'

PR #838, opened 2014-10-24
SERVER-15794 Optionally pass the record to-be-deleted to deleteRecord.

esmet:
The upstream rocks engine will immediately benefit from this. I have not included any changes to rocks, however, as that is outside of my realm.

https://jira.mongodb.org/browse/SERVER-15794


benmccann:
@esmet there's a merge conflict on this. perhaps you could update it?

benety:
Hi @esmet,

Sorry for the delay in getting back to you. Unfortunately for reasons similar to #837, we have to close this request because the API has changed quite a bit since you submitted this patch.

Ben

PR #894, opened 2015-01-01
SERVER-14992 No need to zero-out data files on Windows 7/2008 R2 if ntfs.sys has hotfix 2731284

agherardi:
See https://jira.mongodb.org/browse/SERVER-14992


src/mongo/util/processinfo_win32.cpp
```diff
@@ -116,6 +117,73 @@ namespace mongo {
...
+
+        boost::scoped_array<char> verData(new char[verSize]);
+        if (GetFileVersionInfoA(filePath, NULL, verSize, verData.get()) == 0) {
+            DWORD gle = GetLastError();
+            warning() << "GetFileVersionInfoSizeA on " << filePath << " failed with " << errnoWithDescription(gle);
+            return false;
+        }
+
+        UINT size;
+        LPBYTE lpBuffer;
```
dgolub: [Nitpick] Why is this `LPBYTE` instead of `VS_FIXEDFILEINFO*`?

src/mongo/util/processinfo_win32.cpp
```diff
@@ -116,6 +117,73 @@ namespace mongo {
...
+        fileVersionLS = verInfo->dwFileVersionLS;
+        return true;
+    }
+
+    // If the version of the ntfs.sys driver shows that the KB2731284 hotfix or a later update
+    // is installed, zeroing out data files is unnecessary. The file version numbers used below
+    // are taken from the Hotfix File Information at http://support.microsoft.com/kb/2731284.
+    bool isKB2731284OrLaterUpdateInstalled() {
+        char systemRoot[MAX_PATH];
+        if (FAILED(SHGetSpecialFolderPathA(NULL, systemRoot, CSIDL_SYSTEM, FALSE))) {
```
dgolub: [Question] Any reason not to use `GetSystemDirectory` here?

src/mongo/util/processinfo_win32.cpp
```diff
@@ -116,6 +117,73 @@ namespace mongo {
...
+
+        string ntfsDotSysPath = systemRoot;
+        ntfsDotSysPath.append("\\drivers\\ntfs.sys");
+        DWORD fileVersionMS;
+        DWORD fileVersionLS;
+        if (getFileVersion(ntfsDotSysPath.c_str(), fileVersionMS, fileVersionLS)) {
+            WORD fileVersionFirstNumber = (fileVersionMS >> 16) & 0xffff;
+            WORD fileVersionSecondNumber = fileVersionMS & 0xffff;
+            WORD fileVersionThirdNumber = (fileVersionLS >> 16) & 0xffff;
+            WORD fileVersionFourthNumber = fileVersionLS & 0xffff;
```
dgolub: Should probably use `HIWORD` and `LOWORD` here.

src/mongo/util/processinfo_win32.cpp
```diff
@@ -116,6 +117,73 @@ namespace mongo {
...
+    bool isKB2731284OrLaterUpdateInstalled() {
+        char systemRoot[MAX_PATH];
+        if (FAILED(SHGetSpecialFolderPathA(NULL, systemRoot, CSIDL_SYSTEM, FALSE))) {
+            DWORD gle = GetLastError();
+            warning() << "SHGetSpecialFolderPathA on CSIDL_SYSTEM failed with " << errnoWithDescription(gle);
+            return false;
+        }
+
+        string ntfsDotSysPath = systemRoot;
+        ntfsDotSysPath.append("\\drivers\\ntfs.sys");
```
dgolub: Do we have a guarantee that the returned path will not have a trailing backslash?  If not, then we should check for it.

dgolub:
Are we checking to make sure that the volume where the database is stored uses NTFS?  This check if only meaningful when the database is on an NTFS volume.

dgolub:
Done reviewing.  Just a few questions and minor nitpicks.

agherardi:
>>>> Are we checking to make sure that the volume where the database is stored uses NTFS? This check if only meaningful when the database is on an NTFS volume.

The best place to check the volume type (NTFS vs. non-NTFS) is when we create a new datafile in file_allocator.cpp, We can't have a one-time check since different DBs could be stored on different volumes - using directoryperdb and symbolic links. Agreed?

In Windows 2008/Vista and above there is a GetVolumeInformationByHandleW API that does the job. But we cannot use that API in mongodb because we're compiling for earlier OSs. I can think of the following options:
1. Use GetVolumeInformation: The issue with this approach is that we would have to handle symbolic links, etc.
2.  Since KB2731284 only applies to Windows 7 and 2008 R2, we could try to get a hold of GetVolumeInformationByHandleW dynamically.
3. Do nothing. If a datafile is allocated on a FAT volume on a Windows 7/2008 R2 host and KB2731284 is not installed, mongoD will unnecessarily zero-out the datafile. This is not any better or worse than what happens today.

Thoughts?

src/mongo/util/processinfo_win32.cpp
agherardi: Added check.

agherardi:
I added a check for NTFS vs. non-NTFS in file_allocator.cpp.

src/mongo/util/processinfo_win32.cpp
```diff
@@ -116,6 +116,75 @@ namespace mongo {
...
+        fileVersionLS = verInfo->dwFileVersionLS;
+        return true;
+    }
+
+    // If the version of the ntfs.sys driver shows that the KB2731284 hotfix or a later update
+    // is installed, zeroing out data files is unnecessary. The file version numbers used below
+    // are taken from the Hotfix File Information at http://support.microsoft.com/kb/2731284.
+    bool isKB2731284OrLaterUpdateInstalled() {
+        char systemDirectory[MAX_PATH];
+        if (!GetSystemDirectoryA(systemDirectory, sizeof(systemDirectory))) {
```
dgolub: If you check the documentation for `GetSystemDirectory`, you'll see that the return value is not a Boolean but rather an integer.  If it fails due to the buffer not being large enough, it will return the required buffer size.  Therefore, a nonzero return value does not necessarily mean success.  I recognize it's unlikely that the system directory path will exceed `MAX_PATH` characters, but we should still check this return value properly.

dgolub:
Just one comment, otherwise the changes all look good.

agherardi:
Dynamically allocating buffer for system directory, and checking return value.

dgolub:
:+1:

markbenvenuto:
Thanks for the contribution!

PR #895, opened 2015-01-01
SERVER-16721 Prevent init.d script to fail when pidFilePath isn't configured

jtyr:
This commit sets default value of the `PIDFILEPATH` variable. The variable is not set if the config file is using YAML format. If the variable is not set, the `dirname` command fails which makes the whole init.d script to fail.

### DIFF
--- a/rpm/init.d-mongod
+++ b/rpm/init.d-mongod
@@ -21,7 +21,6 @@ SYSCONFIG="/etc/sysconfig/mongod"
 
 DBPATH=`awk -F'[:=]' -v IGNORECASE=1 '/^[[:blank:]]*dbpath[[:blank:]]*[:=][[:blank:]]*/{print $2}' "$CONFIGFILE" | tr -d '[:blank:]'`
 PIDFILEPATH=`awk -F'[:=]' -v IGNORECASE=1 '/^[[:blank:]]*pidfilepath[[:blank:]]*[:=][[:blank:]]*/{print $2}' "$CONFIGFILE" | tr -d '[:blank:]'`
-PIDDIR=`dirname $PIDFILEPATH`
 
 mongod=${MONGOD-/usr/bin/mongod}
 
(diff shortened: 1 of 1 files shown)


benety:
Perhaps we should be fixing the regular expression for PIDFILEPATH to handle the correct casing of the YAML option name?

http://docs.mongodb.org/manual/reference/configuration-options/#processManagement.pidFilePath

jtyr:
Well there's more than one way to skin a cat. My solution prevents the init.d script from failing and allows the user to specify the correct `PIDDIR`, `PIDFILEPATH` and `DBPATH` in the sysconfig file (`/etc/sysconfig/mongod`).

ehershey:
The init script as it is should work with yaml or ini config files. @jtyr - can you post a gist with your config file that the existing init script doesn't work with? Also, what specific OS version are you in? 

I also don't think it's a good idea to add a new location with a value for the pidfilepath in the init script, especially in a line meant to set PIDDIR instead of directly dealing with the file location itself.

jtyr:
You are right, the init.d script actually works with YAML format, but only if the `pidfilepath` is defined in the config file. My config file doesn't define it so the command ``PIDDIR=`dirname $PIDFILEPATH` fails as the `PIDFILEPATH` variable is empty. This is why I set the default value of this variable.

ehershey:
The general intention of the init script is that it work with the accompanying config file, or equivalent. You could say that the current init script requires the pid file path to be set in your config file. 

I think changing that requirement is reasonable, but it should be more explicit - what about just moving the PIDDIR line after we read in the sysconfig file?

jtyr:
That's actually a good idea, @ehershey. I have changed the init.d script accordingly in the last commit. Please review it.

ehershey:
That looks good to me! Before we merge it, can you sign the contributor agreement here? 

https://www.mongodb.com/legal/contributor-agreement

jtyr:
Done (Transaction ID: XIXN3BKX86M3I4D).

ehershey:
This looks good to me. Thanks @jtyr!

I created a JIRA ticket to go along with the fix - https://jira.mongodb.org/browse/SERVER-16721

benety:
Hi @jtyr,

This pull request has been merged in  https://github.com/mongodb/mongo/commit/f96a473723e8a478f7c3c1adf991ef795e9acebc. Sorry about the incorrect feedback on the casing in the regular expression.

Thank you for contributing to MongoDB!
Ben

benety:
Related to #723

PR #896, opened 2015-01-02
SERVER-16716 Removing host string arg length limitation for mongo shell

akira-kurogane:
A tiny patch to remove the apparently artificial 250 char limit for the --host argument used by the mongo shell command. Users with many-member replica sets and longer hostnames can currently exceed that length.


benety:
Hi @akira-kurogane,

Patch itself looks fine. Could you please create a SERVER ticket for this?

We would also need a similar patch for v8_db-3.25.cpp. You can verify the fix using this scone flag:

scons ... --js-engine=v8-3.25 ...

Thank you
Ben

benety:
https://jira.mongodb.org/browse/SERVER-16716

PR #898, opened 2015-01-05
SERVER-16716 Removing host string arg length limitation for mongo shell

akira-kurogane:
This patch removes the artificial 250 char limit for the --host argumber of the mongo shell command.


benety:
https://jira.mongodb.org/browse/SERVER-16716

benety:
Replaces #898

PR #900, opened 2015-01-05
SERVER-16718 remove LogIndentLevel from log.h/log.cpp

alabid:


benety:
https://jira.mongodb.org/browse/SERVER-16718

PR #901, opened 2015-01-08
SERVER-16723 ServerStatus shell helper needs to handle error cases

daveh86:


benety:
https://jira.mongodb.org/browse/SERVER-16723

PR #907, opened 2015-01-13
SERVER-15876 Added some collection-level lock checks

alabid:


alabid:
https://jira.mongodb.org/browse/SERVER-15876

PR #908, opened 2015-01-14
SERVER-16850 Remove support for touch command for RocksDB storage engine

alabid:


alabid:
https://jira.mongodb.org/browse/SERVER-16850

PR #910, opened 2015-01-15
SERVER-16648 Halve threadCount for group workloads to lower mem usage

alabid:


benety:
https://jira.mongodb.org/browse/SERVER-16648

PR #914, opened 2015-01-23
SERVER-16893 Disallow creation of v0 indexes with non-mmapv1 storage engines

alabid:
Also moved jstests/tool/dumprestore6.js to jstests/mmapv1/dumprestore6.js
since this JS test attempts to restore a v0 index


alabid:
https://jira.mongodb.org/browse/SERVER-16893

PR #922, opened 2015-02-09
SERVER-17216 2.6 mongostat should not crash against 3.0 mongod, and print warning if detected

mpobrien:


ramonfm:
Merged in https://github.com/mongodb/mongo/commit/c9756fd528cc009cfd83c5bcfe9e309344f5e781

PR #923, opened 2015-02-12
SERVER-17267  exclude wtstats template sources from changed path.

rueckstiess:
@michaelcahill or @keithbostic please review, is this the correct path to exclude?


keithbostic:
Yes, that's correct.

You can remove `--exclude '*/tools/3rdparty'` as that directory no longer exists (it's now `tools/wtperf_stats/3rdparty`). If you want to remove that directory to avoid NVD3 code, I'd suggest removing`tools/wtperf_stats` entirely, I think it makes more sense).

kangas:
I implemented Keith's suggestion and squashed my changes into Thomas' commit

PR #929, opened 2015-02-26
SERVER-11840 reduce lock contention in groupcommitwithlimitedlocks

fdoray:
Before this fix, groupCommitMutex was held during the calls to WRITETOJOURNAL and WRITETODATAFILES. That prevented writes to occur during these costly operations. With this fix, the mutex is released before these operations.

The groupCommitMutex lock was required to ensure that the commitNumber didn't change between the call to commitJob.commitingBegin() and the call to commitJob.committingNotifyCommitted(). With the addition of the durBuilder mutex, this condition is still respected.

The groupCommitMutex lock also ensured that there was only one thread using __theBuilder at the time. This condition is also still respected because of the new durBuilder mutex.

See more details about the impact of this fix in the bug report https://jira.mongodb.org/browse/SERVER-11840

### DIFF
--- a/src/mongo/db/dur.cpp
+++ b/src/mongo/db/dur.cpp
@@ -53,15 +53,17 @@
    mutexes:
 
      READLOCK dbMutex (big 'R')
+     LOCK durBuilder
      LOCK groupCommitMutex
        PREPLOGBUFFER()
      READLOCK mmmutex
        commitJob.reset()
+     UNLOCK groupCommitMutex
      UNLOCK dbMutex                      // now other threads can write
        WRITETOJOURNAL()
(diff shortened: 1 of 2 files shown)


kaloianm:
Hi Francois,

Thank you for your submission.

In 3.0 we significantly improved the concurrency control around MMAP V1. Journal writes (WRITETOJOURNAL and WRITETODATAFILES) now happen on a separate thread, outside of locks and don't interfere with ongoing commits.

This pull request is no longer applicable for 3.0 and it is too major of a change to be accepted for 2.6.

I am going to close it and resolve SERVER-11840.

Thanks again for your work.

-Kal.

PR #935, opened 2015-03-13
Update to modern style for README file

randomite:
Completely updated the styling for the README file. Please let me know if you would like anymore changes or improvements. 

New README:
 
new

****Old README: 

old


ramonfm:
Thanks for the pull request @randomite , and apologies for the long delay in getting back to you. While the top-level README looks "ugly" on github, its main goal is to be readable when read with cat, less, etc. Note also that we have plans to change the way README and distrc/README work in this development cycle, so unfortunately we're not accepting pull requests against these files right now.

Regards,
Ramón.

randomite:
No problem, thanks for getting back to me!

PR #937, opened 2015-03-16
SERVER-17596 mongo::OID::asTimeT() and mongo::OID::asDateT() should be const

mapx:
As `mongo::OID::asDateT()` calls `mongo::OID::asTimeT()`, which calls `OID::getTimestamp()`, which is a const member function, `mongo::OID::asDateT()` and `mongo::OID::asTimeT()` should also be marked const.

### DIFF
--- a/src/mongo/bson/oid.cpp
+++ b/src/mongo/bson/oid.cpp
@@ -157,7 +157,7 @@ namespace {
                     kInstanceUniqueSize + kIncrementSize);
     }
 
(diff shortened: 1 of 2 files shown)


amidvidy:
https://jira.mongodb.org/browse/SERVER-17596

amidvidy:
Hey Yuchen,

This PR has been merged in 63cdeb9c5620. Thanks so much for contributing to MongoDB!

Adam

PR #939, opened 2015-03-25
SERVER-16685 use buildInfo over 32bit check for WT

jameswahlin:
An earlier commit on this ticket added a storageEngine list to buildInfo. This commit changes a jstest that was previously using a "32 bit build" check for WT exclusion to now use buildInfo to confirm inclusion.

### DIFF
--- a/jstests/sharding/sharding_system_namespaces.js
+++ b/jstests/sharding/sharding_system_namespaces.js
@@ -17,9 +17,13 @@ var db = st.s.getDB("test");
 var coll = db.sharding_system_namespaces;
 
 // This test relies on the wiredTiger storage engine being compiled
-// into the server. This assumption does not hold on 32-bit platforms.
-// See SERVER-16660.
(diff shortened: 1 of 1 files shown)


milkie:
There are multiple tests that use the 32-bit check to avoid failures on WiredTiger.  Should we change them all?

jameswahlin:
@milkie - can you point me towards the other tests or provide guidance on how to find? This was the one that I found that was clearly using the 32bit buildInfo check as a proxy for WiredTiger inclusion.

milkie:
There are some tests that need wiredTiger actively running, but sharding_system_namespaces.js is not one of them.  This particular test is special in that it only needs wiredTiger support compiled in, to validate storage engine options.
This pull request is complete and there is nothing further to do with the other tests.  Thanks to James for clearing this up.

jameswahlin:
Closed as I now have permission to push directly.

PR #912, opened 2015-01-18
Server 15793 Storage engine's updateRecord function should take oldRec

gevorgvoskanyan:
This is a new pull request for https://jira.mongodb.org/browse/SERVER-15793. The original one - #837, was recently closed due to merge conflicts. This one is based on a newer commit by @esmet : https://github.com/Tokutek/tokumxse/commit/6ce943fc1bb7994d68524d4fb8956a7cd46290ad, and an top of it adds changes to RocksDB and WiredTiger storage engines to take advantage of the passed-in old record.
Many thanks to John Esmet who authored the bulk of the changes in this pull request!

PR #942, opened 2015-03-30
SERVER-15793 Storage engine's updateRecord function should take oldRec (03/15)

gevorgvoskanyan:
Renewed pull request for https://jira.mongodb.org/browse/SERVER-15793, based off #912.
Includes corresponding changes to the WiredTiger storage engine as well.
Many thanks to John Esmet who authored the bulk of the changes in this pull request!


gevorgvoskanyan:
@benety, in the previous PR you indicated your preference for @esmet to submit the bulk interface changes instead of me doing it, and then I can follow-up with my changes to the storage engines. With all due respect, I don't think that would be practical, for the following reasons:
1. The build (WT engine) will get broken after applying an interface-only patch, until mine gets in.
2. Over time I resolved many merge conflicts and compile errors on this. If John were to submit a PR, he would either need to duplicate that work or take it as-is from my branch (in which case I don't see the benefit such additional steps would add).

Since the beginning of my involvement John has given me the permission to prepare and submit a full PR for this, provided I base it from his branch, which I did. Thus I'd really like to avoid causing him the trouble of passing part of the work back to him, when originally he gave me the green light to go ahead and do it fully.

Would it be problematic to apply a combined patch like this? I may be missing the reason why separate PRs would be better - please let me know if that's the case.

monkey101:
Hi Gevorg, we've considered the patch and have ultimately decided to not go in this direction.  Please see the last comment in https://jira.mongodb.org/browse/SERVER-15793.

Thanks
Dan

PR #943, opened 2015-03-30
SERVER-17808 Ensure availability in initial_sync_unsupported_auth_schema.js

kkmongo:
Previously, the test could fail if the primary transitioned to a secondary before reInitiate finished executing.


ramonfm:
Merged.

PR #944, opened 2015-03-31
TOOLS-716: Mongodump: support --queryFile to read the query from file.

kivistein:


ramonfm:
Thanks for the pull request. As discussed in TOOLS-716, the old tools have been baselined and we have a new, completely re-written version of the tools – so unfortunately we can no longer accept any pull requests against the old tools.

Please tune to TOOLS-716 for updates on when this feature will be included in the new tools.

Regards,
Ramón.

PR #946, opened 2015-04-06
Uniform LFs instead of CR/LF mix

dandv:


ramonfm:
Thanks for the pull request @dandv . We have plans to change the way README and distrc/README work in this development cycle, so unfortunately we're not accepting pull requests against these files right now.

Regards,
Ramón.

PR #949, opened 2015-04-13
SERVER-17832 Fix Resource Leak on SSL Connect and Accept Exception

guoyr:


src/mongo/util/net/ssl_manager.cpp
```diff
@@ -831,39 +832,31 @@ namespace mongo {
     }
 
     SSLConnection* SSLManager::connect(Socket* socket) {
-        SSLConnection* sslConn = new SSLConnection(_clientContext, socket, NULL, 0);
-        ScopeGuard sslGuard = MakeGuard(::SSL_free, sslConn->ssl);
-        ScopeGuard bioGuard = MakeGuard(::BIO_free, sslConn->networkBIO);
+        std::unique_ptr<SSLConnection> sslConn = stdx::make_unique<SSLConnection>(_clientContext, socket, (const char*)NULL, 0);
```
acmorrow: The JIRA issue that you referenced indicates that this fix is intended for backport to pre-3.1 branches, where C++11 is not permitted.

guoyr: Understood. The consensus from the CR is that there needs to be another version for the backport. I've submitted another PR on the 3.0 branch which uses auto_ptr instead. https://github.com/mongodb/mongo/pull/950

PR #950, opened 2015-04-13
SERVER-17832 Fix Resource Leak on SSL Connect and Accept Exception (v2.6...

guoyr:
..., v3.0)

use auto_ptr for pre c++ 11 versions


agralius:
The only different to #949 is the change from unique_ptr to auto_ptr right?

guoyr:
@agralius yep. Also using `new` instead of `make_unique` to instantiate the pointer since there's no equivalent for auto_ptr.

PR #951, opened 2015-04-13
SERVER-17913 Log voters at the default loglevel

adkomyagin:


src/mongo/db/repl/elect_cmd_runner.cpp
```diff
@@ -114,7 +114,8 @@ namespace repl {
 
         if (response.isOK()) {
             BSONObj res = response.getValue().data;
-            LOG(1) << "elect res: " << res.toString();
+            LOG(0) << "received " << res["vote"] << " votes from " << request.target;
```
andy10gen: Use log(), not LOG(0).

PR #955, opened 2015-04-19
SERVER-17651: Moved generateCacheKey() to run inside getPlanCacheKey, removed the function inside init()

ParadoxMeltdown:
uses str::string's empty() for the initial check

### DIFF
--- a/src/mongo/db/query/canonical_query.cpp
+++ b/src/mongo/db/query/canonical_query.cpp
@@ -548,8 +548,6 @@ namespace mongo {
             return validStatus;
         }
 
-        this->generateCacheKey();
-
         // Validate the projection if there is one.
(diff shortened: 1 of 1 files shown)


ParadoxMeltdown:
https://jira.mongodb.org/browse/SERVER-17651

ParadoxMeltdown:
After SERVER-17659, this pull is now obsolete

ramonfm:
Thanks for closing the PR @ParadoxMeltdown -- you're right that [SERVER-17659] (https://jira.mongodb.org/browse/SERVER-17659) eliminates the need for this change.

Note also that the code you sent makes a logically const method have a non-const implementation behind the scenes, which may be problematic when it comes to thread safety.

PR #956, opened 2015-04-20
[SERVER-18135] Backport SERVER-15182 to 2.6 and fix mutexdebugger too

dannyfallon:
Clang 3.5+ has issues with the checks on `this`, throwing a
`'this' pointer cannot be null in well-defined C++ code` error.

I didn't write a test because it's just a backport of fixes that have
been in master/2.7 since September 2014.

### DIFF
--- a/src/mongo/client/dbclientcursor.cpp
+++ b/src/mongo/client/dbclientcursor.cpp
@@ -325,9 +325,6 @@ namespace mongo {
     }
 
     DBClientCursor::~DBClientCursor() {
-        if (!this)
-            return;
-
         DESTRUCTOR_GUARD (
 
         if ( cursorId && _ownCursor && ! inShutdown() ) {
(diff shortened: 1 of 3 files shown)


dannyfallon:
It's interesting to consider the two approaches that have been taken to resolve this error in the 2.7.7+ branches:

* In https://github.com/mongodb/mongo/commit/db59c175c736609970b52f91e30fc2277e22b25b (which I have backported here) the solution is to remove the tautological checks.
* In https://github.com/mongodb/mongo/commit/28331b4bedcfe300612d38ef2699eee25097c7ac the solution was to add a compiler instruction (`-Wtautological-undefined-compare`) to ignore the tautological checks which downgrades them to warnings.

ramonfm:
One should be able to downgrade these from errors to warnings and get a successful build by adding the --disable-warnings-as-errors flag to the SCons invocation. We added that flag exactly for this purpose: so that novel warnings in newer compilers don't prevent building older versions of the product.

The warnings should be safe to ignore, since what it is warning about is that the check may be elided by the compiler, which is the same change we actually made in the 2.7 development releases, as you noticed.

Given that this is not a regression, we are not inclined to make code changes on a stable branch when there is a simple workaround.

Regards,
Ramón.

PR #957, opened 2015-04-22
SERVER-18162 Make init.d script for suse work when /var/run/mongodb doesn't exist.

cesarizu:
When using tmpfs for `/var/run` or cleaning up that directory, mongod fails to start because of a missing `PIDDIR` variable in the init.d script. This pull request adds that variable to `rpm/init.d-mongod.suse
` the same way it's defined on `rpm/init.d-mongod`.


ehershey:
closed by 4b5008d8f72ab3cd023aba24ae017cc63db46e8e

PR #959, opened 2015-04-28
SERVER-18239 dumpauth.js should use correct db/collection names

mpobrien:


ramonfm:
PR was merged but not closed (see https://jira.mongodb.org/browse/SERVER-18239). Closing now.

PR #962, opened 2015-04-29
SERVER-18258 use httplib instead of urllib2 to download data

mpobrien:
I am not entirely sure why urllib2 is hanging, but using httplib instead seems to work reliably on a spawnhost.


ehershey:
Should we switch the other http call to httplib too so they're the same?

PR #967, opened 2015-05-06
SERVER-18344 send logs to logkeeper

mpobrien:


ehershey:
lgtm. If @tychoish is cool with it we should merge it.

tychoish:
lgtm

PR #969, opened 2015-05-27
SERVER-18007 support non-default constructible types in StatusWith

amidvidy:
using github PRs while internal CR is down.

- mongo::Status now has a special moved from Status state.
- mongo::StatusWith now holds a std::aligned_storage instead of a
concrete T
- some header hacks to be able to use invariant() from status_with.h


amidvidy:
withdrawn in favor of a simpler approach!

PR #972, opened 2015-06-04
SERVER-18794: Adding aggregation $objectToArray

jefflee1990710:


ramonfm:
Thanks for taking the time to submit a pull request @jefflee1990710. Looks like some other commits were added on top of your change, so I'd like to ask you to submit a new pull request only with your changes to avoid confusion.

Thanks,
Ramón.

PR #990, opened 2015-06-30
SERVER-17410: listCommands returns unordered list

upendrag:


src/mongo/db/dbcommands_generic.cpp
```diff
@@ -275,12 +275,18 @@ class ListCommandsCmd : public Command {
                      int,
                      string& errmsg,
                      BSONObjBuilder& result) {
+        // sort the commands before building the result BSON
+        std::map<string, Command*> sortedCommands;
+        for (CommandMap::const_iterator it = _commands->begin(); it != _commands->end(); ++it) {
+            sortedCommands.emplace(it->first, it->second);
+        }
```
acmorrow: std::map<K, V> and std::unordered_map<K, V> have the same value_type typedef. I wonder if you could just write this as:

```
std::map<CommandMap::key_type, CommandMap::value_type> sortedCommands(
    _commands->begin(), _commands->end());
```
Also, std::map here feels like overkill. Commands number in the hundreds, sorting a vector of that size is going to be very fast.
```
std::vector<CommandMap::value_type> commands(_commands.begin(), _commands.end());
std::sort(commands.begin(), commands.end(), []...);
```
I also notice that in the loop below you only use the first element of the pair once, to filter. In that case, do the filtering when adding to the vector, and only keep the command pointer, avoiding the string copies, then sort that with a lambda based on the name in the command.
```
std::vector<Command*> commands;
for (const auto& entry : _commands)
    if (entry.first == entry.second->name)
        commands.push_back(entry.second);
std::sort(commands.begin(), commands.end(), [](Command* lhs, Command* rhs) {
    return (lhs->name) < (rhs->name);
});
```

upendrag: Thank you for the comments.

I cannot use std::map<CommandMap::key_type, CommandMap::value_type> here because CommandMap::value_type itself is of type pair<string, Command*>.

I am not entirely convinced how std::vector based implementation will be faster than the std::map one. Because if you go step by step,
std::map takes a total of N*log(N) time (for insertion) whereas std::vector based implementation has a total complexity of N + N*log(N) (Insertion + Sorting). 
If you have based your arguement on the size of the input or on how these STL containers are implemented, could you please point me to some literature explaining these things. I am very much interested in knowing more about it.

Finally, I agree with you on performing filtering in the first iteration itself. I will go ahead and make that change.

acmorrow: Sorry, I mistyped that. It should be

```
std::map<CommandMap::key_type, CommandMap::mapped_type>
```
Re performance: this is somewhat academic, because the actual performance difference will be unobservable, and this is not exactly hot-path code.

However, vector is often faster than map in practice due to its low constant factors and better cache utilization, since it is a contiguous container rather than a node based container.

upendrag:
@acmorrow. I have updated the patch with the suggested modifications. Please note that I couldn't use range-based for loop to iterate through `CommandMap` because it looks like `CommandMap` which is of type `UnorderedFastKeyTable` has no support for range-based for loops.
Thank you for your time.

acmorrow:
LGTM

PR #991, opened 2015-06-30
SERVER-18955 Have sendNextBatch set the correct batch

lissom:
ShardedClientCursor::sendNextBatch should set the correct batch size prior to all calls to more() or next().  Currently, if the first call to more() for a getMore requires calling to the shard, batch size 0 is used if the prior iteration of the loop exited due to docCount reaching batchSize.


monkey101:
this was closed by accident

ramonfm:
Looks like [SERVER-18955](https://jira.mongodb.org/browse/SERVER-18955) was fixed a while back but we forgot to close the pull request, so I'm closing it now.

Cheers,
Ramón.

PR #1005, opened 2015-07-29
SERVER-19647: Generate SpiderMonkey config for OpenBSD

i80and:


benety:
https://jira.mongodb.org/browse/SERVER-19647

PR #1007, opened 2015-08-10
SERVER-14864: making logging better when using fallocate or sparse files

ankitm111:
Testing done: I created a mongodb instance with default file allocation as mmap v01 and checked the logs. The logs did not display that they were zeroing out my file as they were using fallocate for allocating the file.

### DIFF
--- a/src/mongo/db/storage/mmap_v1/file_allocator.cpp
+++ b/src/mongo/db/storage/mmap_v1/file_allocator.cpp
@@ -313,6 +313,7 @@ void FileAllocator::ensureLength(int fd, long size) {
             uassert(10443, errnoWithPrefix("FileAllocator: file write failed"), written > 0);
             left -= written;
         }
+		log() << "filling zeroes in file..." << endl;
     }
 }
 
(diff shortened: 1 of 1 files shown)


monkey101:
this was closed by accident

ramonfm:
@ankitm111, I've bumped [SERVER-14864](https://jira.mongodb.org/browse/SERVER-14864) so it gets reconsidered in the next round of planning. Thanks for your patience.

Cheers,
Ramón.

src/mongo/db/storage/mmap_v1/file_allocator.cpp
```diff
@@ -313,6 +313,7 @@ void FileAllocator::ensureLength(int fd, long size) {
             uassert(10443, errnoWithPrefix("FileAllocator: file write failed"), written > 0);
             left -= written;
         }
+		log() << "filling zeroes in file..." << endl;
```
devkev: 1. Please fix the leading whitespace (spaces only, 4 per indent).
2. Please move this to be immediately prior to where the zeroing starts.  Since the zeroing may take some time, it is important to provide feedback in the log prior to when it starts (ie. line 300).  There is no need to log when it finishes, since that's done anyway on line 433.

devkev:
Hi @ankitm111 

Thanks for submitting this pull request.

In addition to the inline comments, could you please add a log message at line 258 to explicitly log the fallocate() case?  The added log message should be similar to the one on line 250 (ie. "LOG(1)").

Best regards,
Kev

devkev:
This has now been resolved by https://github.com/mongodb/mongo/commit/2179beff8f248ce78d2ce5e7952c8c74691c8afc, which incorporates the above suggestions into the original changes.  Thank you again for raising this pull request.

PR #1025, opened 2015-09-21
SERVER-20247 include configsvr:true in default replica set config if …

EshaMaharishi:
…node started with --configsvr


andy10gen:
Merged manually via cherry-pick as https://github.com/mongodb/mongo/commit/88c34d5fe384f80508c3b489ced5fcf4ae914bd1

PR #1027, opened 2015-10-09
SERVER-8839 TimerStats and TimerHolder should work with microseconds.

salessandri:


src/mongo/db/stats/timer_stats_test.cpp
```diff
@@ -38,14 +38,32 @@ namespace {
...
+    );
 }
 
 TEST(TimerStatsTest, GetReportOneRecording) {
     TimerStats timerStats;
     Timer timer;
-    int millis = timerStats.record(timer);
-    ASSERT_EQUALS(BSON("num" << 1 << "totalMillis" << millis), timerStats.getReport());
+    auto micros = timerStats.record(timer);
+    auto microsAsLongLong = static_cast<long long>(micros.count());
```
scotthernandez: Please convert to a Microsecond duration before calling count in case we later change the return type of record to something other than microseconds. Best not to mix auto var + .count().

durationCount<Microseconds>(micros).count()

Also, durationCount in time_support is nice helper to get long long output.

PR #1030, opened 2015-10-21
fixed line break

meghangill:


ramonfm:
This needs proper name and email address. Working with @meghangill on that next.

PR #1041, opened 2015-10-29
SERVER-21209: Fix in configuration file

pkoro:
I noticed (CentOS 6) that the init script fails to parse the contents of the configuration file as expected resulting in failure to stop the service when asked. 

The error steps in from [this][l1] line in the init script:
```
PIDFILEPATH=`awk -F'[:=]' -v IGNORECASE=1 '/^[[:blank:]]*(processManagement\.)?pidfilepath[[:blank:]]*[:=][[:blank:]]*/{print $2}' "$CONFIGFILE" | tr -d "[:blank:]\"'"`
```

What happens is that the `PIDFILEPATH` variable gets populated with all contents following the variable `pidFilePath` in the configuration file with spaces omitted. Hence using the default configuration file the bash variable gets the value:

```
# echo $PIDFILEPATH
/var/run/mongodb/mongod.pid#locationofpidfile
```

I guess another way of fixing this is adding something like `..| awk -F# '{print $1}'` within the command but maybe it's better to have a cleaner configuration file. 

[l1]: https://github.com/mongodb/mongo/blob/master/rpm/init.d-mongod#L22

### DIFF
--- a/rpm/init.d-mongod
+++ b/rpm/init.d-mongod
@@ -19,7 +19,7 @@ CONFIGFILE="/etc/mongod.conf"
 OPTIONS=" -f $CONFIGFILE"
 SYSCONFIG="/etc/sysconfig/mongod"
 
-PIDFILEPATH=`awk -F'[:=]' -v IGNORECASE=1 '/^[[:blank:]]*(processManagement\.)?pidfilepath[[:blank:]]*[:=][[:blank:]]*/{print $2}' "$CONFIGFILE" | tr -d "[:blank:]\"'"`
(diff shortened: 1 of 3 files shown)


ramonfm:
Thanks for tracking this down and taking the time to send in a pull request @pkoro. Before we can consider your changes you'll need to sign the (contributor's agreement)[https://www.mongodb.com/legal/contributor-agreement].

pkoro:
Hi @ramonfm . I have signed the contributor's agreement as requested.

ramonfm:
Thanks for signing the agreement @pkoro. Since we keep track of MongoDB issues via JIRA I've created [SERVER-21209](https://jira.mongodb.org/browse/SERVER-21209) to move the discussion there and discuss a solution.

We discussed this issue internally, and since it's not illegal to have comments in lines with configuration data I think the second solution you propose is more desirable.

pkoro:
Hi @ramonfm . Please review.

ehershey:
lgtm

ramonfm:
Pull request merged -- thanks @pkoro.

PR #1042, opened 2015-10-29
SERVER-18957: Separate isAssociativeAndCommutative into two separate properties

salessandri:


src/mongo/db/pipeline/expression.h
```diff
@@ -443,13 +446,22 @@ class ExpressionFromAccumulator
         return accum.getValue(false);
     }
 
-    bool isAssociativeAndCommutative() const final {
-        // Return false if a single argument is given to avoid a single array argument being treated
-        // as an array instead of as a list of arguments.
+    bool isAssociative() const final {
+        // Return false if a single argument is given to avoid a single array argument being
+        // treated as an array instead of as a list of arguments.
```
cswanson310: I assume you didn't mean to change the comment wrapping. Let's keep it wrapped at 100 characters, please. Ditto lines 459-460.

src/mongo/db/pipeline/expression.h
```diff
@@ -506,9 +526,16 @@ class ExpressionAllElementsTrue final : public ExpressionFixedArity<ExpressionAl
 class ExpressionAnd final : public ExpressionVariadic<ExpressionAnd> {
 public:
     boost::intrusive_ptr<Expression> optimize() final;
+
```
cswanson310: Please leave non-inlined methods condensed. It's more consistent with the rest of the file.

src/mongo/dbtests/expressiontests.cpp
```diff
@@ -1446,260 +1446,6 @@ class AddToBsonArray {
 
 }  // namespace FieldPath
 
-
```
cswanson310: You can ignore this file! Sorry about that. We have a ticket open to delete this: https://jira.mongodb.org/browse/SERVER-21232

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -139,6 +144,239 @@ class ExpressionNaryTestOneArg : public ExpressionNaryTest {
...
+            expr->addOperand(Expression::parseOperand(element, vps));
+        }
+    }
+
+    intrusive_ptr<Testable> _notAssociativeNorCommutative;
+    intrusive_ptr<Testable> _associativeOnly;
+    intrusive_ptr<Testable> _associativeAndCommutative;
+};
+
+TEST_F(ExpressionNaryTest, AddOperand) {
```
cswanson310: Thank you for converting these to the new-style unit tests! While you're here, we're making an effort to improve the names of new tests to make it easier to understand what they're testing. For example, instead of AddOperand, maybe AddedOperandsShouldAppearInSerialization. This will make it easier to scan the file to find obvious testing gaps. If you could update the names to this new format, it would be much appreciated. When it's not possible (without an absurdly long name), consider splitting into two or more tests, or adding a descriptive comment, as some of the tests already have.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -139,6 +144,239 @@ class ExpressionNaryTestOneArg : public ExpressionNaryTest {
...
+    intrusive_ptr<Expression> optimized = _notAssociativeNorCommutative->optimize();
+    ASSERT(_notAssociativeNorCommutative != optimized);
+    ASSERT_EQUALS(BSON("$const" << BSON_ARRAY(1 << 2)), expressionToBson(optimized));
+}
+
+// Verify that the optimization of grouping constant and non-constant operands
+// and then applying the expression to the constant operands to reduce them to
+// one constant operand is only applied if the expression is associative and
+// commutative.
+TEST_F(ExpressionNaryTest, ConstantAndNonConstantGroupingOptimization) {
```
cswanson310: Please split this into three separate tests

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -139,6 +144,239 @@ class ExpressionNaryTestOneArg : public ExpressionNaryTest {
...
+        intrusive_ptr<Expression> optimized = _associativeAndCommutative->optimize();
+        ASSERT(_associativeAndCommutative == optimized);
+        assertContents(_associativeAndCommutative, BSON_ARRAY("$path" << BSON_ARRAY(55 << 66)));
+    }
+}
+
+// Verify that the flattening of the inner operands that are of the same type as the expression
+// being optimized is done correctly and only if the expression is associative and commutative
+// TODO: This actually can be done if the associativity property is true but not the sorting.
+TEST_F(ExpressionNaryTest, FlattenInnerOperandsOptimization) {
```
cswanson310: Also split this test

cswanson310:
Overall this looks very good! Let me know if I can clarify or help with anything else. Otherwise, I'd say go ahead and get started with the optimization part!

salessandri:
I added the optimizations for associative-only operators and add tests for multiple cases of this optimization.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,81 @@ const char* ExpressionLog10::getOpName() const {
 
 /* ------------------------ ExpressionNary ----------------------------- */
 
+/**
+ * \brief Optimize a general Nary expression.
```
cswanson310: We don't use Doxygen keywords like brief or return anywhere else in the codebase, please remove.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,81 @@ const char* ExpressionLog10::getOpName() const {
...
+        if (dynamic_cast<ExpressionConstant*>(operand.get())) {
+            ++constOperandCount;
         }
     }
-
-    // collapse all constant expressions (if any)
-    Value constValue;
-    if (!constExprs.empty()) {
-        vpOperand = constExprs;
+    if (constOperandCount == vpOperand.size()) {
```
cswanson310: Please re-add comment saying why we can return early here, and what we're returning.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,81 @@ const char* ExpressionLog10::getOpName() const {
...
+                continue;
+            }
+            ExpressionNary* nary = dynamic_cast<ExpressionNary*>(operand.get());
+            if (nary && str::equals(nary->getOpName(), getOpName())) {
+                vpOperand.erase(vpOperand.begin() + i);
+                vpOperand.insert(
+                    vpOperand.begin() + i, nary->vpOperand.begin(), nary->vpOperand.end());
+                continue;
+            }
+            if (!isCommutative()) {
```
cswanson310: Please add a comment describing what this block is doing, and maybe (briefly) why it can only done when the operator is commutative

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2365,6 +2373,7 @@ Value ExpressionPow::evaluateInternal(Variables* vars) const {
         // Array indices correspond to exponents 0 through 63. The values in each index are the min
         // and max bases, respectively, that can be raised to that exponent without overflowing a
         // 64-bit int. For max bases, this was computed by solving for b in
+        // b = (2^63-1)^(1/exp) for exp = [0, 63] and truncating b. To calculate min bases,
```
cswanson310: Looks like this was accidental duplication.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,81 @@ const char* ExpressionLog10::getOpName() const {
...
+        ExpressionVector optimizedOperands;
+        for (size_t i = 0; i < vpOperand.size();) {
+            intrusive_ptr<Expression> operand = vpOperand[i];
+            if (dynamic_cast<ExpressionConstant*>(operand.get())) {
+                constExpressions.push_back(operand);
+                ++i;
+                continue;
+            }
+            ExpressionNary* nary = dynamic_cast<ExpressionNary*>(operand.get());
+            if (nary && str::equals(nary->getOpName(), getOpName())) {
```
cswanson310: The check of `nary->isAssociative()` is still needed. It's subtle, but see [this commit](https://github.com/mongodb/mongo/commit/f9828f5a0312801fa5b8592b31648a64f0ea7f67) for why it's necessary, particularly [this test and comment](https://github.com/mongodb/mongo/commit/f9828f5a0312801fa5b8592b31648a64f0ea7f67#diff-22feeb9ade901bc4e660448dcbf9fe61R69).

Unfortunately, I don't think that test ever caught the bug it was intended to catch, since `optimize()` is called first, and the array of operands will be reduced to one constant. If you add a "$a" to the nested expression, it should not optimize to a constant, so the test should fail. Can you please update the test, make sure it fails for your current set of changes, then add back the `nary->isAssociative()` check to ensure it is fixed?

Let me know if that's unclear, as it is very subtle.

src/mongo/db/pipeline/expression.h
```diff
@@ -443,13 +446,22 @@ class ExpressionFromAccumulator
...
         // Return false if a single argument is given to avoid a single array argument being treated
         // as an array instead of as a list of arguments.
         if (this->vpOperand.size() == 1) {
             return false;
         }
-        return Accumulator().isAssociativeAndCommutative();
+        return Accumulator().isAssociative();
+    }
+
+    bool isCommutative() const final {
```
cswanson310: I think this one can always return `Accumulator().isCommutative()`. The operand size check was intended to prevent coalescing in the case described in my comment about adding the `nary->isAssociative()` check above. I don't see anything that can go wrong if `isCommutative()` returns true in that scenario (or any other scenarios).

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -139,6 +144,461 @@ class ExpressionNaryTestOneArg : public ExpressionNaryTest {
...
+TEST_F(ExpressionNaryTest, GroupingOptimizationOnCommutatievAndAssociative) {
+    BSONArray spec = BSON_ARRAY(55 << 66 << "$path");
+    addOperandArrayToExpr(_associativeAndCommutative, spec);
+    assertContents(_associativeAndCommutative, spec);
+    intrusive_ptr<Expression> optimized = _associativeAndCommutative->optimize();
+    ASSERT(_associativeAndCommutative == optimized);
+    assertContents(_associativeAndCommutative, BSON_ARRAY("$path" << BSON_ARRAY(55 << 66)));
+}
+
+TEST_F(ExpressionNaryTest, FlattenInnerOperandsOptimizationOnNotCommutativeNorAssociative) {
```
cswanson310: Can you please add a comment here that describes the structure you're creating? I found the test setup hard to read to figure out what the structure of the arguments were. It might also help here and elsewhere to change `arrBuilder` to be called `fullSpecBuilder` or something more specific. Please also give a comment to `FlattenInnerOperandsOptimizationOnCommutativeAndAssociative` on line 575 which has a similar setup.

cswanson310:
Overall this is looking good! There's a subtle bug I pointed out, but other than that my comments are all to improve readability.

src/mongo/db/pipeline/expression.cpp
salessandri: Got it. Fixed the js test, the implementation and added a unit test just in case.

src/mongo/db/pipeline/expression.h
salessandri: Fixed

src/mongo/db/pipeline/expression_test.cpp
salessandri: I replaced those tests, they were over-complex for what they were testing.

I also added comments to the tests to clarify what they are testing. If it's too verbose, let me know and I'll remove them.

I changed the `arrBuilder` variable name, too.

cswanson310:
Sorry it's taken me so long to get back to you! Was on vacation for a while during the holidays and got a bit busy when I got back. Will take a look today.

jstests/aggregation/bugs/server9625.js
```diff
@@ -67,7 +67,7 @@ load('jstests/aggregation/extras/utils.js');
     // If isAssociativeAndCommutative() did not return false when provided a single argument, the
     // single array argument provided to the inner expression would be ignored instead of treated
     // as a list of arguments, and these tests would fail.
-    testOp({$sum: ["$a", 2, 3, {$sum : [[4, 5]]}]}, 15);
-    testOp({$min: ["$a", 2, 3, {$min : [[4, 5]]}]}, 1);
-    testOp({$max: ["$a", 2, 3, {$max: [[4, 5]]}]}, 5);
+    testOp({$sum: ["$a", 2, 3, {$sum : [["$a", 4, 5]]}]}, 16);
+    testOp({$min: ["$b", 2, 3, {$min : [["$a", 4, 5]]}]}, 1);
+    testOp({$max: ["$a", 2, 3, {$max: [["$b",4, 5]]}]}, 6);
```
cswanson310: Please use consistent spacing here. To be specific, no space between the second `$sum` and the `:` on 70, same for `$min` on 71, and add a space after the `"$b,"` in the array on 72

jstests/aggregation/bugs/server9625.js
```diff
@@ -50,7 +50,7 @@ load('jstests/aggregation/extras/utils.js');
     // isAssociativeAndCommutative() is called. If all arguments are constant, the optimization
```
cswanson310: While you're in this file, can you update this comment to not refer to `isAssociativeAndCommutative()`, and instead to the two new methods? Also on line 67.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -139,6 +144,530 @@ class ExpressionNaryTestOneArg : public ExpressionNaryTest {
...
+    intrusive_ptr<Expression> optimized = _associativeOnly->optimize();
+    ASSERT(_associativeOnly == optimized);
+
+    BSONArray expectedContent = BSON_ARRAY(100 << "$path1" << BSON_ARRAY(101 << 200) << "$path2");
+    assertContents(_associativeOnly, expectedContent);
+}
+
+// Test that inner expressions are correctly flattened and constant operands re-arranged and
+// collapsed when using a commutative and associative expression.
+// {"$testable" : [ 200, "$path3", 201 { "$testable" : [ 100, "$path1", 101] }, 99, "$path2"] } is
```
cswanson310: Missing a comma after `201`

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,101 @@ const char* ExpressionLog10::getOpName() const {
...
+                    vpOperand = std::move(vpOperandSave);
+                } else {
+                    optimizedOperands.insert(
+                        optimizedOperands.end(), constExpressions.begin(), constExpressions.end());
+                }
+                constExpressions.clear();
+            }
+            optimizedOperands.push_back(operand);
+            ++i;
+        }
```
cswanson310: Can you add a blank line here to help legibility?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,101 @@ const char* ExpressionLog10::getOpName() const {
...
+        if (constExpressions.size() > 1) {
+            vpOperand = std::move(constExpressions);
+            Variables emptyVars;
+            optimizedOperands.emplace_back(
+                ExpressionConstant::create(evaluateInternal(&emptyVars)));
+        } else {
+            optimizedOperands.insert(
+                optimizedOperands.end(), constExpressions.begin(), constExpressions.end());
+        }
```
cswanson310: Also here

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2144,74 +2145,101 @@ const char* ExpressionLog10::getOpName() const {
...
-        vpOperand = constExprs;
-        Variables emptyVars;
-        constValue = evaluateInternal(&emptyVars);
-    }
+            // If the operand is exactly the same type as the one we are currently optimizing and
+            // is also associative, replace the expression for the operands it has.
+            // E.g: sum(a, b, sum(c, d), e) => sum(a, b, c, d, e)
+            ExpressionNary* nary = dynamic_cast<ExpressionNary*>(operand.get());
+            if (nary && str::equals(nary->getOpName(), getOpName()) && nary->isAssociative()) {
+                vpOperand.erase(vpOperand.begin() + i);
```
cswanson310: Instead of an erase and then a re-insert here, could you do the following instead?
```cpp
invariant(!nary->vpOperand.empty());
vpOperand[i] = nary->vpOperand[0];
vpOperand.insert(vpOperand.begin() + i, nary->vpOperand.begin() + 1, nary->vpOperand.end());
```

It would save two shifts of the array from erasing and then re-inserting.

src/mongo/db/pipeline/expression.cpp
salessandri: Between which lines do you want the blank line to be?

src/mongo/db/pipeline/expression.cpp
salessandri: Yes. One thing though, the last line should be:
```c++
vpOperand.insert(vpOperand.begin() + i + 1, nary->vpOperand.begin() + 1, nary->vpOperand.end());
```

Otherwise you would get the order of the inner operands altered.

src/mongo/db/pipeline/expression.cpp
salessandri: Between which lines do you want the blank line to be?

src/mongo/db/pipeline/expression.cpp
cswanson310: Between the close brace on 2232 and the if on 2234. It appears there is one now? Either I misread it the first time, or you fixed it... Either way, looks good now :)

src/mongo/db/pipeline/expression.cpp
cswanson310: Between 2242 and 2243. Just to visually separate the vpOperand from the if/else.

salessandri: Fixed

cswanson310:
Great! This looks good to me! I'll run the tests to make sure it hasn't accidentally broken anything, then go ahead and merge it. Thank you!

Schubes:
Merged under https://github.com/mongodb/mongo/commit/83afa046ee7470664f877b7d741d0a0e33780f16

PR #1044, opened 2015-11-17
SERVER-21494: Fix trivial syntax error in stack_locator_unknown.cpp

i80and:


ramonfm:
Thanks @i80and. Fix is in progress as part of [SERVER-21216](https://jira.mongodb.org/browse/SERVER-21216)

PR #1052, opened 2015-12-24
SERVER-18041: Added support for parallel cloning during initial sync

liranms:
It is extremely important to support for parallel cloning, especially during the index build phase.
As it happens, our MongoDB replicas sit on some powerful machines and its a waste using a single core, especially when using engines like wiredTiger and RocksDB which use compression.
Moreover, using the single core initial sync prevents us from adding new replicas as they take alot of time to copy and build the index (in the order of days), which causes them to loose sync, even after enlarging the oplog window.

Changes:
Updated Cloner to lock just the db it clones, instead of the global lock (as an existing code comment suggested but not implemented)
Added maxReplicationThreads as global server parameter to control number of threads used for initial cloning to prevent Scott's comment about possible DOS other replica set members (suggested by @scotthernandez) defaults to #cores + 1
Obviously its possible to set to 1 thread to keep current behavior.

### DIFF
--- a/src/mongo/db/cloner.cpp
+++ b/src/mongo/db/cloner.cpp
@@ -119,9 +119,8 @@ struct Cloner::Fun {
     void operator()(DBClientCursorBatchIterator& i) {
         invariant(from_collection.coll() != "system.indexes");
 
-        // XXX: can probably take dblock instead
         scoped_ptr<ScopedTransaction> scopedXact(new ScopedTransaction(txn, MODE_X));
-        scoped_ptr<Lock::GlobalWrite> globalWriteLock(new Lock::GlobalWrite(txn->lockState()));
+        scoped_ptr<Lock::DBLock> dbWriteLock(new Lock::DBLock(txn->lockState(), _dbName, MODE_X));
         uassert(ErrorCodes::NotMaster,
                 str::stream() << "Not primary while cloning collection " << from_collection.ns()
                               << " to " << to_collection.ns(),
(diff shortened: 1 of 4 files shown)


ramonfm:
@liranms, before we can consider your pull request you'll need to sign up the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement).

Thanks,
Ramón.

scotthernandez:
@liranms,

The pull request looks interesting and is close to some of the work we are doing for the new DataReplicator system, which will replace the cloner and data copy/sync parts of replication. In the new DataReplicator code, which is the server issue you used for this pull request, we plan to support concurrent cloning of collections (not databases). This should address many users who have lots of collections but not a lot of databases. In addition we want to support the ability to resume/restart if the system is shutdown or has an upstream error; That support requires other changes to how the oplog is populated during the initial sync process which we haven't done yet, but should be part of the next release as well.

As we are replacing this cloner code in the next release we are not excited to review or include these kinds of changes. Doing so would require us to do additional testing and maintain multiple forms of concurrent initial sync code on legacy versions, like 3.0/3.2. We already have to do this for the new code, which will be C++ unit testable and provide much better coverage that the current tests.

If you really want to push for this change being backported to older versions (like the current 3.2 stable release) then we should create a new server issue to address these changes to the cloner (not DataReplicator) and mark them for backport to the releases you wish to see them in. We will then need to update the pull request to be ported to those versions, and include tests to verify correctness.

Also, if you are interested in working on the new code please watch the DataReplicator jira issues and chime in as we get stuff more stable, or comment on them as we schedule them for dev releases (happening over the next few months). There should be a lot of room for improvement once that code become more stable and replaces, much, if not all, of the initial sync process.

Thanks for the work on the pull request and it is good to know that our plan to support concurrent cloning will help you a lot when we land it in the next release.

--Scott

ramonfm:
I'm closing this pull request as per @scotthernandez's message above.

PR #1053, opened 2015-12-31
2.6: Fix build on gcc 5.3

brauliobo:


src/mongo/base/global_initializer.cpp
```diff
@@ -24,13 +24,4 @@ namespace mongo {
         return theGlobalInitializer;
     }
 
-namespace {
-
-    // Make sure that getGlobalInitializer() is called at least once before main(), and so at least
-    // once in a single-threaded context.  Otherwise, static initialization inside
-    // getGlobalInitializer() won't be thread-safe.
-    Initializer* _theGlobalInitializer = &getGlobalInitializer();
```
acmorrow: You definitely can't do this, as it will stop all global initializers from running. Did you check that mongod actually starts up with these changes?

src/mongo/client/examples/simple_client_demo.vcxproj
```diff
@@ -182,4 +182,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
```
acmorrow: Seems unrelated to GCC

src/mongo/db/commands/index_filter_commands.cpp
```diff
@@ -167,8 +167,8 @@ namespace mongo {
         // This is a read lock. The query settings is owned by the collection.
         Client::ReadContext readCtx(ns);
         Client::Context& ctx = readCtx.ctx();
-        QuerySettings* querySettings;
-        PlanCache* unused;
+        QuerySettings* querySettings = NULL;
```
acmorrow: This shouldn't be necessary. What warning or error is GCC emitting? Pointers to these are passed into getQuerySettingsAndPlanCache right below.

src/third_party/boost/boost/system/error_code.hpp
```diff
@@ -211,9 +211,6 @@ namespace boost
     inline const error_category &  get_system_category() { return system_category(); }
     inline const error_category &  get_generic_category() { return generic_category(); }
     inline const error_category &  get_posix_category() { return generic_category(); }
-    static const error_category &  posix_category = generic_category();
```
acmorrow: In general, we prefer to not change the boost sources unless overwhelmingly required. What warning or error prompted this change?

acmorrow:
@brauliobo - Thanks for the PR. Are the majority of these fixes related to GCC 5 warnings? If so, the recommended way to build older MongoDB branches with newer compilers is to use the `--disable-warnings-as-errors` SCons flag. You will still get the compiler warnings, but they won't be considered as errors. Can you try that and see if that gets you a working build without these changes? In general we don't upgrade old branches just to support newer compilers.

Thanks,
Andrew

brauliobo:
Hi @acmorrow, will try with that flag.

brauliobo:
@acmorrow it built successfully with `--disable-warnings-as-errors`, so maybe this should became default?

acmorrow:
@brauliobo - No, we don't want it to be the default because we require that 2.6 build on our CI system with out warnings. The `--disable-warnings-as-errors` flag exists to override the default setup only when needed.

ramonfm:
@brauliobo, if this pull request is something you still want us to consider please see the [contributor's guide](https://github.com/mongodb/mongo/blob/411e9810075556fb196278a669fab0f19ea901ce/CONTRIBUTING.rst) for information on the required steps, including [opening a JIRA ticket in the SERVER project](https://jira.mongodb.org) and signing the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement).

Thanks,
Ramón.

Schubes:
Hi @brauliobo,

Since we haven't heard back since Ramon's request to sign the contributor's agreement, and MongoDB 2.6 is not longer supported. I'm going to close this pull request.

Kind regards,
Kelsey

PR #1055, opened 2016-01-09
Use markdown for the README

leo:
This PR cleans up the contents of the `README` file and uses markdown to structure them in a better way. This makes [reading](https://github.com/leo/mongo/tree/patch-1#mongodb) and therefore contributing to the repository much easier.


ramonfm:
Thanks for the pull request @leo  . While the top-level README looks "ugly" on github, its main goal is to be readable when read with cat, less, etc., so we're currently not accepting pull requests for format changes to the README at this time.

Cheers,
Ramón.

PR #1057, opened 2016-01-11
Acknowledge user based environment

vkuznet:
In our environment we build entire Linux toolchain (gcc, python, etc.) and would like to build mongo against it. This is necessary since our build nodes run old version of toolchain. The scons system needs to acknowledge user environment in this case to pick up custom gcc/g++/python tools. The proposed change fix that issue by reading user environment.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -815,6 +815,7 @@ envDict = dict(BUILD_ROOT=buildDir,
                INSTALL_DIR=installDir,
                CONFIG_HEADER_DEFINES={},
                LIBDEPS_TAG_EXPANSIONS=[],
+               ENV = os.environ,
                )
 
 env = Environment(variables=env_vars, **envDict)


acmorrow:
Hi @vkuznet - This isn't the right way to do this. SCons, as part of its design, intentionally does not allow environment variables into the build by default. This is intended to ensure reproducible builds, and this is actually a feature that we support.

However, we do recognize that sometimes, as in your case, you need to do this. So, we have provided a mechanism to allow users to externally script injections into the SCons environment.

This is accomplished by configuring SCons 'Variables'. Among those variables is `ENV`, which contains a dictionary of shell environment variables that should be propagated into the environment of processes started by SCons.

While Variables are typically configured on the command line (e.g. `scons VAR=value`), it is clear that the `ENV` variable cannot be configured in this manner (how would you specify the dictionary?). Instead, we have provided the ability to load Variables from a file, with the SCons flag `--variables-files`.

The `--variables-files` option takes a delimiter separated list of files and loads Variable definitions from them. In particular, we have provided a pre-canned Variables file that explicitly imports the calling shell environment to handle situations like the one you have described:

https://github.com/mongodb/mongo/blob/master/etc/scons/propagate_shell_environment.vars

Invoking SCons as:

`scons --variables-files=etc/scons/propagate_shell_environment.vars ...`

will achieve the same effect as the change you have made in this pull request, without unconditionally forcing it on all users of the MongoDB build system.

vkuznet:
Ok, I understand, you're carrying about consistency at distribution level, while we're carrying at package level. I rebuilt new MongoDB with your proposed option and it is fine, closing this PR.

PR #1060, opened 2016-01-20
[WIP] [2.6] Alpine linux compatibility

Soullivaneuh:
Replaces #1058 #1059.

I add a docker files to test. Will remove it at the end.

To test it, install [Docker](https://www.docker.com/) and run:

``` bash
docker build -t mongo:2.6-build .
docker run --rm -it mongo:2.6-build
```

Ill update the main message with issues and question to resume the progress.

Tell me if you want another PR for 3.x or if you will backport it yourself.

Please not that Alpine Linux use [Musl](http://www.musl-libc.org/) C library.
- [x] Fix `gnu_get_libc_version` usage that does not exist on Musl
- [x] Add `/etc/alpine-release` file map
- [x] `mallinfo (void)` Does not exists on Musl. Find a generic way.
  - http://stackoverflow.com/q/34905260/1731473
- [x] Fix `'pthread_yield' was not declared in this scope`
- [x] `Dl_info_t` missing
- [ ] Fix `'walkcontext' was not declared in this scope`

Don't hesitate to help if you have some suggestion about unfinished tasks! :+1:

PR #1061, opened 2016-01-21
[3.x] Alpine linux compatibility

Soullivaneuh:
Same as #1060 but for 3.x version.

- [x] alpine-release on processinfo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM alpine:3.3
+
+ADD . /code
+
+WORKDIR /code
+
+RUN ldd --version || true
+
+RUN apk add --no-cache curl tar scons g++ linux-headers openssl-dev
+RUN scons mongod -j$(getconf _NPROCESSORS_ONLN) --ssl --disable-warnings-as-errors
+RUN scons mongo -j$(getconf _NPROCESSORS_ONLN) --ssl --disable-warnings-as-errors
+RUN scons mongos-j$(getconf _NPROCESSORS_ONLN) --ssl --disable-warnings-as-errors
+RUN scons dbtest -j$(getconf _NPROCESSORS_ONLN) --ssl --disable-warnings-as-errors
+RUN scons unittests -j$(getconf _NPROCESSORS_ONLN) --ssl --disable-warnings-as-errors
+
+RUN scons --prefix=/opt/mongo install
+
+VOLUME /data/db
+
+COPY docker-entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+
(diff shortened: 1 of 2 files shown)


ncopa:
chriskohlhoff/asio#95 is also relevant

Soullivaneuh:
Another third_party linked issue: gperftools/gperftools#693

ncopa:
Yes. I think they only use tcmalloc from gperftools. Probably because it performs better than glibc malloc. (ptmalloc2). I wonder what the performance difference is compareed to musl libc's malloc. If musl libc malloc is fast enough, then it may make sense to add an option `--use-system-malloc`, but I suspect that tcmalloc works better in the mongodb workload.

So fixing gperftools/tcmalloc is probably a good idea.

We could probably also build gperftools as separate package (or libtcmalloc at least) and build mongodb with  `--use-system-tcmalloc`.

ncopa:
mongodb builds now for me with https://gist.github.com/ncopa/520055e642043684da2d

sylus:
According to gperftools/gperftools#693:

ncopa's commit makes it build, but dynamically linked musl does not support interposing malloc/free. I.e. musl's strdup will call musl's malloc and not interposed version. This makes it impossible to support musl in any malloc implementation not just tcmalloc.

So close to have mongodb compile in alpine. :S

acmorrow:
Just to point out, you could certainly build mongodb without tcmalloc, by building with the `--allocator=system` SCons flag.

sylus:
I am having other errors too when building related to log.cpp and convert from int to char*

ncopa:
you'll find some patches here: http://git.alpinelinux.org/cgit/aports/tree/testing/mongodb

Soullivaneuh:
@sylus So no chance to get this working? :-/

@ncopa What should we do for https://github.com/docker-library/mongo/pull/80? Apply patches proposed by Alpine?

jvduf:
hi guys! Interesting to see there has been good progress on this. Too bad it has stalled. I've tried to apply some of this changes (the ones in this pull request, not the APK patches) but don't think these are up to date with latest v3.4 branch... Still break and not building.

Any chance to pick this up? Would love to help but I have no experience at all with the "low level" stuff that's going on here :(

@Soullivaneuh any ideas if this is still something you guys are considering?

Soullivaneuh:
Unfortunately, I don't have more allowed work time for this implementation right now.

Feel free to cherry-pick and continue my PR. I also gave maintainers edit here.

Regards.

kelly-cs:
Closing since this is merging into 3.2 which is unsupported as of September 30, 2018

PR #1068, opened 2016-02-19
SERVER-7695: implement iso week date aggregators

jigfox:
this implements operators for dealing with week as described in the
ISO-8601 standard (https://en.wikipedia.org/wiki/ISO_week_date).

the implemented oprators are:

- `$isoWeek` returns the ISO-8601 week number of the given date
- `$isoDayOfWeek` returns the number of the day with the week from 1 for Monday to 7 for Sunday
- `$isoWeekYear` returns the year of the ISO-8601 week of the date

example:

    { $isoWeek: ISODate("2016-01-01T00:00:00Z") }

returns 53

    { $isoDayOfWeek: ISODate("2016-01-01T00:00:00Z") }

returns 5

    { $isoWeekYear: ISODate("2016-01-01T00:00:00Z") }

returns 2015

this also implements the specifiers for $dateToString so it is possible
to create Dates defined by week like this:

    {
        $dateToString: {
            format: "%G-W%V-%u",
            date: ISODate("2016-01-01T00:00:00Z")
        }
    }

which will return "2015-W53-5"

Open Issues:

- [x] Make `{$dateToString: { format: "%G-W%V", date: new Date("1900-12-31T23:59:59Z")}}` work


src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,104 @@ const char* ExpressionWeek::getOpName() const {
...
+    Value pDate(vpOperand[0]->evaluateInternal(vars));
+    return Value(extract(pDate.coerceToTm()));
+}
+
+int ExpressionIsoWeekYear::extract(const tm& tm) {
+    if (tm.tm_mon > 0 && tm.tm_mon < 11) {
+        return tm.tm_year + 1900;
+    } else if (tm.tm_mon == 0) {
+        int isoWeek = ExpressionIsoWeek::extract(tm);
+        if (isoWeek > 51) return tm.tm_year + 1900 - 1;
```
cswanson310: Please use braces around all if statements. Also, can you add some comments to this method briefly describing what each case is trying to account for?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,104 @@ const char* ExpressionWeek::getOpName() const {
...
+        } else { // non leap year
+            return 52;
+        }
+    } else {
+        return 52;
+    }
+}
+
+int ExpressionIsoWeek::extract(const tm& tm) {
+    //int isoDayOfWeek = ExpressionIsoDayOfWeek::extract(tm);
```
cswanson310: Please delete.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,104 @@ const char* ExpressionWeek::getOpName() const {
...
+int ExpressionIsoWeek::lastWeek(int year) {
+    struct tm tm;
+    tm.tm_year = year - 1900;
+    tm.tm_mon = 11;
+    tm.tm_mday = 31;
+    tm.tm_hour = 23;
+    tm.tm_min = 59;
+    tm.tm_sec = 59;
+    mktime(&tm);
+    if (tm.tm_wday > 0 && tm.tm_wday < 4) {
```
cswanson310: Everything from here down is confusing and non-intuitive to me. After reading the wikipedia article on "ISO week date" (https://en.wikipedia.org/wiki/ISO_week_date) I'm still not following. Can you add comments and/or point me to where you got this logic?

jstests/aggregation/isodates.js
```diff
@@ -0,0 +1,243 @@
+/*
```
cswanson310: While these tests are wonderful, and have good coverage, is there any way we could add them to [this test file](https://github.com/mongodb/mongo/blob/r3.3.2/jstests/aggregation/bugs/server6189.js)? If we need more coverage, I'd rather test all the other date operators as well. If not, I'd suggest adding some as unit tests in [this file](https://github.com/mongodb/mongo/blob/r3.3.2/src/mongo/db/pipeline/expression_test.cpp)

jigfox: I don't like reducing the coverage, as you already mentioned the algorithms aren't easy to follow, so I'd rather have an extensive test suite, which contains every edge case to prevent regression if someone changes this code.

[jstests/aggregation/bugs/server6189.js](https://github.com/mongodb/mongo/blob/r3.3.2/jstests/aggregation/bugs/server6189.js): ist for testing dates before 1970, I could try adding the new operators here. But I would still keep the other file for testing all edges there are.

[src/mongo/db/pipeline/expression_test.cpp](https://github.com/mongodb/mongo/blob/master/src/mongo/db/pipeline/expression_test.cpp): doesn't include any tests on dates (that I see). I was searching the code for date tests and only found the javascript ones. And also because I'm more familiar with javascript than with c++ I added  the new tests as new javascript file.

cswanson310: Ok points taken. It looks like this test is copied from testall.js, which is not my favorite test. It's a little hard to follow, uses bad variable names, different styles of comments, and relies on data that is provided from another file, which gives no intuition as to why the assertions might be true. Sorry for providing a bad example test. I'd like to eventually refactor and get rid of that test, but it's not very pressing...

I'd suggest rewriting this test to be more in the style of unit tests. [this test](https://github.com/mongodb/mongo/blob/r3.3.2/jstests/aggregation/bugs/server19548.js) comes to mind as a good example of providing good coverage, being easy to read, and having little test fixture overhead. I'll let you decide what's the best way to do this, but I'm thinking there's probably some to write the test such that we can provide one date and the expected values for $week, $isoWeek, $year, $isoWeekYear, $dayOfWeek, and $isoDayOfWeek.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,123 @@ const char* ExpressionWeek::getOpName() const {
...
+/* ------------------------- ExpressionIsoWeekYear ---------------------- */
+
+Value ExpressionIsoWeekYear::evaluateInternal(Variables* vars) const {
+    Value pDate(vpOperand[0]->evaluateInternal(vars));
+    return Value(extract(pDate.coerceToTm()));
+}
+
+int ExpressionIsoWeekYear::extract(const tm& tm) {
+    if (tm.tm_mon > 0 && tm.tm_mon < 11) {
+        // if month isnt January or December it's just the year given
```
cswanson310: Please format comments as full sentences. In particular, this one should start with a capital letter, and end in a period. I realize this practice isn't followed elsewhere in this file, but we're trying to have all new code follow this style.

Ditto for other comments in this file.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,123 @@ const char* ExpressionWeek::getOpName() const {
...
+
+REGISTER_EXPRESSION(isoDayOfWeek, ExpressionIsoDayOfWeek::parse);
+const char* ExpressionIsoDayOfWeek::getOpName() const {
+    return "$isoDayOfWeek";
+}
+
+/* ------------------------- ExpressionIsoWeekYear ---------------------- */
+
+Value ExpressionIsoWeekYear::evaluateInternal(Variables* vars) const {
+    Value pDate(vpOperand[0]->evaluateInternal(vars));
```
cswanson310: I think you can just call this variable 'date'. I believe the others were called 'pDate' because there was once a time when it's type was a `Value *`, so this was a pointer to a date. Now it's just a date, so you can call it that.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,123 @@ const char* ExpressionWeek::getOpName() const {
...
+    // create YYYY-12-31T23:59:59 so only 1 second left to new year
+    struct tm tm;
+    tm.tm_year = year - 1900;
+    tm.tm_mon = 11;
+    tm.tm_mday = 31;
+    tm.tm_hour = 23;
+    tm.tm_min = 59;
+    tm.tm_sec = 59;
+    mktime(&tm);
+    /* From: https://en.wikipedia.org/wiki/ISO_week_date#Last_week
```
cswanson310: Please use `// Comment` style comments when the comment is not documenting a method, and
```
/**
 * Comment
 */
```
style comments when the comment is documenting a method.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,123 @@ const char* ExpressionWeek::getOpName() const {
...
+     *
+     * If 31 December is on a Monday, Tuesday or Wednesday, it is in week 01
+     * of the next year. If it is on a Thursday, it is in week 53 of the year
+     * just ending; if on a Friday it is in week 52 (or 53 if the year just
+     * ending is a leap year); if on a Saturday or Sunday, it is in week 52 of
+     * the year just ending.
```
cswanson310: Great. This helps a lot, thanks!. Please wrap the comment at 100 characters.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,123 @@ const char* ExpressionWeek::getOpName() const {
...
+        } else { // leap year
+            return 53;
+        }
+    } else { // Sat (6) or Sun (0)
+        return 52;
+    }
+}
+
+int ExpressionIsoWeek::extract(const tm& tm) {
+    int isoDayOfWeek = (tm.tm_wday - 7) % 7 + 7;
```
cswanson310: Prefer `int isoDayOfWeek = ExpressionIsoDayOfWeek::extract(tm);`?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,123 @@ const char* ExpressionWeek::getOpName() const {
...
+    int isoDayOfYear = tm.tm_yday + 1;
+
+    int isoWeek = (isoDayOfYear - isoDayOfWeek + 10) / 7;
```
cswanson310: Can you add a comment describing this logic please?

jstests/aggregation/isodates.js
jigfox: I will look at [jstests/aggregation/bugs/server19548.js](https://github.com/mongodb/mongo/blob/r3.3.2/jstests/aggregation/bugs/server19548.js) and see what I can do, but please be patient, It was kind if tedious to write these tests, and yes they were copied from testall.js 😉 I always try to copy the style of the existing code when contributing to open source. Sadly I took the wrong example now. But you're absolutely right about readability of tests, so I will look into it and see what I can do.

cswanson310: No problem! I am sorry to have you re-write them, but I think it will be worth it. Thank you!

jigfox: @cswanson310: I've added a new [test file](https://github.com/jigfox/mongo/blob/SERVER-7695-aggregation-operators-for-iso-week/jstests/aggregation/bugs/server7695_isodates.js), which is much more explanatory.

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,268 @@
...
+        assert.eq(
+            coll.aggregate(pipeline).toArray()[0].result,
+            [{result: expResult}][0].result,
+            msg);
+    }
+
+    // this works
+    testOp('$dateToString', { date: new Date("1900-12-31T23:59:59Z"), format: "%V-%G"}, "01-1901")
+    // this is failing, but it shouldn't as it is the same as above, only rotated
+    testOp('$dateToString', { date: new Date("1900-12-31T23:59:59Z"), format: "%G-%V"}, "1901-01")
```
jigfox: @cswanson310 while creating some more tests for some more edge cases, I found a bug in the calculation of the isoWeek. Sadly I don't have any idea right now, why this is happening, because it only appears if I ask for the year before I ask for the week. The other way around it works perfectly fine.

jigfox: db.test.insert({})
    db.test.aggregate([{
      $project:{
        _id:0,
        result:{
          $dateToString: {
            date:new Date("1900-12-31T23:59:59Z"),
            format: "%G-W%V"
          }
        }
      }
    }]) should result in "1901-W01" but it results in "1901-W53"

and

    db.test.insert({})
    db.test.aggregate([{
      $project:{
        _id:0,
        result:{
          $dateToString: {
            date:new Date("1900-12-31T23:59:59Z"),
            format: "W%V-%G"
          }
        }
      }
    }]) results in "W01-1901"

jigfox:
@cswanson310 : I tried to debug the problem which returns invalid Date for `{$dateToString: { format: "%G-W%V", date: new Date("1900-12-31T23:59:59Z")}}` by adding log output, but as soon as I add logging, the tests pass.

jstests/aggregation/bugs/server7695_isodates.js
cswanson310: The first one works for me:
```
db.test.insert({})
> db.test.aggregate([{
   $project:{
     _id:0,
     result:{
       $dateToString: {
         date:new Date("1900-12-31T23:59:59Z"),
         format: "%G-W%V"
       }
     }
   }
 }]) // should result in "1901-W01" but it results in "1901-W53"
{ "result" : "1901-W01" }
```

The second one seems correct to me? Dec 31, 1900 is a Monday, Jan 3 is a Thursday, and Jan 4 is a Friday. By my understanding, this means it is part of the first week of 1901.

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,269 @@
...
+
+    // Helper for testing that op returns expResult.
+    function testOp(op, value, expResult) {
+        testOpCount++;
+        var pipeline = [{$project: {_id: 0, result: {}}}];
+        pipeline[0].$project.result[op] = value;
+        var msg = "Exptected {"+op+": "+value+"} to equal: "+expResult;
+        assert.eq(
+            coll.aggregate(pipeline).toArray()[0].result,
+            [{result: expResult}][0].result,
```
cswanson310: Just `expResult`?

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,269 @@
...
+            msg);
+    }
+
+    // while development there was a bug which caused an error with $dateToString if the order of
+    // %V and %G changed, so I added this test to prevent regression
```
cswanson310: Start comments with capitals, end with periods please.

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,269 @@
...
+    testOp('$dateToString', { date: new Date("1900-12-31T23:59:59Z"), format: "%V-%G"}, "01-1901")
+    // this is failing, but it shouldn't as it is the same as above, only rotated
```
cswanson310: This doesn't fail for me... Is it still failing for you?

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,269 @@
...
+
+    var MONDAY = 1;
+    var TUESDAY = 2;
+    var WEDNESDAY = 3;
+    var THURSDAY = 4;
+    var FRIDAY = 5;
+    var SATURDAY = 6;
+    var SUNDAY = 7;
+
+    function padded(val) { return ("00" + val).slice(-2) }
```
cswanson310: Missing a semicolon after the `return` statement.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3087,149 @@ const char* ExpressionWeek::getOpName() const {
...
+    // floor((7-4+10)/7) = floor(13/7) => 1
+    //
+    // week(date)    = isoWeek
+    // ordinal(date) = isoDayOfYear
+    // weekday(date) = isoDayOfWeek
+    int isoDayOfWeek = ExpressionIsoDayOfWeek::extract(tm);
+    int isoDayOfYear = tm.tm_yday + 1;
+    int isoWeek = (isoDayOfYear - isoDayOfWeek + 10) / 7;
+
+    if (tm.tm_year == 0) {
```
cswanson310: You can remove this now?

jstests/aggregation/bugs/server7695_isodates.js
cswanson310: Edit: I removed the log line, and it started failing for me again. I think I may have found the source, but I'm still unsure why it's happening:
```
Process 15116 stopped
* thread #2: tid = 0x4c5827, 0x000000010097f0be mongod`mongo::ExpressionIsoWeek::lastWeek(year=1900) + 62 at expression.cpp:3181, stop reason = step over
    frame #0: 0x000000010097f0be mongod`mongo::ExpressionIsoWeek::lastWeek(year=1900) + 62 at expression.cpp:3181
   3178	    tm.tm_hour = 23;
   3179	    tm.tm_min = 59;
   3180	    tm.tm_sec = 59;
-> 3181	    mktime(&tm);
   3182	
   3183	    // From: https://en.wikipedia.org/wiki/ISO_week_date#Last_week :
   3184	    // If 31 December is on a Monday, Tuesday or Wednesday, it is in week 01 of
(lldb) n
2016-03-14T11:13:55.799-0400 I STORAGE  [DataFileSync] flushing mmaps took 178685ms  for 5 files
2016-03-14T11:13:55.799-0400 I NETWORK  [initandlisten] connection accepted from 127.0.0.1:52490 #2 (2 connections now open)
Process 15116 stopped
* thread #2: tid = 0x4c5827, 0x000000010097f0c6 mongod`mongo::ExpressionIsoWeek::lastWeek(year=1900) + 70 at expression.cpp:3191, stop reason = step over
    frame #0: 0x000000010097f0c6 mongod`mongo::ExpressionIsoWeek::lastWeek(year=1900) + 70 at expression.cpp:3191
   3188	    // 52 (or 53 if the year just ending is a leap year); if on a Saturday or
   3189	    // Sunday, it is in week
   3190	    // 52 of the year just ending.
-> 3191	    if (tm.tm_wday > 0 && tm.tm_wday < 4) {  // Mon(1), Tue(2), and Wed(3)
   3192	        return 1;
   3193	    } else if (tm.tm_wday == 4) {  // Thu (4)
   3194	        return 53;
(lldb) p tm.tm_wday
(int) $17 = 6938768
(lldb) p tm
(tm) $18 = {
  tm_sec = 59
  tm_min = 59
  tm_hour = 23
  tm_mday = 31
  tm_mon = 11
  tm_year = 0
  tm_wday = 6938768
  tm_yday = 28672
  tm_isdst = 1
  tm_gmtoff = 25769803782
  tm_zone = 0x0000000700000005 ""
}
```

I'll continue to investigate.

cswanson310: Ok, its still unclear to me why adding the log line or reversing the order changes anything, but I think if you want to create a `tm` struct, you should go through [Value::coerceToTm()|https://github.com/mongodb/mongo/blob/r3.3.2/src/mongo/db/pipeline/value.cpp#L533]. That code looks like it's been designed to handle weird edge cases, and will work on windows as well.

Try that and see if it fixes anything.

jstests/aggregation/bugs/server7695_isodates.js
jigfox: Yes, why?

jstests/aggregation/bugs/server7695_isodates.js
jigfox: `Value::coerceToTm()` only works on a value, but I just have the year in this Method, because the date wouldn't work. I need the end of the year not just any date in the year. But your debugging output helped a lot and I got it working with `mktime()`

jstests/aggregation/bugs/server7695_isodates.js
cswanson310: I mean, could you replace `[{result: expResult}][0].result` with `expResult`?

jigfox: Of course, I don't know why I made it this complicated

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,150 @@ const char* ExpressionWeek::getOpName() const {
...
+    tm.tm_year = year - 1900;
+    tm.tm_mon = 11;
+    tm.tm_mday = 31;
+    tm.tm_hour = 23;
+    tm.tm_min = 59;
+    tm.tm_sec = 59;
+    tm.tm_wday = -1; // this was enough to solve the bug for 1900
+    tm.tm_yday = -1;
+    tm.tm_gmtoff = 0;
+    tm.tm_isdst = 0;
```
cswanson310: Unfortunately, this doesn't work on windows, which is one of our supported platforms.
```
Compiling build\cached\mongo\db\pipeline\expression.obj
expression.cpp
src\mongo\db\pipeline\expression.cpp(3127) : error C2039: 'tm_gmtoff' : is not a member of 'tm'
        C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\INCLUDE\wchar.h(1170) : see declaration of 'tm'
scons: *** [build\cached\mongo\db\pipeline\expression.obj] Error 2
scons: building terminated because of errors.
```
Here's the definition:
```c
#ifndef _TM_DEFINED
struct tm {
 int tm_sec; /* seconds after the minute - [0,59] */
 int tm_min; /* minutes after the hour - [0,59] */
 int tm_hour; /* hours since midnight - [0,23] */
 int tm_mday; /* day of the month - [1,31] */
 int tm_mon; /* months since January - [0,11] */
 int tm_year; /* years since 1900 */
 int tm_wday; /* days since Sunday - [0,6] */
 int tm_yday; /* days since January 1 - [0,365] */
 int tm_isdst; /* daylight savings time flag */
 };
#define _TM_DEFINED
#endif /* _TM_DEFINED */
```

This was fixed for me by changing the initialization of tm: 
```c++
struct tm tm = {};  // This will initialize all struct members with 0.
```
Then removing lines 3158 and 3159. See [this question on stackoverflow](http://stackoverflow.com/questions/1069621/are-members-of-a-c-struct-initialized-to-0-by-default) for struct initialization details.

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,268 @@
...
+        testOpCount++;
+        var pipeline = [{$project: {_id: 0, result: {}}}];
+        pipeline[0].$project.result[op] = value;
+        var msg = "Exptected {"+op+": "+value+"} to equal: "+expResult;
+        assert.eq(coll.aggregate(pipeline).toArray()[0].result, expResult, msg);
+    }
+
+    // While development, there was a bug which caused an error with $dateToString if the order of
+    // %V and %G changed, so I added this test to prevent regression.
+    testOp('$dateToString', {date: new Date("1900-12-31T23:59:59Z"), format: "%V-%G"}, "01-1901");
```
cswanson310: These tests do not pass on some windows machines. I think you'll have to wrap these, and other pre-1970 tests in conditionals like we do [here](https://github.com/mongodb/mongo/blob/r3.3.2/jstests/aggregation/bugs/server6189.js#L43-L47).

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,150 @@ const char* ExpressionWeek::getOpName() const {
...
+    // 52 (or 53 if the year just ending is a leap year); if on a Saturday or Sunday, it is in week
+    // 52 of the year just ending.
+    if (tm.tm_wday > 0 && tm.tm_wday < 4) { // Mon(1), Tue(2), and Wed(3)
+        return 1;
+    } else if (tm.tm_wday == 4) { // Thu (4)
+        return 53;
+    } else if (tm.tm_wday == 5) { // Fri (5)
+        // On Fri it's week 52 for non leap years and 53 for leap years.
+        // https://en.wikipedia.org/wiki/Leap_year#Algorithm
+        if (year % 4 != 0) { // non leap year
```
cswanson310: Can you define a helper for `isLeapYear`? I think it will help readability. See an example helper defined [here](https://github.com/mongodb/mongo/blob/r3.3.2/src/mongo/db/pipeline/expression.cpp#L2514-L2519). It should go right above the expression that uses it.

jstests/aggregation/bugs/server7695_isodates.js
jigfox: That is sad, but I will change it.

jigfox: So I just made the tests less aggressive on windows and or 32-bit systems, like [here](https://github.com/mongodb/mongo/blob/r3.3.2/jstests/aggregation/bugs/server6189.js#L43-L47)

cswanson310:
This is looking good to me! The jstest still doesn't pass on windows, mostly because 'value' isn't always a date. I'm guessing you don't have easy access to a windows machine? Would you mind if I took over from here to make the test pass on windows? I'm happy with the C++ changes, with some minor exceptions around comments that I can take care of.

If you're okay with that, I can give you a chance to look my changes before I push them.

jigfox:
I am happy to almost everything looks fine, I just changed the tests already to work with `$dateToString` values. I don't mind at all, if you took over. But I'm still happy to answer any questions you might have.

ValentinH:
This is a really cool PR since I have been watching SERVER-7695 for more than a year.
I don't have enough knowledge to tell something about the implementation, but as a user, this would be a really good feature.

:+1:

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,156 @@ const char* ExpressionWeek::getOpName() const {
...
+REGISTER_EXPRESSION(isoWeekYear, ExpressionIsoWeekYear::parse);
+const char* ExpressionIsoWeekYear::getOpName() const {
+    return "$isoWeekYear";
+}
+
+/* ------------------------- ExpressionIsoWeek -------------------------- */
+
+namespace {
+bool isLeapYear(int year) {
+    if (year % 4 != 0) { // non leap year
```
cswanson310: Can you reflow these comments to be a little more descriptive, like this:
```C++
if (year % 4 != 0) {
    // Not divisible by 4, so not a leap year.
   return false;
} else if (year % 100 != 0) {
    // Divisible by 4, but not by 100, so it is a leap year.
    return true;
} ...
```

Particularly by the time I got to the last case, it was hard to keep track of which conditions were true.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3072,6 +3085,156 @@ const char* ExpressionWeek::getOpName() const {
...
+}
+
+// Quote https://en.wikipedia.org/wiki/ISO_week_date :
+// Weeks start with Monday. The first week of a year is the week that contains the first Thursday of
+// the year (and, hence, always contains 4 January).
+int ExpressionIsoWeek::extract(const tm& tm) {
+    // Calculation taken from:
+    // https://en.wikipedia.org/wiki/ISO_week_date#Calculating_the_week_number_of_a_given_date
+    //
+    //              │ ordinal(date) - weekday(date) + 10 │
```
cswanson310: As a general rule, we try not to use unicode characters in our source code where it can be avoided. Can you re-work these lines to be `week(date) = floor((ordinal(date) - weekday(date) + 10) / 7)`?

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,284 @@
...
+    }
+    function getNewYearsEve(year) {
+        return new Date("" + year + "-12-31T23:59:59Z");
+    }
+    function shiftWeekday(day, days) {
+        return ((day - 1 + days) % 7) + 1;
+    }
+
+    ['common', 'leap', 'commonAfterLeap'].forEach(function(type) {
+        years[type].forEach(function(year, day) {
```
cswanson310: `day` is not a very descriptive variable name here... it really refers to the index in the array. I'd suggest renaming this variable to `i`, then re-working the comment on line 123 to describe how the index corresponds to the day.

jstests/aggregation/bugs/server7695_isodates.js
```diff
@@ -0,0 +1,284 @@
...
+
+    var MONDAY = 1;
+    var TUESDAY = 2;
+    var WEDNESDAY = 3;
+    var THURSDAY = 4;
+    var FRIDAY = 5;
+    var SATURDAY = 6;
+    var SUNDAY = 7;
+
+    function padded(val) {
```
cswanson310: Can you add brief comments to these helpers to describe what they're doing, when it's not obvious. Or rename them to be more obvious. For example, 'padded' really only pads with up to 2 '0's, what it's padding with and how many of them it inserts is unclear without reading the implementation.

cswanson310:
Ok great! All my remaining comments are suggestions on commenting. The test now passes on windows. I'm going to run it through all our test suites to make sure there aren't any other bugs I didn't catch.

jstests/aggregation/bugs/server7695_isodates.js
jigfox: The comment will be actually longer than the implementation itself ;-)

For the other helper methods, I already thought I had very descriptive names for the helper methods, could you please point me two helper which aren't obvious enough?

cswanson310:
As you may have noticed, I merged this pull request, so this is now in master. Thanks for the good work!

jigfox:
Thank You very much, especially baca use it was the very first time I wrote C++ Code in my life. My expertise is JavaScript and Ruby! I really appreciate the acceptance of the pull request.

PR #1070, opened 2016-02-25
SERVER-22847: Online change oplog size when using wiredtiger 

zydcom:
Support online change oplog size by collMod command.

eg:

change maxSize limit

```
var maxSize = 8192;
assert.commandWorked(
        db.runCommand({collMod: coll, cappedSize: maxSize})
        );
```

change maxDocs limit

```
var maxDocs = 10;
assert.commandWorked(
        db.runCommand({collMod: coll, cappedMaxDocs: maxDocs})
        );
```

### DIFF
--- a/src/mongo/db/storage/wiredtiger/wiredtiger_record_store.h
+++ b/src/mongo/db/storage/wiredtiger/wiredtiger_record_store.h
@@ -206,6 +206,16 @@ public:
     int64_t cappedMaxDocs() const;
     int64_t cappedMaxSize() const;
 
+    virtual bool setCappedSize(long long cappedSize) {
+        _cappedMaxSize = cappedSize;
+        return true;
+    }
+
+    virtual bool setCappedMaxDocs(long long cappedMaxDocs) {
+        _cappedMaxDocs = cappedMaxDocs;
+        return true;
+    }
+
     const std::string& getURI() const {
         return _uri;
     }
@@ -277,9 +287,9 @@ private:
     const bool _isEphemeral;
     // True if the namespace of this record store starts with "local.oplog.", and false otherwise.
     const bool _isOplog;
-    const int64_t _cappedMaxSize;
+    int64_t _cappedMaxSize;
     const int64_t _cappedMaxSizeSlack;  // when to start applying backpressure
-    const int64_t _cappedMaxDocs;
+    int64_t _cappedMaxDocs;
     RecordId _cappedFirstRecord;
     AtomicInt64 _cappedSleep;
     AtomicInt64 _cappedSleepMS;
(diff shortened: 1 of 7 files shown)


monkey101:
Hi @zydcom, thanks for your pull request.  We are reviewing it for inclusion in the next major release (v3.4).

src/mongo/db/storage/wiredtiger/wiredtiger_record_store.h
```diff
@@ -277,9 +287,9 @@ class WiredTigerRecordStore : public RecordStore {
     const bool _isEphemeral;
     // True if the namespace of this record store starts with "local.oplog.", and false otherwise.
     const bool _isOplog;
-    const int64_t _cappedMaxSize;
```
wolfkdy: a atomic64 should be better ?
collMod is in a different thread, partial_read/write problem should happen on a 32-bit machine, some cpu-visitable problem may also be considerable

src/mongo/db/storage/wiredtiger/wiredtiger_record_store.h
```diff
@@ -277,9 +287,9 @@ class WiredTigerRecordStore : public RecordStore {
...
+    int64_t _cappedMaxSize;
     const int64_t _cappedMaxSizeSlack;  // when to start applying backpressure
-    const int64_t _cappedMaxDocs;
```
wolfkdy: I think it a good idea of changing the two variables in order to support capped-set online change size.
But wt's mongo layer uses OplogStones to track whether it is necessary to truncate oplogs from the beginning.
_oplogStones is initialized once the service is started and never ever changed unless a manually truncate is applied to it. I didnt see any changes to _oplogStones in this patch, maybe I'm wrong?

zydcom: You are right，OplogStones parameter should be changed when resize oplog, like belows.

```
void WiredTigerRecordStore::OplogStones::adjust() {
    stdx::lock_guard<stdx::mutex> lk(_mutex);
    unsigned long long maxSize = _rs->cappedMaxSize();

    const unsigned long long kMinStonesToKeep = 10ULL;
    const unsigned long long kMaxStonesToKeep = 100ULL;

    unsigned long long numStones = maxSize / BSONObjMaxInternalSize;
    _numStonesToKeep = std::min(kMaxStonesToKeep, std::max(kMinStonesToKeep, numStones));
    _minBytesPerStone = maxSize / _numStonesToKeep;
    invariant(_minBytesPerStone > 0);
}   bool WiredTigerRecordStore::setCappedSize(long long cappedSize) {
    _cappedMaxSize = cappedSize;
    if (_oplogStones) {
        _oplogStones->adjust();
    }
    return true;
}

bool WiredTigerRecordStore::setCappedSize(long long cappedSize) {
    _cappedMaxSize = cappedSize;
    if (_oplogStones) {
        _oplogStones->adjust();
    }
    return true;
}

```

PR #1076, opened 2016-03-26
V3.2.3

mbasij1:
Add find in different level in document! for example
my json:
`{
id: 1,
table1:
[
{name: "ali"},
{name:"yy", family: "xx", a:10},
{name:'we', a:10}
]
}`

in most time need to get just 'table1' and just some element!
for example wherre in table1 that a=10 and want output like this!

`{
{table1_id:12, name:"yy", family: "xx", a:10},
{table1_id:13, name:'we', a:10}
}`
table1_id created like id of document and path to reach table1 in document!
why need this! think we put all table in one collection and don't have relation some times need to just show some detail that contain in child path! this need mostly ever that we have run-time forms and data-grids,
yes, mongodb have $ but $ just work one level and can't use more than one $ in find.
maybe say using map-reduce but for map-reduce should write query and have more table and if this map-reduce  every-time call can get extended time! think if this use for monitoring that want call every .5 sec in hundred computer


ramonfm:
Please take a look at [the aggregation framework](https://docs.mongodb.org/manual/aggregation/), since it should meet your needs.

For MongoDB-related support discussion please post on the [mongodb-user group](http://groups.google.com/group/mongodb-user) or [Stack Overflow with the "mongodb" tag](http://stackoverflow.com/questions/tagged/mongodb), where your question will reach a larger audience. A question like this involving more discussion would be best posted on the [mongodb-user group](http://groups.google.com/group/mongodb-user).

PR #1077, opened 2016-04-13
SERVER-23671 fix build failure with --use-system-tcmalloc but not --u…

mwhudson:
…se-system-valgrind


markbenvenuto:
@mwhudson Thanks for the contribution.

PR #1078, opened 2016-05-02
Page eviction can't follow up if there's a lot of disk read

SunguckLee:
MongoDB need a lot of disk read and pages are read into wiredtiger cache when data working-set is bigger than wiredtiger cache.
With high performance storage (like PCI-e or NVME SSD Drive), pages read into wiredtiger cache much faster than HDD.

For pages read into cache, Wiredtiger have to evict sufficient pages from cache.
Sometimes, WiredTiger can't make it. Once this happens, cache usage is going up and all foreground threads are responsible for page eviction also. But page eviction of foreground thread is not so optimized, so eventually all foreground threads are waiting for write-tickets.

So I propose, changing the formula of eviction target calculation and removing suppression of internal pages eviction ratio.

Some of MongoDB servers suffer this issue, and finally all foregrounds have been falling into waiting status.
After I changed this code block, MongoDB servers are really healthy and doing well.

### DIFF
--- a/src/third_party/wiredtiger/src/evict/evict_lru.c
+++ b/src/third_party/wiredtiger/src/evict/evict_lru.c
@@ -962,41 +962,16 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
 			if (read_gen_oldest != WT_READGEN_OLDEST)
 				break;
 		}
-
-		/*
-		 * Take all candidates if we only gathered pages with an oldest
-		 * read generation set.
-		 *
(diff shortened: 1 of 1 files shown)


SunguckLee:
I've forgot the JIRA Ticket.

https://jira.mongodb.org/browse/SERVER-23778

michaelcahill:
Discussion has moved to https://jira.mongodb.org/browse/SERVER-23778, if there are WiredTiger changes as a result, they will first go into https://github.com/wiredtiger/wiredtiger.

PR #1079, opened 2016-05-05
Some extreme case, internal page can never be evicted

SunguckLee:
This might be extreme case, But looks like this happens to me.

If __wt_tree_walk_count() return only internal page(btree->evict_ref),
expression "internal_pages >= (int)(evict - start) / 2)" always be true and internal page never can be evicted because both "evict" and "internal_pages" var are never changed.

So I think we should change "evict" to "end".

### DIFF
--- a/src/third_party/wiredtiger/src/evict/evict_lru.c
+++ b/src/third_party/wiredtiger/src/evict/evict_lru.c
@@ -1392,7 +1392,7 @@ __evict_walk_file(WT_SESSION_IMPL *session, uint32_t queue_index, u_int *slotp)
 		/* Limit internal pages to 50% unless we get aggressive. */
 		if (WT_PAGE_IS_INTERNAL(page) &&
 		    !FLD_ISSET(cache->state, WT_EVICT_PASS_AGGRESSIVE) &&
-		    internal_pages >= (int)(evict - start) / 2)
+		    internal_pages >= (int)(end - start) / 2)
 			continue;
 
 fast:		/* If the page can't be evicted, give up. */


src/third_party/wiredtiger/src/evict/evict_lru.c
```diff
@@ -1392,7 +1392,7 @@ __evict_walk_file(WT_SESSION_IMPL *session, uint32_t queue_index, u_int *slotp)
 		/* Limit internal pages to 50% unless we get aggressive. */
 		if (WT_PAGE_IS_INTERNAL(page) &&
 		    !FLD_ISSET(cache->state, WT_EVICT_PASS_AGGRESSIVE) &&
-		    internal_pages >= (int)(evict - start) / 2)
+		    internal_pages >= (int)(end - start) / 2)
```
michaelcahill: @SunguckLee, I'm very concerned about making this change in general: I think there must be something going wrong elsewhere when WiredTiger is running your workload if there are no leaf pages in cache.

We could change it to this condition to make sure at least one internal page is considered:

```c
 		if (WT_PAGE_IS_INTERNAL(page) &&
  		    !FLD_ISSET(cache->state, WT_EVICT_PASS_AGGRESSIVE) &&
		    internal_pages > (int)(evict - start) / 2)
```

Can you check if that helps?

SunguckLee:
I don't know much things about Wiredtiger. So I don't know why this case is happening.
Anyway I am sure there's a lot of leaf pages in wiredtiger cache. (We can see that in wiredtiger metric graphs). And they(leaf pages) are also evicted a lot. 

I applied your suggestion to our service, And still looks good and internal page evicted about 3K/seconds out of total 15K/seconds eviction (it's been about 10 minutes).

I will comment about this in 5~6 hours.
Thanks.

SunguckLee:
Hi Michael Cahill.

First of all, I am sorry. I was wrong. 
Actually the pull-request I've changed has nothing to do with internal node eviction throughput and user request blocking.

I have changed 3 parts of mongodb source code (You might already know all of them).
I have applied each part to mongodb 3.2.6(with WT-2560) and checked mongodb processing status.

*Part-1* : No effect, still the eviction of internal pages is so small(0~50 pages/sec) and wiredtiger cache usage is going up 100%, and user write request is queued up.
*Part-2* : Eviction throughput of Internal page is increased (5K~8K pages/sec) but wiredtiger cache usage is going up 100%. And write requests are queued up and never resolved (all query processing is stopped).
*Part-3* : I applied 3 changes together, Eviction throughput is increased and but sometimes cache usage is going up to 100% but cache usage get back to 80% around. (service user requests are queued Message queue first, and worker processor doing mongodb update, so right after mongodb start, there's a lot of DMLs and Queries are requested, Now user reuqests are stable and mongodb also get back to stable status).

I wish this could be help to solve this issue.
And tell me anything if you need.

*Part-1*
{code}
__evict_walk_file(WT_SESSION_IMPL *session, uint32_t queue_index, u_int *slotp){
	...
        if (WT_PAGE_IS_INTERNAL(page) &&
            !FLD_ISSET(cache->state, WT_EVICT_PASS_AGGRESSIVE) &&
            internal_pages > (int)(evict - start) / 2)
    ...
}
{code}

*Part-2*
{code}
static int __evict_lru_walk(WT_SESSION_IMPL *session){
	...
                else {
                        /* Save the calculated oldest generation. */
                        cache->read_gen_oldest = read_gen_oldest;

                        /* Find the bottom 25% of read generations. */
                        //cutoff =
                        //    (3 * read_gen_oldest + __evict_read_gen(
                        //    &cache->evict_queue[entries - 1])) / 4;

                        /*
                         * Don't take less than 10% or more than 50% of entries,
                         * regardless. That said, if there is only one entry,
                         * which is normal when populating an empty file, don't
                         * exclude it.
                         */
                        //for (candidates = 1 + entries / 10;
                        //    candidates < entries / 2;
                        //    candidates++)
                        //      if (__evict_read_gen(
                        //          &cache->evict_queue[candidates]) > cutoff)
                        //              break;
                        //cache->evict_candidates = candidates;

                        /* ADDED BY MATT */
                        WT_CACHE *cache = S2C(session)->cache;
                        uint64_t bytes_max = S2C(session)->cache_size + 1;
                        uint64_t bytes_inuse = __wt_cache_bytes_inuse(cache);
                        float cache_usage_per_target = (bytes_inuse / cache->eviction_target * bytes_max);
                        if(cache_usage_per_target>=1) cache_usage_per_target = 1.0;

                        cache->evict_candidates = entries * cache_usage_per_target;
                        /* ADDED BY MATT */
                }
	...
}
{code}

*Part-3*
{code}
void __wt_txn_update_oldest(WT_SESSION_IMPL *session, bool force){
 	...
        /*
         * For pure read-only workloads, or if the update isn't forced and the
         * oldest ID isn't too far behind, avoid scanning.
         */
        if (prev_oldest_id == current_id ||
            (!force && WT_TXNID_LT(current_id, prev_oldest_id + 100)))
                return;

 		/* ADDED BY MATT */
        // If there's over two threads updating txn_global->oldest_id, then skip it in this thread
        if(txn_global->scan_count>2){
                return;
        }
        /* ADDED BY MATT */
    ...
}
{code}

SunguckLee:
Hi Michael Cahill.

According to the source code, All threads(both eviction worker or user thread) evicting modified page have to call "__wt_txn_update_oldest()" to update oldest_id and last_running. __wt_txn_update_oldest have to iterate all sessions to refresh oldest_id and last_running. Two variables are locked down during this iteration. I think it's a little bit heavy to run for each page.

Do you think my change(part-3 in above comment) is harm to protect consistent data ?
And do you have any progress about this issue ?

Thanks.

michaelcahill:
Discussion has moved to https://jira.mongodb.org/browse/SERVER-23778, if there are WiredTiger changes as a result, they will first go into https://github.com/wiredtiger/wiredtiger.

PR #1082, opened 2016-05-17
SERVER-23398 Fixed. Allowing read sharding data from config replica

danunafig:
Allowing read sharding data from any config replica nodes rather than Primary only.


ramonfm:
Thanks for taking the time to do a pull request. Unfortunately [this is not a viable approach as per Andy's explanation](https://jira.mongodb.org/browse/SERVER-23398?focusedCommentId=1266673&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1266673), so I'm going to close it for now.

Thanks,
Ramón.

PR #1085, opened 2016-05-23
SERVER-12489: Make shell warn if version does not match server's

joannac-:


markbenvenuto:
@joannac-  Thanks for the contribution!

PR #1093, opened 2016-06-15
SERVER-24519 Added framework for micro benchmark tools

danharel:


src/mongo/db/exec/SConscript
```diff
@@ -148,3 +148,20 @@ env.CppUnitTest(
     ],
     NO_CRUTCH = True,
 )
+
+# env.CppUnitTest(
```
RedBeard0531: delete commented out code

src/mongo/db/exec/SConscript
```diff
@@ -148,3 +148,20 @@ env.CppUnitTest(
...
+#         "benchmark_test.cpp"
+#     ],
+#     LIBDEPS = [],
+# )
+
+
+env.CppBenchmark(
+    target = "benchmark_test",
+    source = [
+        "benchmark_test.cpp"
```
RedBeard0531: Lets move this file to src/mongo/benchmarks and rename it map_bench.cpp. You'll need to create a new Sconscript file and add it to the list of Sconscripts in src/mongo/sconscript

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include "mongo/db/jsobj.h"
+#include "mongo/db/json.h"
+#include "mongo/db/storage/snapshot.h"
+#include "mongo/platform/random.h"
+#include "mongo/unittest/unittest.h"
+#include "mongo/util/assert_util.h"
+#include "mongo/util/string_map.h"
+#include "mongo/util/time_support.h"
+
+#include <unordered_map>
```
RedBeard0531: non-mongo headers go above mongo headers (local convention). Also the first include in cpp files should be "mongo/platform/basic.h"

https://github.com/mongodb/mongo/wiki/Server-Code-Style#for-cpp-files

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+ *    all of the code used other than as permitted herein. If you modify file(s)
+ *    with this exception, you may extend this exception to your version of the
+ *    file(s), but you are not obligated to do so. If you do not wish to do so,
+ *    delete this exception statement from your version. If you delete this
+ *    exception statement from all source files in the program, then also delete
+ *    it in the license file.
+ */
+
+/**
+ * This file contains a sample usage of TEST_F for benchmarking
```
RedBeard0531: Try to end comments with a period (even if not a full sentence) so it is clear that you were finished.

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+    std::vector<const char*> strings;
+    for (int i = 0; i < TRIALS * 2; i++) {
+        strings.push_back(stdStringList[i].c_str());
+    }
+    return strings;
+})();
+
+// http://stackoverflow.com/questions/440133/how-do-i-create-a-random-alpha-numeric-string-in-c
+std::string gen_random(int length) {
+    static string charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";
```
RedBeard0531: use const char charset[] rather than string. There is no need to heap-allocate this.

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+const auto cStringList = ([]() -> std::vector<const char*> {
+    std::vector<const char*> strings;
+    for (int i = 0; i < TRIALS * 2; i++) {
+        strings.push_back(stdStringList[i].c_str());
+    }
+    return strings;
+})();
+
+// http://stackoverflow.com/questions/440133/how-do-i-create-a-random-alpha-numeric-string-in-c
+std::string gen_random(int length) {
```
RedBeard0531: Use size_t for lengths (with very few exceptions)

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include "mongo/util/string_map.h"
+#include "mongo/util/time_support.h"
+
+#include <unordered_map>
+
+#define TRIALS 100000
+#define STR_LEN 32
+#define OUTFILE "benchmarks.json"
+
+namespace mongo {
```
RedBeard0531: Everything in this file should be in an anon namespace inside the mongo namespace like:
namespace mongo{
namespace {
...
} // namespace
} // namespace mongo

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include "mongo/db/storage/snapshot.h"
+#include "mongo/platform/random.h"
+#include "mongo/unittest/unittest.h"
+#include "mongo/util/assert_util.h"
+#include "mongo/util/string_map.h"
+#include "mongo/util/time_support.h"
+
+#include <unordered_map>
+
+#define TRIALS 100000
```
RedBeard0531: Prefer real constants to #defines.

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+#define STR_LEN 32
+#define OUTFILE "benchmarks.json"
+
+namespace mongo {
+
+using std::string;
+
+std::string gen_random(const int);
```
RedBeard0531: Can you just move the definition up here?

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+    return strings;
+})();
+
+// http://stackoverflow.com/questions/440133/how-do-i-create-a-random-alpha-numeric-string-in-c
+std::string gen_random(int length) {
+    static string charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";
+    string result;
+    result.resize(length);
+    for (int i = 0; i < length; i++) {
+        result[i] = charset[rand->nextInt32(charset.length())];
```
RedBeard0531: It turns out that the operator[] in std::string can have a surprising amount of overhead. It would be better to use a range-for loop like for (char& c : result) { c = something(); }

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    return result;
+}
+
+class BenchmarkMapInsertFixture : public mongo::unittest::Benchmark {
```
RedBeard0531: You could cut out some duplication by using a template here:
template <typename Map>
class BenchmarkFixtureBase : public mongo::unittest::Benchmark {
  Map ints;
  typename Map::const_iterator it;
};

using BenchmarkMapInsertFixture =  BenchmarkFixtureBase<std::map<std::string, int>>;

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+protected:
+    std::map<std::string, int> ints;
+    std::map<std::string, int>::const_iterator it;
+
+    void setUp() {}
+};
+
+class BenchmarkMapLookupFixture : public mongo::unittest::Benchmark {
```
RedBeard0531: This can also be boilerplate reduced:
```
template <typename BaseFixture>
class FillingFixtureBase : public BaseFixture {
    void setUp() {
        // Insert a bunch of strings into the map
        int i = 0;
        for (const std::string& str : stdStringList) {
            ints[str] = i++;
        }
    }
};

using MapLookupFixture =  FillingFixtureBase<FixtureBase<std::map<std::string, int>>>;
```

src/mongo/executor/SConscript
```diff
@@ -337,3 +337,30 @@ env.Library(
         'task_executor_interface',
     ],
 )
+
+
+# env.CppBenchmark(
```
RedBeard0531: delete commented out code

etc/perf.yml
```diff
@@ -164,6 +171,19 @@ functions:
         name: "perf"
         file: "src/perf/perf.json"
 
+  ## Runs the benchmarks and output to perf.json in order to display the results on Mongo Perf
```
RedBeard0531: Can you leave a comment like TODO run all microbenchmarks

etc/perf.yml
```diff
@@ -641,6 +672,14 @@ tasks:
         variant1 : "linux-mmap-standalone"
         variant2 : "linux-mmap-repl"
     - func: "analyze"
+- name: benchmark
```
RedBeard0531: Lets rename this cppMicroBenchmarks. I know it's silly, but that will cause it to sort immediately after compile which should the order to be nicer when viewed on evergreen.

site_scons/site_tools/mongo_microbenchmarks.py
```diff
@@ -0,0 +1,51 @@
...
+            print '\t' + str(s)
+            ofile.write('%s\n' % s)
+    finally:
+        ofile.close()
+
+def build_cpp_benchmark(env, target, source, **kwargs):
+    libdeps = kwargs.get('LIBDEPS', [])
+    libdeps.append( '$BUILD_DIR/mongo/unittest/benchmarks_main' )
+
+    # # Delete this if possible
```
RedBeard0531: delete commented out code

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
+/**
+ *    Copyright (C) 2013 10gen Inc.
```
RedBeard0531: When writing a new file, make sure the copyright is 2016 MongoDB Inc.

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include "mongo/unittest/unittest.h"
+#include "mongo/util/assert_util.h"
+#include "mongo/util/string_map.h"
+#include "mongo/util/time_support.h"
+
+#include <unordered_map>
+
+#define TRIALS 100000
+#define STR_LEN 32
+#define OUTFILE "benchmarks.json"
```
RedBeard0531: Is this actually being used?

src/mongo/unittest/benchmark_options.h
```diff
@@ -0,0 +1,57 @@
...
+ * file(s) with this exception, you may extend this exception to your
+ * version of the file(s), but you are not obligated to do so. If you do not
+ * wish to do so, delete this exception statement from your version. If you
+ * delete this exception statement from all source files in the program,
+ * then also delete it in the license file.
+ */
+
+#pragma once
+
+#include <fstream>
```
RedBeard0531: The first two includes aren't used in this file.

src/mongo/unittest/benchmarks_main.cpp
```diff
@@ -0,0 +1,72 @@
+// mongo/unittest/unittest_main.cpp
```
RedBeard0531: New files don't need this.

src/mongo/unittest/benchmark_options.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Global struct to store any relevant global information
+struct BenchmarkGlobalParams {
+    // File to write results to
+    std::string outFileName;
+};
+
+extern BenchmarkGlobalParams benchmarkGlobalParams;
+
+Status addSASLSSPIOptions(moe::OptionSection* options);
```
RedBeard0531: delete these, and the forward declares above.

src/mongo/unittest/benchmark_options.h
```diff
@@ -0,0 +1,57 @@
...
+class Status;
+
+namespace optionenvironment {
+class OptionSection;
+class Environment;
+}  // namespace optionenvironment
+
+namespace moe = mongo::optionenvironment;
+
+// Global struct to store any relevant global information
```
RedBeard0531: use /** doxygen style */ comments when documenting public interfaces.

src/mongo/unittest/benchmarks_main.cpp
```diff
@@ -0,0 +1,72 @@
...
+ *    must comply with the GNU Affero General Public License in all respects
+ *    for all of the code used other than as permitted herein. If you modify
+ *    file(s) with this exception, you may extend this exception to your
+ *    version of the file(s), but you are not obligated to do so. If you do not
+ *    wish to do so, delete this exception statement from your version. If you
+ *    delete this exception statement from all source files in the program,
+ *    then also delete it in the license file.
+ */
+
+#include <cstdio>
```
RedBeard0531: Are all of these includes needed?

src/mongo/unittest/unittest.cpp
```diff
@@ -32,11 +32,13 @@
 
 #include "mongo/unittest/unittest.h"
 
-#include <iostream>
+//#include <fstream>
 #include <map>
+#include <stdio.h>
```
RedBeard0531: use <cstdio> in c++. All of the c stdlib headers have c++ equivalents that should be used instead.

src/mongo/unittest/unittest.cpp
```diff
@@ -159,6 +169,56 @@ void Test::run() {
 void Test::setUp() {}
 void Test::tearDown() {}
 
+//
+void Benchmark::run() {
+
```
RedBeard0531: kill the blank line here.

src/mongo/unittest/unittest.h
```diff
@@ -368,6 +411,42 @@ class Test {
     logger::MessageLogDomain::AppenderAutoPtr _captureAppender;
 };
 
+class Benchmark : public Test {
+protected:
+    int trial_num = 0;
+
+public:
+    /**
+     * Called to prepare the benchmark for execution
```
RedBeard0531: Prefer to phase the first sentence of method comments like "Prepares the ..."

They should say what the function does, not why you would call it. If further explanation is needed (which should generally be rare),  do so after the first sentence.

src/mongo/unittest/unittest.h
```diff
@@ -231,6 +234,39 @@
...
+ * protected:
+ *   int myVar;
+ *   void setUp() { myVar = 10; }
+ * };
+ *
+ * BENCHMARK(FixtureClass, TestThatUsesFixture) {
+ *     myVar += 1;
+ * }
+ */
+#define BENCHMARK(FIXTURE_NAME, BENCHMARK_NAME, ITERATIONS, FILE)                              \
```
RedBeard0531: File is unused

src/mongo/unittest/unittest.h
```diff
@@ -240,6 +276,13 @@ namespace mongo {
 
 namespace unittest {
 
+typedef struct benchmark_stats {
```
RedBeard0531: Just do struct BenchmarkStats {...};

src/mongo/unittest/unittest_main.cpp
```diff
@@ -34,6 +34,9 @@
 #include "mongo/unittest/unittest.h"
 #include "mongo/util/signal_handlers_synchronous.h"
 
+#include "mongo/util/options_parser/startup_option_init.h"
```
RedBeard0531: why?

src/mongo/unittest/unittest.h
```diff
@@ -231,6 +234,39 @@
...
+ */
+#define BENCHMARK(FIXTURE_NAME, BENCHMARK_NAME, ITERATIONS, FILE)                              \
+    class _TEST_TYPE_NAME(FIXTURE_NAME, BENCHMARK_NAME) : public FIXTURE_NAME {                \
+    private:                                                                                   \
+        NOINLINE_DECL virtual void _doTest();                                                  \
+        int numIterations() {                                                                  \
+            return ITERATIONS;                                                                 \
+        }                                                                                      \
+        std::string testName() {                                                               \
+            return QUOTE(BENCHMARK_NAME);                                                      \
```
RedBeard0531: Any reason you can't just use #BENCHMARK_NAME here? If you can, delete QUOTE above.

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+namespace mongo {
+
+using std::string;
+
+std::string gen_random(const int);
+
+PseudoRandom* rand = new PseudoRandom((uint32_t)time(NULL));
```
RedBeard0531: Does this need to be heap-allocated? Just do PseudoRandom rand(0xSOMETHING);

You also want the results to be predictable, which means you should use the same seed every time.

src/mongo/db/exec/benchmark_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+const auto stdStringList = ([]() -> std::vector<std::string> {
+    std::vector<std::string> strings;
+    for (int i = 0; i < TRIALS * 2; i++) {
```
RedBeard0531: // Why TRIALS*2?

src/mongo/unittest/unittest.cpp
```diff
@@ -45,19 +47,27 @@
 #include "mongo/stdx/functional.h"
 #include "mongo/stdx/memory.h"
 #include "mongo/stdx/mutex.h"
+#include "mongo/unittest/benchmark_options.h"
 #include "mongo/util/assert_util.h"
 #include "mongo/util/log.h"
+#include "mongo/util/print.h"
 #include "mongo/util/timer.h"
 
+#define RUNS 5
```
RedBeard0531: Use a real constant

src/mongo/unittest/unittest.cpp
```diff
@@ -159,6 +169,56 @@ void Test::run() {
...
+        Date_t start_time = Date_t::now();
+        // An uncaught exception does not prevent the tear down from running. But
+        // such an event still constitutes an error. To test this behavior we use a
+        // special exception here that when thrown does trigger the tear down but is
+        // not considered an error.
+        try {
+            for (trial_num = 0; trial_num < numIterations(); trial_num++) {
+                _doTest();
+            }
+        } catch (FixtureExceptionForTesting&) {
```
RedBeard0531: Since this is a benchmark, not a unittest, I don't think these exceptions should be caught. Instead I'd just catch std::exception and log what(). You should also set error_values to 1 for this benchmark.

src/mongo/unittest/unittest.h
```diff
@@ -368,6 +411,42 @@ class Test {
...
 
+class Benchmark : public Test {
+protected:
+    int trial_num = 0;
+
+public:
+    /**
+     * Called to prepare the benchmark for execution
+     */
+    virtual void run();
```
RedBeard0531: I'd declare this like virtual Stats run() final;

We don't want fixtures to override Benchmark::run(), even though they are allowed to override Test::run(). Also by returning Stats, we don't need to add it to a global list which is generally in poor taste.

src/mongo/unittest/unittest.cpp
```diff
@@ -386,6 +447,150 @@ int Suite::run(const std::vector<std::string>& suites, const std::string& filter
...
+              << " suites failed";
+    } else {
+        log() << "SUCCESS - All tests in all suites passed";
+    }
+
+    // Start building the JSON output using BSONObjBuilder
+    BSONObjBuilder b;
+
+    // TODO: Get the start and end time
+    b.append("start", "PLACEHOLDER");
```
RedBeard0531: Fill in the PLACEHOLDERS. You can use dateToISOStringUTC()

src/mongo/unittest/unittest.cpp
```diff
@@ -386,6 +447,150 @@ int Suite::run(const std::vector<std::string>& suites, const std::string& filter
...
+    b.append("end", "PLACEHOLDER");
+
+    // No errors
+    BSONArrayBuilder errorsList(b.subarrayStart("errors"));
+    errorsList.doneFast();
+
+    BSONArrayBuilder resultsList(b.subarrayStart("results"));
+    // Construct the results list by pulling the data from the static BenchmarkStats struct
+    for (BenchmarkStats stat : statsList) {
```
RedBeard0531: const&. It is also common to use auto in loops like for (const auto& thing : collection) {...} but feel free to include the type if you think it aides readability.

etc/perf.yml
```diff
@@ -641,6 +672,14 @@ tasks:
...
+  depends_on:
+  - variant: linux-wt-standalone
+    name: compile
+  commands:
+    # Commands to run to get the thing going
+    - func: "start server"
```
RedBeard0531: Do you need this?

src/mongo/unittest/unittest.cpp
```diff
@@ -386,6 +447,150 @@ int Suite::run(const std::vector<std::string>& suites, const std::string& filter
...
+        BSONObjBuilder statObj;
+        statObj.append("name", stat.name);
+        BSONObjBuilder benchResults(statObj.subobjStart("results"));
```
RedBeard0531: It is helpful to create a nested scope when building inner BSONObjects. It makes the shape of the code match the shape of the Object.

src/mongo/unittest/unittest.cpp
```diff
@@ -386,6 +447,150 @@ int Suite::run(const std::vector<std::string>& suites, const std::string& filter
...
+        benchResults.doneFast();
+
+        resultsList.append(statObj.done());
+    }
+    resultsList.doneFast();
+
+    // write to outFile or cout
+    std::string json = b.done().jsonString();
+    if (benchmarkGlobalParams.outFileName == "") {
+        std::cout << json << std::endl;
```
RedBeard0531: If you are going to do this, you should set the pretty arg to 1 in jsonString() so the output is readable.

src/mongo/unittest/unittest.cpp
```diff
@@ -296,6 +356,7 @@ Result* Suite::run(const std::string& filter, int runsPerTest) {
 }
 
 int Suite::run(const std::vector<std::string>& suites, const std::string& filter, int runsPerTest) {
+
```
RedBeard0531: ??

src/mongo/unittest/unittest.cpp
```diff
@@ -159,6 +169,56 @@ void Test::run() {
...
+        } catch (FixtureExceptionForTesting&) {
+            tearDown();
+            return;
+        } catch (TestAssertionFailureException&) {
+            tearDown();
+            throw;
+        }
+
+        Milliseconds elapsed_time = Date_t::now() - start_time;
+        // Do I leave this line or nah?
```
RedBeard0531: I don't think so.

src/mongo/unittest/unittest.cpp
```diff
@@ -159,6 +169,56 @@ void Test::run() {
...
+        // std::cout << "** ELAPSED_TIME -- " << elapsed_time << " **" << std::endl;
+
+        float opsPerSec = 1.0 * numIterations() / elapsed_time.count();
```
RedBeard0531: You are producing ops/millisecond.

src/mongo/unittest/unittest.cpp
```diff
@@ -159,6 +169,56 @@ void Test::run() {
...
+    BenchmarkStats stats = BenchmarkStats();
+    stats.name = testName();
+
+    float totalOpsPerSec = 0;
+
+    for (int r = 0; r < RUNS; r++) {
+
+        setUp();
+
+        Date_t start_time = Date_t::now();
```
RedBeard0531: Use camelCase thoughout. Please check everything in this commit and fix anything that isn't named correctly.

kelly-cs:
This work was completed in 2018, closing this PR

PR #1094, opened 2016-06-16
SERVER-24555 made compilation_db.py compatible with scons 2.4

purpleladydragons:


acmorrow:
LGTM if you have tested with both. Thanks!

acmorrow:
PS remember that we don't use a merge workflow for the server. Please rebase and push from the CLI, do not use the merge button here.

tychoish:
merged this.

PR #1099, opened 2016-07-06
Allow specifying sample rate of slow queries

ebroder:
Ideally, we'd like to be able to set a lower slowms threshold so that we can get a broader sample of queries being run against our databases. However, dropping the threshold low enough to get useful data results in an unacceptable amount of load.

This allows us to drop the threshold, but only capture a fraction of queries.

(We're currently running with a similar patch backported to an older MongoDB version in production, and it's been very helpful for analyzing query traffic)

### DIFF
--- a/jstests/core/profile_sampling.js
+++ b/jstests/core/profile_sampling.js
@@ -1,0 +1,49 @@
+// Confirms that the number of profiled operations is consistent with the sampleRate, if set.
+
+(function () {
+    "use strict";
+
+    // Use a special db to support running other tests in parallel.
+    var profileDB = db.getSisterDB("profile_sampling");
+    var username = "jstests_profile_sampling_user";
+
+    var coll = profileDB.profile_sampling;
+
+    profileDB.dropDatabase();
+
+    var origprof = 0;
+    try {
+        origprof = profileDB.setProfilingLevel(0).was;
+        profileDB.system.profile.drop();
+        assert.eq(0, profileDB.system.profile.count());
+
+        profileDB.createCollection(coll.getName());
+        assert.writeOK(coll.insert({x: 1}));
+
+        assert.commandWorked(profileDB.setProfilingLevel(2, {sampleRate: 0}));
+
+        assert.neq(null, coll.findOne({x: 1}));
+        assert.eq(1, coll.find({x: 1}).count());
+        assert.writeOK(coll.update({x: 1}, {$inc: {a: 1}}));
+
+        profileDB.setProfilingLevel(0);
+
+        profileDB.system.profile.find().forEach(printjson);
+        assert.eq(0, profileDB.system.profile.count());
+
+        profileDB.system.profile.drop();
+        assert.commandWorked(profileDB.setProfilingLevel(2, {sampleRate: 0.5}));
+
+        // This should generate about 500 profile log entries.
+        for (var i = 0; i < 500; i++) {
+            assert.neq(null, coll.findOne({x: 1}));
+            assert.writeOK(coll.update({x: 1}, {$inc: {a: 1}}));
+        }
+
+        profileDB.setProfilingLevel(0);
+
+        assert.between(10, profileDB.system.profile.count(), 990);
+    } finally {
+        profileDB.setProfilingLevel(origprof);
+    }
+}());
--- a/src/mongo/db/commands/dbcommands.cpp
+++ b/src/mongo/db/commands/dbcommands.cpp
@@ -39,6 +39,7 @@
 #include "mongo/base/status_with.h"
 #include "mongo/bson/simple_bsonobj_comparator.h"
 #include "mongo/bson/util/builder.h"
+#include "mongo/bson/util/bson_extract.h"
 #include "mongo/db/audit.h"
 #include "mongo/db/auth/action_set.h"
 #include "mongo/db/auth/action_type.h"
@@ -349,7 +350,7 @@ public:
                                        const BSONObj& cmdObj) {
         AuthorizationSession* authzSession = AuthorizationSession::get(client);
 
-        if (cmdObj.firstElement().numberInt() == -1 && !cmdObj.hasField("slowms")) {
+        if (cmdObj.firstElement().numberInt() == -1 && !cmdObj.hasField("slowms") && !cmdObj.hasField("sampleRate")) {
             // If you just want to get the current profiling level you can do so with just
             // read access to system.profile, even if you can't change the profiling level.
             if (authzSession->isAuthorizedForActionsOnResource(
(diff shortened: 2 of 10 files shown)


ramonfm:
Thanks for your pull request @ebroder. Before we can consider it, please see the the [information on contributing](https://github.com/mongodb/mongo/blob/master/CONTRIBUTING.rst) as you'll need to create a JIRA ticket in the SERVER project and [sign the contributor agreement](https://www.mongodb.com/legal/contributor-agreement).

Thanks,
Ramón.

ebroder:
Hi Ramón - thanks for the quick reply. I've gone ahead and signed the CLA. In terms of tickets, I think that this can arguably be viewed as a fix for [SERVER-4786](https://jira.mongodb.org/browse/SERVER-4786) (with a slightly different interface). However, I can open a new ticket if that's preferable.

ebroder:
Folks - any chance of getting this looked at?

ramonfm:
Apologies for the radio silence @ebroder. I've linked your pull request to [SERVER-4786](https://jira.mongodb.org/browse/SERVER-4786) and asked for a reviewer.

dstorch:
Hi @ebroder,

Thanks for taking the time to put together this pull request! I think a sampling mode for the system profiler would be an immensely useful feature. Unfortunately, we have recently entered our stabilization period in preparation for the release of MongoDB 3.4. We our doing our best to freeze new feature development in order to focus on quality and performance, and to ensure that downstream work (such as that required for the documentation and drivers) can be completed. In particular, the system.profile collection is consumed by downstream tooling, so there is some possibility that those tools would have to be aware of this change in order to fully support version 3.4 of the server.

Based on the above, I would like to table this change until after we branch for 3.4, at which point I would be happy to evaluate this for inclusion in MongoDB 3.6. Please let me know if you have any questions.

Best,
Dave

dstorch:
Hi @ebroder,

I'm currently carrying your proposed change through our design review process. Stay tuned!

Best,
Dave

src/mongo/db/mongod_options.cpp
```diff
@@ -150,6 +150,13 @@ Status addMongodOptions(moe::OptionSection* options) {
                            "value of slow for profile and console log")
         .setDefault(moe::Value(100));
 
+    general_options
+        .addOptionChaining("operationProfiling.slowOpSampleRate",
+                           "samplerate",
```
dstorch: Please make the command line parameter camelCase as well.

src/mongo/db/commands/dbcommands.cpp
```diff
@@ -316,7 +316,7 @@ class CmdProfile : public Command {
                                        const BSONObj& cmdObj) {
         AuthorizationSession* authzSession = AuthorizationSession::get(client);
 
-        if (cmdObj.firstElement().numberInt() == -1 && !cmdObj.hasField("slowms")) {
+        if (cmdObj.firstElement().numberInt() == -1 && !cmdObj.hasField("slowms") && !cmdObj.hasField("samplerate")) {
```
dstorch: I'd like to see a test proving that if you include sampleRate in the profile command, then enableProfiler privileges are required. This is probably easiest to accomplish by adding a new test case to jstests/auth/lib/commands_lib.js. You can see examples of the other profile command authorization tests here:

[https://github.com/mongodb/mongo/blob/r3.5.1/jstests/auth/lib/commands_lib.js#L3515-L3573](url)

evan-stripe:
That was an incredibly helpful review! I didn't find it nitpicky at all.

I've rebased, fixed up (I believe) all the issues you called out, and included some test cases both for permissions and general behavior.

I did re-discover in testing something we discovered internally recently, which is that the write commands have a separate code path for determining whether to log and profile operations than (as far as I can tell) everything else. I added logic to `finishCurOp` in `src/mongo/db/ops/write_ops_exec.cpp`. I'm not wild about the duplication of logic, but I don't feel like I understand the codebase well enough to know where else to put it. If you think there's a better option, open to suggestions.

Just to make sure, I looked through anything else that calls `curOp->shouldDBProfile()` to make sure there wasn't anything else that didn't seem like it would get caught by this, and all of the other callsites seem to just be retaining the query plan for later use by the profiler.

ebroder:
That was an incredibly helpful review! I didn't find it nitpicky at all.

I've rebased, fixed up (I believe) all the issues you called out, and included some test cases both for permissions and general behavior.

I did re-discover in testing something we discovered internally recently, which is that the write commands have a separate code path for determining whether to log and profile operations than (as far as I can tell) everything else. I added logic to `finishCurOp` in `src/mongo/db/ops/write_ops_exec.cpp`. I'm not wild about the duplication of logic, but I don't feel like I understand the codebase well enough to know where else to put it. If you think there's a better option, open to suggestions.

Just to make sure, I looked through anything else that calls `curOp->shouldDBProfile()` to make sure there wasn't anything else that didn't seem like it would get caught by this, and all of the other callsites seem to just be retaining the query plan for later use by the profiler.

jstests/core/profile_sampling.js
```diff
@@ -0,0 +1,48 @@
+// special db for running tests in parallel
+var stddb = db;
```
evan-stripe: I was copying the pattern in, e.g., `jstests/core/profile1.js`, which restores the value of `db` at the end of the file. Is that not necessary?

dstorch: Oooh, I see, this is so that `db` is rebound to the "test" database after this jstest completes. Instead of doing this, it's probably better to do something like

`var profileDB = db.getSisterDB("profile_sampling");`

Then you can use `profileDB` throughout the test instead of `db`. That way we're not updating the state of `db`, which other tests may depend on.

jstests/core/profile_sampling.js
```diff
@@ -0,0 +1,48 @@
...
+    db.auth(username, 'password');
+
+    db.setProfilingLevel(0);
+    db.system.profile.drop();
+    assert.eq(0, db.system.profile.count());
+
+    db.createCollection(t.getName());
+    t.insert({x: 1});
+
+    assert.commandWorked(db.runCommand({profile: 2, sampleRate: 0}));
```
ebroder: Should `options` be a 3rd option, or should I coerce the 2nd argument based on whether it's an object or a numeric?

dstorch: Good question. Both seem reasonable. I'll let you make the call on which you think seems best. I'm not a huge fan of arguments that could mean two things, but the advantage is that it prevents users from having to do something like `db.setProfilingLevel(2, undefined, {sampleRate: 0.5});`.

ebroder:
Great, cleaned up the tests, and updated `setProfilingLevel`. I agree that the type-based dispatch is a little awkward, but I think the new interface is probably a better one long-term.

dstorch:
Looks great. I'm going to rebase this locally, run it through our continuous integration system, and merge. Thanks for contributing.

PR #1100, opened 2016-07-11
SERVER-24948: add nameOnly option to listDatabases

zydcom:
Add 'nameOnly' option to listDatabases command, using this option in initial sync to avoid socket timeout when listDatabases.

### DIFF
--- a/jstests/core/dbadmin.js
+++ b/jstests/core/dbadmin.js
@@ -10,6 +10,9 @@ t.save( { x : 1 } );
 var res = db._adminCommand( "listDatabases" );
 assert( res.databases && res.databases.length > 0 , "listDatabases 1 " + tojson(res) );
 
+var res = db.adminCommand( {listDatabases: 1, nameOnly: true} );
+assert( res.databases && res.databases.length > 0 && res.nameOnly , "listDatabases 1 " + tojson(res) );
+
 var now = new Date();
 var x = db._adminCommand( "ismaster" );
 assert( x.ismaster , "ismaster failed: " + tojson( x ) );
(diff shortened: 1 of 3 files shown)


onyxmaster:
Since we're really hit by https://jira.mongodb.org/browse/SERVER-3181, I really would like this integrated.

jstests/core/dbadmin.js
```diff
@@ -10,6 +10,9 @@ t.save( { x : 1 } );
 var res = db._adminCommand( "listDatabases" );
 assert( res.databases && res.databases.length > 0 , "listDatabases 1 " + tojson(res) );
 
+var res = db.adminCommand( {listDatabases: 1, nameOnly: true} );
+assert( res.databases && res.databases.length > 0 && res.nameOnly , "listDatabases 1 " + tojson(res) );
```
GeertBosch: nit: The "listDatabases 1" message isn't very helpful, maybe do "listDatabases: 1, nameOnly: true" or just "listDatabases nameOnly".

src/mongo/db/commands/list_databases.cpp
```diff
@@ -104,7 +111,7 @@ class CmdListDatabases : public Command {
                 const DatabaseCatalogEntry* entry = db->getDatabaseCatalogEntry();
                 invariant(entry);
 
-                int64_t size = entry->sizeOnDisk(txn);
+                int64_t size = nameOnly ? 0 : entry->sizeOnDisk(txn);
```
GeertBosch: Rather than setting `size = 0` here, why not skip the entire block from line 103 to 119? Reporting `sizeOnDisk: 0` and `empty: true` would seem undesirable.

GeertBosch:
A more extensive patch to implement this functionality, and test it, has now been incorporated. Thank you for your request!

PR #1101, opened 2016-07-14
SERVER-19507 scan all keys in index for distinct key

Shekeen:
The distinct hack can also be applied when the distinct command key is not the first key in the index. The corresponding JIRA ticket gives one such example.

### DIFF
--- a/jstests/core/distinct_multikey_index.js
+++ b/jstests/core/distinct_multikey_index.js
@@ -3,30 +3,26 @@
 
     load("jstests/libs/analyze_plan.js");
 
-    var t = db.distinct_multikey_index;
+    var coll = db.distinct_multikey_index;
 
-    t.drop();
+    coll.drop();
     for (var i = 0; i < 10; i++) {
-        t.save({a: 1, b: 1});
-        t.save({a: 1, b: 2});
-        t.save({a: 2, b: 1});
-        t.save({a: 2, b: 3});
+        assert.writeOK(coll.save({a: 1, b: 1}));
+        assert.writeOK(coll.save({a: 1, b: 2}));
+        assert.writeOK(coll.save({a: 2, b: 1}));
+        assert.writeOK(coll.save({a: 2, b: 3}));
     }
-    t.createIndex({a: 1, b: 1});
+    coll.createIndex({a: 1, b: 1});
 
-    var explain = t.distinct('b', {a: 1}).explain("executionStats");
-    assert.commandWorked(explain);
-    assert(planHasStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN"));
-    assert(planHasState(explain.queryPlanner.winningPlan, "PROJECTION"));
-    assert.eq(2, explain.executionStats.nReturned);
+    var explain_distinct_with_query = coll.explain("executionStats").distinct('b', {a: 1});
+    assert.commandWorked(explain_distinct_with_query);
+    assert(planHasStage(explain_distinct_with_query.queryPlanner.winningPlan, "DISTINCT_SCAN"));
+    assert(planHasStage(explain_distinct_with_query.queryPlanner.winningPlan, "PROJECTION"));
+    assert.eq(2, explain_distinct_with_query.executionStats.nReturned);
 
-    var stage = getPlanStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN");
-    assert.eq({a: 1, b: 1}, stage.keyPattern);
-    assert.eq("a_1_b_1", stage.indexName);
-    assert.eq(false, stage.isMultiKey);
-    assert.eq(false, stage.isUnique);
-    assert.eq(false, stage.isSparse);
-    assert.eq(false, stage.isPartial);
-    assert.eq(1, stage.indexVersion);
-    assert("indexBounds" in stage);
+    var explain_distinct_without_query = coll.explain("executionStats").distinct('b');
+    assert.commandWorked(explain_distinct_without_query);
+    assert(planHasStage(explain_distinct_without_query.queryPlanner.winningPlan, "COLLSCAN"));
+    assert(!planHasStage(explain_distinct_without_query.queryPlanner.winningPlan, "DISTINCT_SCAN"));
+    assert.eq(40, explain_distinct_without_query.executionStats.nReturned);
 })();
--- a/jstests/core/distinct_multikey_index.js
+++ b/jstests/core/distinct_multikey_index.js
@@ -1,0 +1,32 @@
+(function() {
+    "use strict";
+
+    load("jstests/libs/analyze_plan.js");
+
+    var t = db.distinct_multikey_index;
+
+    t.drop();
+    for (var i = 0; i < 10; i++) {
+        t.save({a: 1, b: 1});
+        t.save({a: 1, b: 2});
+        t.save({a: 2, b: 1});
+        t.save({a: 2, b: 3});
+    }
+    t.createIndex({a: 1, b: 1});
+
+    var explain = t.distinct('b', {a: 1}).explain("executionStats");
+    assert.commandWorked(explain);
+    assert(planHasStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN"));
+    assert(planHasState(explain.queryPlanner.winningPlan, "PROJECTION"));
+    assert.eq(2, explain.executionStats.nReturned);
+
+    var stage = getPlanStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN");
+    assert.eq({a: 1, b: 1}, stage.keyPattern);
+    assert.eq("a_1_b_1", stage.indexName);
+    assert.eq(false, stage.isMultiKey);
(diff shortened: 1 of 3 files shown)


deividaspetraitis:
Any updates with this?

ramonfm:
Unfortunately I can't give you any additional information. This PR is in the queue to be evaluated, we'll post updates on [SERVER-19507](https://jira.mongodb.org/browse/SERVER-19507) when we have them.

Thanks,
Ramón.

src/mongo/db/query/get_executor.cpp
```diff
@@ -1389,9 +1387,7 @@ StatusWith<unique_ptr<PlanExecutor>> getExecutorDistinct(OperationContext* txn,
     while (ii.more()) {
         const IndexDescriptor* desc = ii.next();
         IndexCatalogEntry* ice = ii.catalogEntry(desc);
-        // The distinct hack can work if any field is in the index but it's not always clear
-        // if it's a win unless it's the first field.
-        if (desc->keyPattern().firstElement().fieldName() == parsedDistinct->getKey()) {
+        if (desc->keyPattern().hasField(parsedDistinct->getKey())) {
```
dstorch: The special distinct optimization implemented by getExecutorDistinct() has two cases:

1) The case where there is no query predicate, which is handled on the block that opens on line 1441, and

2) the case where this is a query predicate, which is handled after the aforementioned block closes.

This change looks to me like it will work for case 2 but not for case 1. Specifically, line 1450 is still setting the distinct field to 0, even though the distinct field might not be the leading field. I actually think that we should continue to disallow the DISTINCT_SCAN fast path over a trailing field in the index when there is no query predicate, since it is not obvious that this is a performance win. My suggestion about how to implement this is to leave line 1450 alone, but instead modify getDistinctNodeIndex() to reject indices whose leading field is not equal to the distinct field.

Does this suggestion make sense?

dstorch:
Hi @Shekeen,

Thanks very much for taking the time to implement this pull request! Apologies for the delay. I will be doing the code review and working with you to get your code merged.

In addition to the comment I already left on the diff, I would like to see tests written to demonstrate that the new functionality works correctly. The DISTINCT_SCAN stage was originally implemented such that it could support distinct over a trailing field of the index, but since this was never turned on, I doubt we have test coverage.

I think we should test this in a few ways. First, we should add a unit test case to https://github.com/mongodb/mongo/blob/master/src/mongo/dbtests/query_stage_distinct.cpp in order to demonstrate that DISTINCT_SCAN does the write thing when the field number is non-zero. Second, we should add a new integration test in the jstests/core/ directory. This test can use explain(), combined with the helpers in https://github.com/mongodb/mongo/blob/master/jstests/libs/analyze_plan.js, to assert that a distinct operation is using DISTINCT_SCAN as expected.

Let me know if you have any questions or concerns about all this.

Best,
Dave

src/mongo/db/query/get_executor.cpp
Shekeen: Yes, it does, I will fix that. I forgot about the case when the query is empty, sorry.

Shekeen:
Hi again, @dstorch! I've implemented a couple of simple tests, but I couldn't find how to run tests from `jstests` directory. Could you please help me with it?

dstorch:
Hi @Shekeen, great thanks! You can run a jstest using our smoke testing script. From the root directory of the mongo repository, that looks like this:

python buildscripts/resmoke.py path/to/file.js

You can also run the entire core suite of jstests like this

python buildscripts/resmoke.py --suites=core -j n

where "n" is the number of concurrent jobs to run. (I use -j 30 to make things go extra fast.) Thanks again for your contribution. I'll take a look at the new commits shortly.

src/mongo/db/query/get_executor.cpp
```diff
@@ -1137,6 +1137,10 @@ bool getDistinctNodeIndex(const std::vector<IndexEntry>& indices,
         if (indices[i].multikey && isDottedField) {
             continue;
         }
+        // Skip indices where the first key is not field
```
dstorch: Nitpick: we try to have our comments end in a period so that it's clear that the comment is complete.

src/mongo/dbtests/query_stage_distinct.cpp
```diff
@@ -236,6 +236,73 @@ class QueryStageDistinctMultiKey : public DistinctBase {
...
+        }
+
+        addIndex(BSON("a" << 1 << "b" << 1));
+
+        AutoGetCollectionForRead ctx(&_txn, ns());
+        Collection* coll = ctx.getCollection();
+
+        std::vector<IndexDescriptor*> indices;
+        coll->getIndexCatalog()->findIndexesByKeyPattern(&_txn, BSON("a" << 1 << "b" << 1), false, &indices);
+        verify(indices.size() == 1);
```
dstorch: Since this is a test, use ASSERT_EQ(indices.size(), 1), rather than verify(). Same for line 270, and elsewhere in this test.

src/mongo/dbtests/query_stage_distinct.cpp
```diff
@@ -236,6 +236,73 @@ class QueryStageDistinctMultiKey : public DistinctBase {
...
+
+        DistinctParams params;
+        params.descriptor = indices[0];
+        verify(params.descriptor);
+
+        params.direction = 1;
+        params.fieldNo = 1;
+        params.bounds.isSimpleRange = false;
+
+        OrderedIntervalList a_oil{"a"};
```
dstorch: Nitpick: We typically name local variables in our codebase with camelCase rather than using_underscores.

src/mongo/dbtests/query_stage_distinct.cpp
```diff
@@ -236,6 +236,73 @@ class QueryStageDistinctMultiKey : public DistinctBase {
...
+        WorkingSet ws;
+        DistinctScan distinct(&_txn, params, &ws);
+
+        WorkingSetID wsid;
+        PlanStage::StageState state;
+
+        std::vector<int> seen;
+
+        while (PlanStage::IS_EOF != (state = distinct.work(&wsid))) {
+            if (PlanStage::ADVANCED == state) {
```
dstorch: Add

ASSERT_NE(state, PlanStage::FAILURE);
ASSERT_NE(state, PlanStage::DEAD);

dstorch:
A colleague of mine also pointed out that we have some information on testing the server available here: https://github.com/mongodb/mongo/wiki/Test-The-Mongodb-Server

jstests/core/distinct_multikey_index.js
```diff
@@ -0,0 +1,32 @@
+(function() {
+    "use strict";
+
+    load("jstests/libs/analyze_plan.js");
+
+    var t = db.distinct_multikey_index;
```
dstorch: Nitpick: Rename "t" to "coll". I think "t" stands for "table", but MongoDB calls them "collections" :)

jstests/core/distinct_multikey_index.js
```diff
@@ -0,0 +1,32 @@
...
+
+    t.drop();
+    for (var i = 0; i < 10; i++) {
+        t.save({a: 1, b: 1});
```
dstorch: assert.writeOK(coll.save({a: 1, b: 1}));

Same for lines 11 through 13.

jstests/core/distinct_multikey_index.js
```diff
@@ -0,0 +1,32 @@
...
+    }
+    t.createIndex({a: 1, b: 1});
+
+    var explain = t.distinct('b', {a: 1}).explain("executionStats");
+    assert.commandWorked(explain);
+    assert(planHasStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN"));
+    assert(planHasState(explain.queryPlanner.winningPlan, "PROJECTION"));
+    assert.eq(2, explain.executionStats.nReturned);
+
+    var stage = getPlanStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN");
```
dstorch: I'd actually say go ahead and delete lines 24-31.

jstests/core/distinct_multikey_index.js
```diff
@@ -0,0 +1,32 @@
...
+        t.save({a: 2, b: 3});
+    }
+    t.createIndex({a: 1, b: 1});
+
+    var explain = t.distinct('b', {a: 1}).explain("executionStats");
+    assert.commandWorked(explain);
+    assert(planHasStage(explain.queryPlanner.winningPlan, "DISTINCT_SCAN"));
+    assert(planHasState(explain.queryPlanner.winningPlan, "PROJECTION"));
+    assert.eq(2, explain.executionStats.nReturned);
+
```
dstorch: Perhaps also add a negative test showing that we don't use DISTINCT_SCAN when there is an empty query predicate.

dstorch:
@Shekeen code and tests look good. I've left a few comments on some of the specifics. Once those are addressed, this should be good to merge.

dstorch:
@Shekeen this LGTM! I'll squash the commits and merge.

Shekeen:
@dstorch Great, thank you!

PR #1102, opened 2016-07-19
SERVER-24127 validate max size of capped coll

asya999:


src/mongo/db/catalog/collection_options.cpp
```diff
@@ -142,6 +142,8 @@ Status CollectionOptions::parse(const BSONObj& options) {
             cappedSize = e.numberLong();
             if (cappedSize < 0)
                 return Status(ErrorCodes::BadValue, "size has to be >= 0");
+            if ( cappedSize >= 0x7fffffffffffffffLL)
```
GeertBosch: As there are no `long long` values larger than `0x7fffffffffffffffLL`, this essentially just checks for one specific value, however...

src/mongo/db/catalog/collection_options.cpp
```diff
@@ -142,6 +142,8 @@ Status CollectionOptions::parse(const BSONObj& options) {
...
+                return Status(ErrorCodes::BadValue, "size has to be < 2^63");
             cappedSize += 0xff;
```
GeertBosch: ... the problem is really here. This statement causes signed integer overflow for any capped size less than 256 bytes away from the max. The overflow will cause the value to turn negative (or arbitrary badness due to undefined execution) after the check for negative values. The correct fix is to write the alignment expression so it doesn't overflow. Of course checking correct parsing of these huge collections is impractical at the integration level, but correct behavior can be tested in the `collection_options_test` unit test.

GeertBosch:
Hi Asya, thank you for your pull request. Unfortunately, your patch doesn't fully address the underlying issue, see comments inline. We'll fix this by correcting the code for rounding down the size value, so I'm closing this PR.

PR #1107, opened 2016-08-16
SERVER-25628 make FileAllocator::get() thread-safe

yjhjstz:


src/mongo/db/storage/mmap_v1/file_allocator.cpp
```diff
@@ -472,11 +472,9 @@ void FileAllocator::run(FileAllocator* fa) {
     }
 }
 
-FileAllocator* FileAllocator::_instance = 0;
+FileAllocator* FileAllocator::_instance = new FileAllocator();
```
andy10gen: Instead of this approach, what about removing the _instance static member from FileAllocator, and instead making a static instance of FileAllocator inside of FileAllocator::get().

That is, 
    FileAllocator* FileAllocator::get() {
        static FileAllocator instance;
        return &instance;
    }
Now that our minimum supported compilers provide thread-safe function static initialization, we can use this approach and ensure that the file allocator is allocated on the first call to get() instead of during static initialization. This is important, in case get() is never called or if the file allocator constructor some day depends on items that are constructed during or after static initialization.

yjhjstz: That's better!

andy10gen: Would you mind putting up a pull request to that effect, and I'll take care of merging it?

yjhjstz: I have modified it by `git commit  --amend`.

andy10gen:
Great. I'm running it through our test suite, and will merge afterward.

andy10gen:
Committed as 4debb625a6fcee351f6e2508dbe2c67709ec3427! Thanks!

PR #1108, opened 2016-08-17
SERVER-25557 Database names must be less than 64 bytes

bjori:


ksuarz:
Thank you for your contribution, @bjori !

PR #1112, opened 2016-09-01
SERVER-25717 Allow for negative $push indexes.

amcgregor:
It is now possible to issue `$push` updates with negative `$position` indexes. These will be treated as an offset from the end of the array.  Ref: the [associated JIRA ticket](https://jira.mongodb.org/browse/SERVER-25717).

For example, given the following document:

``` js
{acl: [
    {grant: true, user: "amcgregor"},
    {grant: true, user: "bdole"},
    {grant: false}
]}
```

Issuing the following update:

``` js
{$push: {acl: {$each: [{grant: true, user: "algore"}], $position: -1}}}
```

Will result in:

``` js
{acl: [
    {grant: true, user: "amcgregor"},
    {grant: true, user: "bdole"},
    {grant: true, user: "algore"},
    {grant: false}
]}
```

This matches Python list manipulation semantics.

Tests have not been updated, range checking should probably be put in there somewhere (or may be covered by the "32-bit integer" test above the modified prep section), and the approach used may be terrible given that this is the first C++ code I've touched in 10 years.

### DIFF
--- a/src/mongo/db/ops/modifier_push.cpp
+++ b/src/mongo/db/ops/modifier_push.cpp
@@ -546,7 +546,11 @@ Status ModifierPush::apply() const {
 
     // Handle negative starting positions as a distance from the end of the array.
     if (_startPosition < 0) {
-        _preparedState->startPosition = size_t(_preparedState->arrayPreModSize + _startPosition);
+        if (_preparedState->arrayPreModSize + _startPosition < 0) {
+            _preparedState->startPosition = 0;
+        } else {
+            _preparedState->startPosition = size_t(_preparedState->arrayPreModSize + _startPosition);
+        }
     } else {
         _preparedState->startPosition = size_t(_startPosition);
     }
--- a/src/mongo/db/ops/modifier_push.cpp
+++ b/src/mongo/db/ops/modifier_push.cpp
@@ -546,7 +546,7 @@ Status ModifierPush::apply() const {
 
     // Handle negative starting positions as a distance from the end of the array.
     if (_startPosition < 0) {
-        if (_preparedState->arrayPreModSize + _startPosition < 0) {
+        if (size_t(abs(_startPosition)) > _preparedState->arrayPreModSize) {
             _preparedState->startPosition = 0;
         } else {
             _preparedState->startPosition = size_t(_preparedState->arrayPreModSize + _startPosition);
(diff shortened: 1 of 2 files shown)


amcgregor:
Added additional paranoid protections to ensure extremely large negative indexes (`abs(startPosition) > arrayPreModSize`) correctly insert at position zero.

dstorch:
Hi @amcgregor,

Thanks for taking the time to put together this pull request and thanks for your interest in MongoDB!

Any change to the MongoDB query language must be taken very seriously, since this has implications on how users interact with the system, requires documentation, could constrain future implementation choices, may interact poorly with other features we want to add, and so forth. Although we are interested in making enhancements to the update capabilities to MongoDB, we are currently in the design phase for 3.6, and this change may not align with how we want to evolve the update language. The update enhancements we are designing are tracked by https://jira.mongodb.org/browse/SERVER-27089. I am going to close this pull request.

Best,
Dave

amcgregor:
So this was bumped due to the lock-down for 3.4, and is now rejected "because 3.6", with a list of complaints that hold true for literally _any_ change one could possibly propose.  (Updating documentation, of course.  Constrain future implementation choices?  Really?  For a code path that is currently reports an error…)

You are literally choosing to persist a race condition.  I am not able rightly to apprehend the kind of confusion of ideas that could provoke such a resolution.  (… Babbage.)

amcgregor:
Upon review of SERVER-27089 it clearly has no relation at all to the problem resolved by this pull.  In fact, that ticket has nothing to do with `$push` operations _at all_ beyond the "multiple array updates in one operation" ancillary issue.

* Updating all scalars in an array.  **Nope.**
* Updating all scalars in an array which match a predicate.  **Nope.**
* Updating all nested documents in an array.  **Nope.**
* Updating all nested documents in an array which match a query predicate.  **Nope.**
* Nested positional array updates…  **Nope.**
* Nested poly array updates.  **Nope.**
* Related: SERVER-831 Positional Operator Matching Nested Arrays  **Not a nesting problem.**
* Related: SERVER-1243 New operator to update all matching items in an array  **Nothing to do with pushing a single value.**

Please, please explain why this pull was rejected with more than boilerplate. The race condition persists, and none of this additional JIRA material in any way obviates this pull, or the original JIRA ticket.  (Which, interestingly, isn't referenced by this new one, either.)

PR #1113, opened 2016-09-07
SERVER-25173 substrBytes rejects negative starting index and length.

benjaminmurphy:


benjaminmurphy:
@dstorch: Will need you to sign off.

dstorch:
I'm going to close this pull request. Although the behavior change it implements is desirable, it also could break existing applications. Instead we plan to provide safer handling of negative length values  meaning "include the rest of the string".

PR #1115, opened 2016-10-02
Rewrite readme in markdown format

nakulpathak3:


ramonfm:
Thanks for the pull request @nakulpathak3. Please note that the main goal of the README is not to look pretty on github, but to be readable with system tools like cat, less, etc., so we're not accepting pull requests for format changes to the README at this time.

Cheers,
Ramón.

PR #1118, opened 2016-11-17
GeneralImprovements

ZharkovIlia:
In general, there is performance improvement. Some bugs are fixed and some redundancy is eliminated.

### DIFF
--- a/src/mongo/client/connection_pool.cpp
+++ b/src/mongo/client/connection_pool.cpp
@@ -110,7 +110,7 @@ void ConnectionPool::closeAllInUseConnections() {
 void ConnectionPool::_cleanUpStaleHosts_inlock(Date_t now) {
     if (now > _lastCleanUpTime + kCleanUpInterval) {
         for (HostLastUsedMap::iterator itr = _lastUsedHosts.begin(); itr != _lastUsedHosts.end();
-             itr++) {
+             ++itr) {
             if (itr->second <= _lastCleanUpTime) {
                 ConnectionList connList = _connections.find(itr->first)->second;
                 _cleanUpOlderThan_inlock(now, &connList);
--- a/src/mongo/client/dbclient_rs.cpp
+++ b/src/mongo/client/dbclient_rs.cpp
@@ -124,7 +124,7 @@ ReadPreferenceSetting* _extractReadPref(const BSONObj& query, int queryOptions) 
     }
 
     // Default read pref is primary only or secondary preferred with slaveOK
-    ReadPreference pref = queryOptions & QueryOption_SlaveOk
+    ReadPreference pref = (queryOptions & QueryOption_SlaveOk)
         ? mongo::ReadPreference::SecondaryPreferred
         : mongo::ReadPreference::PrimaryOnly;
     return new ReadPreferenceSetting(pref, TagSet());
--- a/src/mongo/client/dbclient_rs_test.cpp
+++ b/src/mongo/client/dbclient_rs_test.cpp
@@ -98,7 +98,7 @@ private:
 
 void assertOneOfNodesSelected(MockReplicaSet* replSet,
                               ReadPreference rp,
-                              const std::vector<std::string> hostNames) {
+                              const std::vector<std::string>& hostNames) {
     DBClientReplicaSet replConn(replSet->getSetName(), replSet->getHosts(), StringData());
     ReplicaSetMonitor::get(replSet->getSetName())->startOrContinueRefresh().refreshAll();
     bool secondaryOk = (rp != ReadPreference::PrimaryOnly);
--- a/src/mongo/client/native_sasl_client_session.cpp
+++ b/src/mongo/client/native_sasl_client_session.cpp
@@ -38,7 +38,7 @@
 namespace mongo {
 namespace {
 
-SaslClientSession* createNativeSaslClientSession(const std::string mech) {
+SaslClientSession* createNativeSaslClientSession(const std::string& mech) {
     return new NativeSaslClientSession();
 }
 
--- a/src/mongo/db/auth/auth_index_d.cpp
+++ b/src/mongo/db/auth/auth_index_d.cpp
@@ -100,7 +100,7 @@ Status verifySystemIndexes(OperationContext* txn) {
     std::vector<IndexDescriptor*> indexes;
     indexCatalog->findIndexesByKeyPattern(txn, v1SystemUsersKeyPattern, false, &indexes);
 
-    if (indexCatalog && !indexes.empty()) {
+    if (!indexes.empty()) {
         fassert(ErrorCodes::AmbiguousIndexKeyPattern, indexes.size() == 1);
         return Status(ErrorCodes::AuthSchemaIncompatible,
                       "Old 2.4 style user index identified. "
--- a/src/mongo/db/catalog/collection.cpp
+++ b/src/mongo/db/catalog/collection.cpp
@@ -1129,7 +1129,6 @@ public:
 
                 if (indexEntryCount != 0) {
                     indexEntryCount--;
-                    //dassert(indexEntryCount >= 0);
                     if (indexEntryCount == 0) {
                         _indexKeyCountTableNumEntries--;
                     }
--- a/src/mongo/db/catalog/collection.cpp
+++ b/src/mongo/db/catalog/collection.cpp
@@ -1129,7 +1129,7 @@ public:
 
                 if (indexEntryCount != 0) {
                     indexEntryCount--;
-                    dassert(indexEntryCount >= 0);
+                    //dassert(indexEntryCount >= 0);
                     if (indexEntryCount == 0) {
                         _indexKeyCountTableNumEntries--;
                     }
--- a/src/mongo/db/commands/feature_compatibility_version.cpp
+++ b/src/mongo/db/commands/feature_compatibility_version.cpp
@@ -207,8 +207,6 @@ void FeatureCompatibilityVersion::set(OperationContext* txn, StringData version)
         // featureCompatibilityVersion to 3.4. This is a new index version that isn't supported by
         // versions of MongoDB earlier than 3.4 that will cause 3.2 secondaries to crash when it is
         // replicated.
-        std::vector<BSONObj> indexSpecs{k32IncompatibleIndexSpec};
-
         {
             ScopedTransaction transaction(txn, MODE_IX);
             AutoGetOrCreateDb autoDB(txn, nss.db(), MODE_X);
@@ -298,8 +296,6 @@ void FeatureCompatibilityVersion::setIfCleanStartup(OperationContext* txn,
         // featureCompatibilityVersion to 3.4. This is a new index version that isn't supported by
         // versions of MongoDB earlier than 3.4 that will cause 3.2 secondaries to crash when it is
         // cloned.
-        std::vector<BSONObj> indexSpecs{k32IncompatibleIndexSpec};
-
         {
             ScopedTransaction transaction(txn, MODE_IX);
             AutoGetOrCreateDb autoDB(txn, nss.db(), MODE_X);
--- a/src/mongo/db/concurrency/lock_manager.cpp
+++ b/src/mongo/db/concurrency/lock_manager.cpp
@@ -229,7 +229,7 @@ struct LockHead {
 
     // Methods to maintain the granted queue
     void incGrantedModeCount(LockMode mode) {
-        invariant(grantedCounts[mode] >= 0);
+        //invariant(grantedCounts[mode] >= 0);
         if (++grantedCounts[mode] == 1) {
             invariant((grantedModes & modeMask(mode)) == 0);
             grantedModes |= modeMask(mode);
@@ -246,7 +246,7 @@ struct LockHead {
 
     // Methods to maintain the conflict queue
     void incConflictModeCount(LockMode mode) {
-        invariant(conflictCounts[mode] >= 0);
+        //invariant(conflictCounts[mode] >= 0);
         if (++conflictCounts[mode] == 1) {
             invariant((conflictModes & modeMask(mode)) == 0);
             conflictModes |= modeMask(mode);
--- a/src/mongo/db/concurrency/lock_manager.cpp
+++ b/src/mongo/db/concurrency/lock_manager.cpp
@@ -229,7 +229,6 @@ struct LockHead {
 
     // Methods to maintain the granted queue
     void incGrantedModeCount(LockMode mode) {
-        //invariant(grantedCounts[mode] >= 0);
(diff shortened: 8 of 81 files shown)


src/mongo/client/dbclient_rs.cpp
```diff
@@ -124,7 +124,7 @@ ReadPreferenceSetting* _extractReadPref(const BSONObj& query, int queryOptions)
     }
 
     // Default read pref is primary only or secondary preferred with slaveOK
-    ReadPreference pref = queryOptions & QueryOption_SlaveOk
+    ReadPreference pref = (queryOptions & QueryOption_SlaveOk)
```
acmorrow: Bitwise and has a higher precedence than the ternery operator, so I don't think this code needed to change.

src/mongo/client/connection_pool.cpp
```diff
@@ -110,7 +110,7 @@ void ConnectionPool::closeAllInUseConnections() {
 void ConnectionPool::_cleanUpStaleHosts_inlock(Date_t now) {
     if (now > _lastCleanUpTime + kCleanUpInterval) {
         for (HostLastUsedMap::iterator itr = _lastUsedHosts.begin(); itr != _lastUsedHosts.end();
-             itr++) {
+             ++itr) {
```
acmorrow: When the result is not captured/used, the definition of the increment operator is inline, it seems unlikely that the compiler will generate different code for ++itr and itr++. Can you actually demonstrate a performance or codegen difference from this or any of the other identical changes?

src/mongo/client/dbclient_rs_test.cpp
```diff
@@ -98,7 +98,7 @@ class BasicRS : public unittest::Test {
 
 void assertOneOfNodesSelected(MockReplicaSet* replSet,
                               ReadPreference rp,
-                              const std::vector<std::string> hostNames) {
+                              const std::vector<std::string>& hostNames) {
```
acmorrow: While you aren't wrong that this is wasteful, this is also in a test. I don't see that it matters.

src/mongo/db/catalog/collection.cpp
```diff
@@ -1129,7 +1129,7 @@ class RecordStoreValidateAdaptor : public ValidateAdaptor {
 
                 if (indexEntryCount != 0) {
                     indexEntryCount--;
-                    dassert(indexEntryCount >= 0);
+                    //dassert(indexEntryCount >= 0);
```
acmorrow: - Why did you comment out the dassert
- If the dassert should be removed, please just delete it, rather than commenting it out.

src/mongo/client/native_sasl_client_session.cpp
```diff
@@ -38,7 +38,7 @@
 namespace mongo {
 namespace {
 
-SaslClientSession* createNativeSaslClientSession(const std::string mech) {
+SaslClientSession* createNativeSaslClientSession(const std::string& mech) {
```
acmorrow: @spencerjackson - Why is this parameter unused?

src/mongo/db/auth/auth_index_d.cpp
```diff
@@ -100,7 +100,7 @@ Status verifySystemIndexes(OperationContext* txn) {
     std::vector<IndexDescriptor*> indexes;
     indexCatalog->findIndexesByKeyPattern(txn, v1SystemUsersKeyPattern, false, &indexes);
 
-    if (indexCatalog && !indexes.empty()) {
+    if (!indexes.empty()) {
```
acmorrow: I presume you are removing this because if indexCatalog was null, line 99 above would be undefined behavior, and therefore we don't need the check?

src/mongo/db/commands/feature_compatibility_version.cpp
```diff
@@ -207,8 +207,6 @@ void FeatureCompatibilityVersion::set(OperationContext* txn, StringData version)
         // featureCompatibilityVersion to 3.4. This is a new index version that isn't supported by
         // versions of MongoDB earlier than 3.4 that will cause 3.2 secondaries to crash when it is
         // replicated.
-        std::vector<BSONObj> indexSpecs{k32IncompatibleIndexSpec};
```
acmorrow: @tessavitabile - Did this declaration (and the identical on below) become unused as some point? I presume that is why it has been deleted here.

src/mongo/db/concurrency/lock_manager.cpp
```diff
@@ -229,7 +229,7 @@ struct LockHead {
 
     // Methods to maintain the granted queue
     void incGrantedModeCount(LockMode mode) {
-        invariant(grantedCounts[mode] >= 0);
+        //invariant(grantedCounts[mode] >= 0);
```
acmorrow: Why did you remove this invariant? Do you believe it will never fire? If you can prove that, just delete it, don't comment it out.

src/mongo/db/concurrency/lock_manager.cpp
```diff
@@ -246,7 +246,7 @@ struct LockHead {
 
     // Methods to maintain the conflict queue
     void incConflictModeCount(LockMode mode) {
-        invariant(conflictCounts[mode] >= 0);
+        //invariant(conflictCounts[mode] >= 0);
```
acmorrow: Same here re invariant.

src/mongo/db/pipeline/granularity_rounder_preferred_numbers_test.cpp
```diff
@@ -51,7 +51,7 @@ const vector<string> preferredNumberSeries{
  * we check to see if the values are approximately equal with tolerance 'delta'.
  */
 void testEquals(Value actual, Value expected, double delta = DELTA) {
-    if (actual.getType() == BSONType::NumberDouble || actual.getType() == BSONType::NumberDouble) {
+    if (actual.getType() == BSONType::NumberDouble || expected.getType() == BSONType::NumberDouble) {
```
acmorrow: Nice catch. How did you find this?

src/mongo/db/range_deleter.cpp
```diff
@@ -129,7 +129,7 @@ static void logCursorsWaiting(RangeDeleteEntry* entry) {
 }
 
 struct RangeDeleter::NSMinMax {
-    NSMinMax(std::string ns, const BSONObj min, const BSONObj max) : ns(ns), min(min), max(max) {}
+    NSMinMax(const std::string& ns, const BSONObj min, const BSONObj max) : ns(ns), min(min), max(max) {}
```
acmorrow: A better change here might be to leave the parameter as by-value, and just re-write the member initializer as ```ns(std::move(ns))```

src/mongo/db/repl/data_replicator_test.cpp
```diff
@@ -131,7 +131,7 @@ class DataReplicatorTest : public executor::ThreadPoolExecutorTest, public SyncS
         return _syncSourceSelector->shouldChangeSyncSource(currentSource, metadata);
     }
 
-    void scheduleNetworkResponse(std::string cmdName, const BSONObj& obj) {
+    void scheduleNetworkResponse(const std::string& cmdName, const BSONObj& obj) {
```
acmorrow: I don't care too much about eliminating a string copy inside a unit test.

src/mongo/db/repl/rs_rollback.cpp
```diff
@@ -125,7 +125,7 @@ namespace {
 
 class RSFatalException : public std::exception {
 public:
-    RSFatalException(std::string m = "replica set fatal exception") : msg(m) {}
+    RSFatalException(const std::string& m = "replica set fatal exception") : msg(m) {}
```
acmorrow: Much like above, you could just leave the parameter as by-value, and then write the initailizer for msg as ```msg(std::move(m))```.

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
```diff
@@ -185,7 +185,7 @@ void* MemoryMappedFile::map(const char* filenameIn, unsigned long long& length)
     filename[255] = 0;
     {
         size_t len = strlen(filename);
-        for (size_t i = len - 1; i >= 0; i--) {
+        for (int i = (len > 0 ? static_cast<int>(len - 1) : -1); i >= 0; i--) {
```
acmorrow: Maybe simpler to just only execute the for loop if len > 0?

src/mongo/dbtests/jstests.cpp
```diff
@@ -1439,7 +1439,7 @@ class DeeplyNestedObject : public TestRoundTrip {
         } else {
             std::stringstream ss;
             ss << "{\"" << depth << "\":" << buildJson(depth - 1) << "}";
-            depth--;
+            depth--; // For developers: Is this operation redundant? Is something missed?
```
acmorrow: Why do you think it is redundant?

src/mongo/util/net/sockaddr.cpp
```diff
@@ -144,7 +144,7 @@ bool SockAddr::isLocalHost() const {
         default:
             return false;
     }
-    fassert(16502, false);
+    //fassert(16502, false);
```
acmorrow: If you want to remove code, remove it. Even better here would be to write MONGO_UNREACHABLE, in which case you can probably remove the ```return false``` as well, since MONGO_UNREACHABLE has ```noreturn``` semantics.

ramonfm:
@ZharkovIlia , please note that you'll need to sign the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement) before we can consider incorporating any of your changes into our tree.

src/mongo/client/native_sasl_client_session.cpp
spencerjackson: Polymorphism. SaslClientSession::create is set to point to this function. Another implementation, createCyrusSaslClientSession, uses the parameter.

src/mongo/client/dbclient_rs.cpp
ZharkovIlia: I only assumed that it can improve code readability. Of course, ignore this change, if you want.

src/mongo/client/connection_pool.cpp
ZharkovIlia: I can demonstrate a codegen difference only when compiling via gcc without optimizations.

src/mongo/client/dbclient_rs_test.cpp
ZharkovIlia: I've understood.

src/mongo/db/auth/auth_index_d.cpp
ZharkovIlia: 'indexCatalog' is assigned a pointer to a field of object 'collection' of class 'Collection', so it cannot be null.

src/mongo/db/commands/feature_compatibility_version.cpp
ZharkovIlia: indexSpecs variable is never used.

src/mongo/db/pipeline/granularity_rounder_preferred_numbers_test.cpp
ZharkovIlia: I just study one of static analyzer and I chose your project at random.

src/mongo/db/catalog/collection.cpp
ZharkovIlia: indexEntryCount is unsigned integer variable. If it can be negative in your mind consider using signed integer variable.

src/mongo/db/concurrency/lock_manager.cpp
ZharkovIlia: grantedCounts is an array of unsigned integers.

src/mongo/db/concurrency/lock_manager.cpp
ZharkovIlia: Same as in the example above.

src/mongo/db/range_deleter.cpp
ZharkovIlia: Yes, I agree with you.

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
ZharkovIlia: I am only concerned about comparison between unsigned integer variable and 0. Even simpler it may be if len will have type int, because size of 'filename' is not more than 256.

src/mongo/dbtests/jstests.cpp
ZharkovIlia: 'depth' is a function parameter and is never used after line 1441.

src/mongo/util/net/sockaddr.cpp
ZharkovIlia: I will take it into account.

ZharkovIlia:
@acmorrow , Is there anything that prevents you from acceptance this pull request?

src/mongo/client/connection_pool.cpp
acmorrow: OK. Then I would prefer that you back out the iterator changes, everywhere.

src/mongo/client/dbclient_rs.cpp
acmorrow: Please back this change out.

src/mongo/client/dbclient_rs_test.cpp
acmorrow: Please unwind the change then

src/mongo/db/auth/auth_index_d.cpp
acmorrow: Yes, we are saying the same thing then. This change is fine.

src/mongo/db/commands/feature_compatibility_version.cpp
acmorrow: ping @tessavitabile re the unused indexSpecs variable. Also, I would have expected a compiler diagnostic if these are actually unused.

acmorrow:
@ZharkovIlia - Have you signed the contributor agreement?

src/mongo/db/commands/feature_compatibility_version.cpp
acmorrow: No, you can leave this one for now.

ZharkovIlia:
@acmorrow - Yes, I have signed the contributor agreement.

ramonfm:
Thanks for signing the contributor agreement @ZharkovIlia 
 
I see you've submitted #1122 , does that PR replaced this one? If so, can we close this one to avoid confusion?

Thanks,
Ramón.

ZharkovIlia:
Yes, of course.

ramonfm:
Thanks, closing now.

PR #1121, opened 2016-11-30
Fix for negative Date timezone offsets 

thesebas:
When uncommented this `// print a non-UTC time` block and changed `var UTC = '';` got invalid timezone offsets like: `ISODate("2016-11-14T00:00:00+-100")`. This makes Date be displayed as `ISODate("2016-11-14T00:00:00+0100")`


ramonfm:
@thesebas, please note that before we inspect your pull request you'll need to sign the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement). You'll also need to [open a SERVER ticket](https://jira.mongodb.org).

Regards,
Ramón.

PR #1122, opened 2016-12-01
General Improvements v2

ilia1243:
Elimination of some redundancy. Some function arguments are passed by reference. Few bugs are fixed.

### DIFF
--- a/src/mongo/db/commands/feature_compatibility_version.cpp
+++ b/src/mongo/db/commands/feature_compatibility_version.cpp
@@ -207,8 +207,6 @@ void FeatureCompatibilityVersion::set(OperationContext* txn, StringData version)
         // featureCompatibilityVersion to 3.4. This is a new index version that isn't supported by
         // versions of MongoDB earlier than 3.4 that will cause 3.2 secondaries to crash when it is
         // replicated.
-        std::vector<BSONObj> indexSpecs{k32IncompatibleIndexSpec};
-
         {
             ScopedTransaction transaction(txn, MODE_IX);
             AutoGetOrCreateDb autoDB(txn, nss.db(), MODE_X);
@@ -298,8 +296,6 @@ void FeatureCompatibilityVersion::setIfCleanStartup(OperationContext* txn,
         // featureCompatibilityVersion to 3.4. This is a new index version that isn't supported by
         // versions of MongoDB earlier than 3.4 that will cause 3.2 secondaries to crash when it is
         // cloned.
-        std::vector<BSONObj> indexSpecs{k32IncompatibleIndexSpec};
-
         {
             ScopedTransaction transaction(txn, MODE_IX);
             AutoGetOrCreateDb autoDB(txn, nss.db(), MODE_X);
--- a/src/mongo/db/storage/mmap_v1/mmap_windows.cpp
+++ b/src/mongo/db/storage/mmap_v1/mmap_windows.cpp
@@ -184,8 +184,8 @@ void* MemoryMappedFile::map(const char* filenameIn, unsigned long long& length) 
     strncpy(filename, filenameIn, 255);
     filename[255] = 0;
     {
-        size_t len = strlen(filename);
-        for (size_t i = len - 1; i >= 0; i--) {
+        int len = strlen(filename);
+        for (int i = len - 1; i >= 0; i--) {
             if (filename[i] == '/' || filename[i] == '\\')
                 break;
 
--- a/src/mongo/dbtests/jstests.cpp
+++ b/src/mongo/dbtests/jstests.cpp
@@ -1439,7 +1439,6 @@ class DeeplyNestedObject : public TestRoundTrip {
         } else {
             std::stringstream ss;
             ss << "{\"" << depth << "\":" << buildJson(depth - 1) << "}";
-            depth--;
             return ss.str();
         }
     }
--- a/src/mongo/client/native_sasl_client_session.cpp
+++ b/src/mongo/client/native_sasl_client_session.cpp
@@ -38,7 +38,7 @@
 namespace mongo {
 namespace {
 
-SaslClientSession* createNativeSaslClientSession(const std::string mech) {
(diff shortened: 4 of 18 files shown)


src/mongo/db/commands/feature_compatibility_version.cpp
```diff
@@ -207,8 +207,6 @@ void FeatureCompatibilityVersion::set(OperationContext* txn, StringData version)
         // featureCompatibilityVersion to 3.4. This is a new index version that isn't supported by
         // versions of MongoDB earlier than 3.4 that will cause 3.2 secondaries to crash when it is
         // replicated.
-        std::vector<BSONObj> indexSpecs{k32IncompatibleIndexSpec};
```
acmorrow: ping @visemet - Looks like you added these. I'm surprised these didn't result in a compiler warning?

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
```diff
@@ -184,8 +184,8 @@ void* MemoryMappedFile::map(const char* filenameIn, unsigned long long& length)
     strncpy(filename, filenameIn, 255);
     filename[255] = 0;
     {
-        size_t len = strlen(filename);
-        for (size_t i = len - 1; i >= 0; i--) {
+        int len = strlen(filename);
```
acmorrow: strlen really does return a size_t:

http://pubs.opengroup.org/onlinepubs/009695399/functions/strlen.html

In the event that len is zero, what should happen here? I don't think just avoiding this cleaning loop is sufficient. Does an empty file name actually make sense here?

src/mongo/dbtests/jstests.cpp
```diff
@@ -1439,7 +1439,6 @@ class DeeplyNestedObject : public TestRoundTrip {
         } else {
             std::stringstream ss;
             ss << "{\"" << depth << "\":" << buildJson(depth - 1) << "}";
-            depth--;
```
acmorrow: Why did this get removed?

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
ZharkovIlia: If len cannot be zero, it can have size_t type. But if type of iterator 'i' is size_t, condition 'i >= 0' is always true and it either is redundant or access to array at index ((size_t) -1) is possible. And endless cycle which ends with SIGSEGV is also possible.

src/mongo/dbtests/jstests.cpp
ZharkovIlia: Because depth is never used after line 1442.

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
acmorrow: Yes I understand that. My point is: what does a zero length filename mean overall to this function. Is it sensible? Or should it be rejected earlier, in which case we would know that strlen would return >0, so use of size_t would be correct both for capturing the return value of strlen, and for iterating.

ZharkovIlia: I'm sorry, if I don't understand you clearly. Will it be better, if I change only line 188?

src/mongo/dbtests/jstests.cpp
acmorrow: Ack

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
acmorrow: - Think about what the function MemoryMappedFile::map does. What would it mean if you passed it the empty string for the `filenameIn` parameter? Should that be allowed?
- If the answer is "no", then we should prohibit passing in an empty string, in which case we know that strlen will return a value >= 1, so we wouldn't need to change the loop.

In other words, what I'm asking you to do is to enforce a stronger invariant on the filenameIn parameter, such that the code here is correct as written and your change is not needed.

src/mongo/db/commands/feature_compatibility_version.cpp
ZharkovIlia: I'm new to github. How can I ping @visemet? I've always been convinced that if a local variable is declared but not used in appropriate scope, it cannot influence program behaviour. Am I wrong?

visemet: > Second, I want to make sure that the unused variable is just an accident, and doesn't signal that the code was incorrectly written or modified. Perhaps the variable is supposed to be being used!

The `indexSpecs` variable is indeed unnecessary. The initial version of the patch was modeled off logic in `CmdCreateIndex::run()` and used `MultiIndexBlock::init(const std::vector<BSONObj>& specs)` to create the index. While iterating on the patch, I realized that I could write a more compact of this logic by using the `IndexBuilder` class. It appears I mistakenly forgot to remove the `indexSpecs` variable when making that change.

> First, I would have expected the compiler to issue a warning, and I'd like to understand why it didn't in this case.

Are there particular `scons` options needed to cause such a warning to be emitted?

acmorrow: @visemet - I suspect that since this object has non-trivial constructor/destructor semantics, that the compiler must assume that simply creating the object could be meaningful even if it is not used, and therefore it does not issue a diagnostic.

acmorrow: @visemet Thanks for clarifying about the unused variable. @ZharkovIlia, your change here is correct.

src/mongo/db/storage/mmap_v1/mmap_windows.cpp
acmorrow: @ZharkovIlia If you would like to see this PR merged (and, I hope you do, because you put a lot of work in to it!), we will still need to sort out this one last issue. Please let me know if I can help clarify anything in this discussion.

sb0210:
@CodiumAI-Agent /describe

PR #1126, opened 2016-12-29
Fix start-stop-daemon stop issue

guiguan:
Fix `No /usr/bin/mongod found running; none killed.` error when using `service mongod stop`

### DIFF
--- a/debian/init.d
+++ b/debian/init.d
@@ -155,8 +155,7 @@ stop_server() {
 # Stop the process using the wrapper
             start-stop-daemon --stop --quiet --pidfile $PIDFILE \
                         --retry 300 \
-                        --user $DAEMONUSER \
-                        --exec $DAEMON
+                        --user $DAEMONUSER
             errcode=$?
         return $errcode
 }


ehershey:
Hi @guiguan ! Thanks for the contribution. Can you give some more detail about your error and how this solves the problem? 

Also - two requirements we have for accepting contributions are filling out our contributors' agreement, and having a SERVER ticket in our JIRA system. Can you fill out the agreement and enter a SERVER ticket? 

The contributor agreement can be found here - https://www.mongodb.com/legal/contributor-agreement. 

You can create a JIRA ticket here - https://jira.mongodb.org/secure/CreateIssue!default.jspa

harishanand95:
I had a similar issue and this fix solved the stop issue.
`service mongod stop                                                                                 [fail]` 

@ehershey I use a ubuntu docker image and db version v3.4.7.

Schubes:
Hi @guiguan,

Since we haven't heard back since Ernie's request to sign the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement). I'm going to close this pull request. If you are still interested in contributing, please sign the agreement and open a [SERVER ticket](https://jira.mongodb.org/browse/SERVER/)  so we can consider this patch.

Thank you,
Kelsey

peetersdiet:
@Schubes This is a real issue, at least in Docker containers, with an exceptionally easy fix. Can't someone with a contributor agreement replicate this and have a PR approved?

I fixed it manually, but it took me considerable time to analyse the cause and finally produce a relevant google search result, pointing me to this fix. I'm sure I'm not the only one wasting time on something that's ridiculuously easy to fix in your codebase.

mattlord:
@peetersdiet and @guiguan, I'm looking at this now to see if it's still relevant, and why. 
1. Debian and Ubuntu have both moved from sysv init to systemd for service management.
2. Since $DAEMON == "/usr/bin/mongod", I'm unsure how that would help resolve the "No /usr/bin/mongod found running; none killed." error. Is the real underlying issue that the PIDFILE contents aren't valid/correct for some reason in your specific case?

I know that this seems very basic, but any change also has the potential for negative impacts on our millions of users, so w/o a clear cause and test case it's difficult to make this type of change. So any addition info you can provide would be very helpful. Thanks!

mattlord:
Are you running mongod in a container on a linux host, and then trying to manage that containerized process using init.d on the host machine?

peetersdiet:
@mattlord It's some time ago, but IIRC I was building a docker from the `Ubuntu 18.04` image.
During the build, the mongodb apt repository was added, after which mongodb server was installed and started (`service mongodb start`). Then the buildscript went on to provision configuration data to mongodb. During the cleanup of the build, `service mongodb stop` was called, but failed to stop the service.

I don't remember exactly what issue it caused, but I think it made the build fail or would pose problems on subsequent starts of the container.

mattlord:
@peetersdiet thank you! So this error was seen during the generation of a custom docker image via `docker build`? If so, it definitely sounds like something odd/unexpected was happening as an image build should not start any processes -- which would also explain why there was nothing to stop. Do you happen to have the Dockerfile around still?

Or do you mean you saw it during your usage of the container image via `docker run`? Either way, the Dockerfile would be helpful. This is the base image for Ubuntu 18.04: 
https://github.com/tianon/docker-brew-ubuntu-core/blob/1cc295b1507b68a66942b2ff5c2dbf395850208a/bionic/Dockerfile

To start and stop services in a container would be unusual as the container is just an isolation layer around a process running on the host. So the command executed inside the container (isolated context) -- the CMD run via the instantiation of `docker run` -- would be /usr/bin/mongod and you would stop that process/container using `docker stop`. So I'm unsure where/why the `service` calls are coming into play at all.

peetersdiet:
@mattlord I fail to see why a docker build should not start any processes. You can execute commands just fine during a build of a custom image using the RUN directive, which is what I did to prepare a base image for use in deployments later on. Seems like a pretty standard use-case to me.

It's possible my description earlier was not clear enough, but these actions were explicitly in the DockerFile and executed during the build. Not that it really matters, because I manually reproduced the build steps from a bash-shell running in the base Ubuntu docker container and that resulted in the same problematic behaviour. The command `service mongodb stop` fails to stop the mongodb daemon either way.

I looked up the DockerFile. It looks something like the redacted snippet below. Not tested it, but it should give you enough info to reproduce the issue.
```DockerFile
FROM ubuntu:18.04
WORKDIR /
ARG DEBIAN_FRONTEND=noninteractive
ARG APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1
RUN \
    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 9DA31620334BD75D9DCB49F368818C72E52529D4 && \
    echo "deb [ arch=amd64 ] https://repo.mongodb.org/apt/ubuntu bionic/mongodb-org/4.0 multiverse" | tee /etc/apt/sources.list.d/mongodb-org-4.0.list && \
apt-get update && apt-get install -y mongodb-server
RUN service mongodb start
RUN # did stuff here that added data to mongodb 
RUN service mongodb stop
```

mattlord:
Hi @peetersdiet, I don't understand what that Dockerfile is supposed to do. Is that the entire Dockerfile? I suspect that the additional line added to the init.d in the proposed patch simply papers over the fact that there's no mongod running when you call `service stop`? I'm unsure. 

The result of starting a container via `docker run` is that there should be a final CMD running -- this is the actual process running on the docker host within the containerized/isolated context. Starting and stopping a daemon will not work as expected in this context, AFAIK. 

You can see [the "official" MongoDB 4.0 container image from Docker](https://github.com/docker-library/mongo/blob/0a423591d7243256819edc97cef579206d01f6c7/4.0/Dockerfile). What were you trying to do in your custom Dockerfile? Are you trying to e.g. create some initial databases and collections that serve as a base for all of your mongod instances/containers? If so, then you could use the default Docker mongo image, start an instance/container, create your databases and collections, then use `docker commit` to save that as your new custom image so that there's an additional FS layer with your base data in that image. Or was it something else? Thanks!

mattlord:
I cannot find any official Docker container images [that use "service"](https://github.com/docker-library/official-images/search?q=%22service+start%22&unscoped_q=%22service+start%22) or [that use "systemctl"](https://github.com/docker-library/official-images/search?q=%22systemctl%22&unscoped_q=%22systemctl%22) to start/stop services. It just doesn't work properly AFAIK. Is there any info out there that can help me understand what I'm missing?

mattlord:
To try and further explain the situation here... _AFAIUI_, when doing the `docker build` the `service start` call within the Dockerfile will result in a mongod instance running in _a separate container_ (should see "Running in <hash>" in the build output), and likewise the `service stop` will also run in its own container (again, should see "Running in <hash>" in the build output), and there's no mongod actually running in that container (as it was started in a separate container) so we get the expected error. Again, I think that the additional clause just causes it to ignore the fact that there's no matching process to stop. Please let me know if I'm missing or misunderstanding anything though.

mattlord:
@peetersdiet, just as an aside FWIW, in your noted Dockerfile you're adding the official MongoDB Community 4.0 repo but you end up installing the MongoDB package from the ubuntu repo. mongodb-server is the Ubuntu/Canonical package (version '3.6.3-0ubuntu1' on Ubuntu 18.04) in the main repo, where as [mongdb-org-server](https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/#install-the-mongodb-packages) is the one from us.

peetersdiet:
@mattlord You're right about adding the repo and installing from ubuntu packages. Apparently that was a remnant from previous experimtentation. Thanks for spotting that. I also reviewed the original Dockerfile and was a bit quick in my previous example. The start and stop are effectively executed in a single command. I adjusted the example Dockerfile accordingly.

```Dockerfile
FROM ubuntu:18.04
WORKDIR /
ARG DEBIAN_FRONTEND=noninteractive
ARG APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1
RUN \
    apt-get update && apt-get install -y mongodb-server && \
    service mongodb start && \
    sleep 10  && \
    service mongodb stop
```

Doing a `docker build .` with the file above will fail because mongodb cannot be stopped.
```bash
$> docker build .
...
 * Starting database mongodb
   ...done.
 * Stopping database mongodb
   ...fail!
The command '/bin/sh -c apt-get update && apt-get install -y mongodb-server &&     service mongodb start &&     sleep 10  &&     service mongodb stop' returned a non-zero code: 1
$> 
```
Tbh, I don't understand what you're trying to achieve with your inquiries and teachings. It's all fairly simple: this is an issue that is solved by relaxing the constraints on start-stop-daemon and **removing** 1 line specifying the processname-constraint from the init-script. It's a working fix and the root cause is something you can research by using the Dockerfile I provided. I explained two times what it does, i.e. the trivial steps to start and stop mongodb during a build of a docker container. Nothing fancy. You don't need to understand my reasons to reproduce the issue.

Also, instead of doing a root cause analysis, why not assess if the proposed fix is an issue? **Is it an issue to remove the processname constraint?** The patch removes the processname constraint from the start-stop-daemon call, but it will still match only processes with the exact pidfile and running under the correct user.
```diff
diff --git a/debian/init.d b/debian/init.d
index a2b525a7b80..46f04e5f014 100644
--- a/debian/init.d
+++ b/debian/init.d
@@ -155,8 +155,7 @@ stop_server() {
 # Stop the process using the wrapper
             start-stop-daemon --stop --quiet --pidfile $PIDFILE \
                         --retry 300 \
-                        --user $DAEMONUSER \
-                        --exec $DAEMON
+                        --user $DAEMONUSER
             errcode=$?
         return $errcode
 }
```

I fixed this issue on our end by automatically patching the initfile with the proposed change. No sweat. For the resolution of this merge request, if you want to find the root cause and fix that, fine, if you want to dismiss this patch, fine, mark it as wontfix, fine, but I have put in too much time already in this minor issue than I care for. I hope you can appreciate that sentiment.

mattlord:
@peetersdiet you're right, I did get the patch reversed in my comments. This PR is *removing* the "--exec" line rather than adding it. Sorry, trying to do too many things at once. :-) So without that option we would be killing whatever pid was listed in the pidfile, regardless of whether it's actually a mongod instance or not. That doesn't seem like good behavior to me. Given that:
1. Ubuntu and Debian both use the exact same `start-stop-daemon` call in their init.d scripts that we are (e.g. [see the one you were installing from canonical](https://launchpad.net/ubuntu/+source/mongodb/1:3.6.3-0ubuntu1)) 
2. It's standard/good practice to use --exec to ensure that you're actually going to be stopping the specific daemon associated with the service -- which is /usr/bin/mongod, the same binary we're starting with the service (e.g. see `grep "start-stop-daemon" "exec" /etc/init.d/*`)
3. There's something odd/unexpected happening in your specific case, which I don't fully understand. 

Changing our behavior here doesn't seem wise and I'm going to close the related Jira ticket on our side as won't fix. I appreciate the time that you offered here, and I completely understand that you cannot spend more on it. I was only trying to find out more details to see if perhaps I was making a mistake by closing it as won't fix. 

Thank you for using MongoDB, and for helping us to make it better!

Matt

guiguan:
@mattlord this pr was more than two years ago... I think the problem comes when users are trying to containerize mongod themselves, because, for example, they want mongod be part of their app’s integration test. I think all Debian container base OSes are still using sysv init. However, your stop logic failed to work in all cases. Isn’t it better to make your cleanup/stop logic more robust by handling those edge cases? Probably, you guys should test more on those scenarios for your end user.

mattlord:
@guiguan, I'm happy to explore more specific testing. But using either `service` or `systemctl` is not appropriate in a container as a container is a process, it's not a VM. And that's the only sort of specific test case I've seen so far. 

When using the --exec option, the stop call would *only* fail as described if the daemon started by the service is not running -- the daemon started by the service being /usr/bin/mongod. If the daemon started by the service isn't running when you try and stop it then the error is correct and there's another problem happening that should be dealt with directly -- this proposed change is not the correct thing to do as a default.

peetersdiet:
@mattlord Against my intentions, I couldn't let this go and did some more analysis. :)

Apparently start-stop-daemon uses /proc/$pid/exe to test whether the binary from its exec-option is running. But these files seem to be protected when running in unprivileged docker containers, and start-stop-daemon cannot stat the file /proc/$pid/exe, thus failing the match.

I tested again by adding the --privileged flag to docker run and running the start/stop commands manually. With `docker run --privileged -it <image> /bin/bash` I was able to stop mongodb successfully with `service mongodb stop` (exit code 0).

Alas, --privileged is not an option for `docker build`, which imo it should be. Also, I assume running privileged containers needs to be avoided where possible.

My takeaway:
- Most docker usecases will make use of unprivileged containers.
- Imo start-stop-daemon is working correctly, by failing when it encounters an undetermined state that it was instructed to test by the specified arguments. Even if it would be container-aware it would not be able to verify it's exec option in unprivileged containers, thus producing the same issue.
- An appropriate solution could be to make the init-script container-aware and have it drop the exec-parameter when detecting it's running inside a container. Iow, this merq request, but with an additional check to detect the container.

guiguan:
@mattlord you are right for a normal app container. However, for most CI systems, each test is run inside a container. Imaging a UAT test that includes mongod is running inside a CI system’s worker container, and the user wants to start and stop mongod before and after each test case. You will probably argue with me with docker-in-docker approach, but most people just bundle those test deps, such as mongod, straight away.

@peetersdiet that explains, thanks for the research 👍

mattlord:
Thank you @guiguan and @peetersdiet! The `start-stop-deamon` stat behavior + the `docker run` w/o `--privileged` behavior was a helpful combination to put together. Great investigative work. :-) 

I don't disagree that it could be nice to alter the behavior when we know we're operating within a container context. I'm not sure how to reliably detect that though. It [looks like docker is doing it by looking to see if PID 1 is either `systemd` or `init` (both have "init.scope" in their /proc/cgroup file)](https://github.com/docker/docker-ce/blob/8a582e89b407639018e067123eb17f61dbfd10ee/components/engine/pkg/parsers/operatingsystem/operatingsystem_linux.go#L65-L77) -- if not then we're in a container where PID 1 will be the process running in the container.  

The problem is that I'm not sure this is 100% reliable today (docker always returns false for [windows](https://github.com/docker/docker-ce/blob/8a582e89b407639018e067123eb17f61dbfd10ee/components/engine/pkg/parsers/operatingsystem/operatingsystem_windows.go#L47-L51) and [unix](https://github.com/docker/docker-ce/blob/8a582e89b407639018e067123eb17f61dbfd10ee/components/engine/pkg/parsers/operatingsystem/operatingsystem_windows.go#L47-L51)) and I'm even less sure about that moving into the future. I don't think that this is necessarily a documented stable API. I'm also unsure how this would work in the `docker build` context as that seems like a unique context with docker.

Hmmm...

guiguan:
@mattlord we only need the `isRunningInContainer` for the Linux kernel container, because that’s where the sysv init and `start-stop-daemon` are being used. For Windows kernel container, there is no `start-stop-daemon`, so irrelevant. For Darwin kernel container, not yet possible AFAIK, so irrelevant as well.

That process hierarchy based detection looks good to me, here is a better bash version: 
https://stackoverflow.com/a/51352271

For `docker build`, I think it is still inside a container, because if it is not inside a container, it should not have such permission issue. A research should be performed here though.

mattlord:
@guiguan, the problem is that the contents of /proc/self/cgroup is dependent on the specific Linux OS variant and container service (which are changing at a very fast pace). Given that we use this same init script on all Debian based linux variants and the contents of those /proc/cgroup files are dependent on the specific Linux OS variant and container service, it doesn't seem wise. 

We can look at [the specific example you called out above](https://stackoverflow.com/a/51352271), which simply looks for any line in the /proc/self/cgroup file that ends without a specific cgroup prefix. **It doesn't work properly in my setup**:
```
matt@mongodlord:~$ lsb_release -d
Description:	Ubuntu 18.04.2 LTS
matt@mongodlord:~$ uname -a
Linux mongodlord 4.15.0-45-generic #48-Ubuntu SMP Tue Jan 29 16:28:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
matt@mongodlord:~$ docker --version 
Docker version 18.06.1-ce, build e68fc7a
matt@mongodlord:~$ cat /tmp/is_container_test.sh 
#!/bin/sh

is_running_in_container() {
  awk -F: '$3 ~ /^\/$/{ c=1 } END { exit c }' /proc/self/cgroup
}

if is_running_in_container; then
  echo "Aye!! I'm in a container"
else
  echo "Nay!! I'm not in a container"
fi

exit
matt@mongodlord:~$ cat /proc/self/cgroup 
12:devices:/user.slice
11:rdma:/
10:pids:/user.slice/user-1000.slice/session-239.scope
9:memory:/user.slice
8:blkio:/user.slice
7:perf_event:/
6:freezer:/
5:cpuset:/
4:cpu,cpuacct:/user.slice
3:net_cls,net_prio:/
2:hugetlb:/
1:name=systemd:/user.slice/user-1000.slice/session-239.scope
0::/user.slice/user-1000.slice/session-239.scope
matt@mongodlord:~$ /tmp/is_container_test.sh 
Nay!! I'm not in a container
matt@mongodlord:~$ sudo docker run -it --rm -v /tmp:/tmp bash 
bash-5.0# cat /proc/self/cgroup 
12:devices:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
11:rdma:/
10:pids:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
9:memory:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
8:blkio:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
7:perf_event:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
6:freezer:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
5:cpuset:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
4:cpu,cpuacct:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
3:net_cls,net_prio:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
2:hugetlb:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
1:name=systemd:/docker/a90a559bfb6e1c1f5d5c31b7589e0b015f0615898a40adb694d87e11a4264fba
0::/system.slice/docker.service
bash-5.0# /tmp/is_container_test.sh 
Nay!! I'm not in a container
```

I'm going to have to park this discussion for the time being. I just don't see a great solution given how the init file is used.

PR #1127, opened 2016-12-29
[SERVER-7716] change DBClientWithCommands count() to take a Query instead of a BSON object

kolisergej:
Changed signature of mongo::DBClientWithCommands::count function and its dependent classes for complex queries handles.

### DIFF
--- a/src/mongo/client/dbclient.cpp
+++ b/src/mongo/client/dbclient.cpp
@@ -323,8 +323,10 @@ bool DBClientWithCommands::runPseudoCommand(StringData db,
 }
 
 unsigned long long DBClientWithCommands::count(
-    const string& myns, const BSONObj& query, int options, int limit, int skip) {
-    BSONObj cmd = _countCmd(myns, query, options, limit, skip);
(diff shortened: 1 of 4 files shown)


acmorrow:
@kolisergej - What is the motivation for this change? The ticket you are referencing, SERVER-7716, was filed before we pulled the C++ driver out of the server repository, and into its own project:

https://github.com/mongodb/mongo-cxx-driver/tree/legacy

The APIs within the server repository are therefore "internal only". Unless this change would significantly improve our ability to develop the server itself, it is unlikely to be adopted.

If you would like to see this change applied in the C++ driver, then the above repository is the correct one for a PR. Please note however, that the legacy C++ driver is in maintenance-only mode. All work has shifted to furthering the development of the mainline C++ driver:

https://github.com/mongodb/mongo-cxx-driver/tree/releases/stable

The new C++ driver offers a modern interface, aligned with driver standards found across all mognodb drivers, and is implemented in modern C++.

kolisergej:
@acmorrow yes I see now, thanks for clear reply

PR #1128, opened 2016-12-29
[SERVER-8958] Replace strtod/strtol/atoi/atol/etc calls with parseNumberFromString<T>

kolisergej:


src/mongo/db/storage/mmap_v1/btree/btree_test_help.cpp
```diff
@@ -186,9 +186,9 @@ string ArtificialTreeBuilder<OnDiskFormat>::expectedKey(const char* spec) {
 
     // parsing a long long is a pain, so just allow shorter keys for now
     unsigned long long num = strtol(spec + 1, &endPtr, 16);
```
kolisergej: can't remove strtol usage here, because of we need to know pointer position to byte after digit.
parseNumFromString don't provide such possibility

src/mongo/util/text.cpp
```diff
@@ -154,10 +155,8 @@ long long parseLL(const char* n) {
     long long ret;
     uassert(13307, "cannot convert empty string to long long", *n != 0);
 #if !defined(_WIN32)
-    char* endPtr = 0;
-    errno = 0;
-    ret = strtoll(n, &endPtr, 10);
-    uassert(13305, "could not convert string to long long", *endPtr == 0 && errno == 0);
+    const Status parse_res = parseNumberFromString<long long>(n, &ret);
+    uassert(13305, "could not convert string to long long", parse_res == Status::OK());
```
andy10gen: uassert(13305, parse_res.toString(), parse_res.isOK());

Then, we get the error message out of the bad status in the exception that is thrown.

src/mongo/bson/bsonelement.cpp
```diff
@@ -145,8 +145,9 @@ string BSONElement::jsonString(JsonStringFormat format, bool includeFieldNames,
                         for (int x = 0; x < pretty; x++)
                             s << "  ";
                     }
-
-                    if (strtol(e.fieldName(), 0, 10) > count) {
+                    long int field_digit = 0;
+                    if (parseNumberFromString<long int>(e.fieldName(), &field_digit) == Status::OK()
```
andy10gen: We typically use the isOK() method on Status to check if it's OK rather than comparing directly to Status::OK(). For consistency with other code in our codebase, please do that in this patch.

src/mongo/shell/shell_utils_launcher.cpp
```diff
@@ -304,7 +304,7 @@ ProgramRunner::ProgramRunner(const BSONObj& args, const BSONObj& env) {
         if (str == "--port") {
             _port = -2;
         } else if (_port == -2) {
-            _port = strtol(str.c_str(), 0, 10);
+            parseNumberFromString<int>(str.c_str(), &_port);
```
andy10gen: In the event that str does not parse as a number, the old behavior set _port to 0, and this will leave it unchanged at -2.

However, this function uses verify(), above, which means that it also sometimes throws exceptions on errors. It would be OK here to wrap this parseNumberFromString call in uassertStatusOK() to get a bad status converted into a thrown exception:

uassertStatusOK(parseNumberFromString<int>(str, &_port));

I mentioned this on another change in this patch. Also, you don't have to call c_str(), as parseNumberFromString can accept std::strings directly.

src/mongo/util/net/httpclient.cpp
```diff
@@ -85,8 +86,8 @@ int HttpClient::_go(const char* command, string url, const char* body, Result* r
     string::size_type idx = host.find(":");
     if (idx != string::npos) {
         server = host.substr(0, idx);
-        string t = host.substr(idx + 1);
-        port = atoi(t.c_str());
+        const string t = host.substr(idx + 1);
+        parseNumberFromString<int>(t.c_str(), &port);
```
andy10gen: Here's another place where the value of port will be different when the string does not parse in your change vs in the original code. Rather than match the old behavior, here, you could simply convert the bad status to a DBException as follows:

uassertStatusOK(parseNumberFromString<int>(t, &port));

This is OK here, because this method already clearly uses uassert to throw exceptions on errors.

Note also that you don't need or want to call c_str on strings passed into parseNumberFromString, as they can operate on std::string.

src/mongo/bson/bsonelement.cpp
kolisergej: yes, much pretty

kolisergej:
I thought there are 2 main scopes of cpp tests (unittests and dbtest). Is "core" not included into them? If not, are there any others cpp ones ?

kolisergej:
actually there is one failed test case jstests/core/compact_keeps_indexes.js
but I don't think it depends from these changes because of if I run it separately it passes.
Yosemite platform

andy10gen:
Regarding the test failure you're seeing, it's because the ulimit for maximum number of open files is too low. Try running ulimit -n 1024 in the shell before running the script.

The "core" suite are actually integration tests of the behavior of a standalone mongod instance. They run quickly and provide good smoke coverage of some parts of the system that the unit tests do not yet cover.

src/mongo/util/net/httpclient.cpp
```diff
@@ -32,6 +32,7 @@
 #include "mongo/util/net/httpclient.h"
 
 #include "mongo/bson/util/builder.h"
+#include "mongo/base/parse_number.h"
```
andy10gen: Please keep the includes in accordance with the [style guide](https://github.com/mongodb/mongo/wiki/Server-Code-Style#for-cpp-files).

src/mongo/util/text.cpp
```diff
@@ -42,6 +42,7 @@
 #endif
 
 #include "mongo/platform/basic.h"
+#include "mongo/base/parse_number.h"
```
andy10gen: Please keep the includes in accordance with the [style guide](https://github.com/mongodb/mongo/wiki/Server-Code-Style#for-cpp-files).

andy10gen:
Other than the requested changes to #include ordering/style, this looks good. After you make these last changes, I'll have someone run the change through our automated testing system and merge this in.

andy10gen:
Sorry for the long delay. I had some other work holding my attention for the last month. I'll run your patch through our test battery today, and assuming everything works out, will have it merged soon. Thanks, again for your contribution!

andy10gen:
Two errors cropped up during the test run. First, `jstests/core/numberlong.js` was apparently checking for a specific error message on overflows, and will need to be changed. You can run this test with the following command line from the mongo repository root directory:

`python /buildscripts/resmoke.py --executor no_passthrough jstests/core/numberlong.js`

Second, the linter found a minor deviation from our style guide. You can find this using the following invocation of the build system:

`scons lint`

Or, if you have clang-format 3.8.0 installed, you can run the following in the shell and it will perform the necessary reformatting:

`git diff origin/master --name-only | grep -E '\.(h|cpp|js)' | xargs /path/to/clang-format-3.8.0 -i --style=file`

Again, I'm sorry for the long delay in getting back to you. I'm actually quite pleased someone took the time to do this much-needed cleanup.

thisch:
Does anyone know why the corresponding jira ticket has been marked as "won't fix"?

acmorrow:
@thisch - We have re-opened the ticket.

acmorrow:
@kolisergej and @thisch - If you are interested in moving this work forward, please see the most recent discussion in https://jira.mongodb.org/browse/SERVER-8958. The number of remaining instances is quite small, and internal agreement was reached that `std::sto` instances should also be replaced with `parseNumberFromString`.

acmorrow:
This branch is no longer mergeable, and there doesn't seem to be much interest in getting it back in shape per the discussion here and in https://jira.mongodb.org/browse/SERVER-8958, so I'm going to close it. I'll also close out SERVER-8958 as 'wont fix', so the PR and ticket will be in a consistent state.

PR #1132, opened 2017-01-10
Use the overrided environment when generating compile_command.json

laudrup:
When the compilation_db scons emitter saved compilation database
entried it always used the "global" environment which might not match
the compilation command line actually used to build the target.

That meant that using eg.:

env.Library("foo", ["foo.cpp"], CXXFLAGS=["-Dfoo"])

Would not write the extra "-Dfoo" flag actually used to build the foo
library to the compile_commands.json file.

Instead save the environment used to generate the target and use that
when writing the compile_command.json entry.

This probably won't benefit the mongodb project (at least for now), but this should be more correct I hope.

### DIFF
--- a/site_scons/site_tools/compilation_db.py
+++ b/site_scons/site_tools/compilation_db.py
@@ -51,7 +51,8 @@ def makeEmitCompilationDbEntry(comstr):
             source=[],
             __COMPILATIONDB_UTARGET=target,
             __COMPILATIONDB_USOURCE=source,
-            __COMPILATIONDB_UACTION=user_action)
+            __COMPILATIONDB_UACTION=user_action,
+            __COMPILATIONDB_ENV=env)
 
         # TODO: Technically, these next two lines should not be required: it should be fine to
         # cache the entries. However, they don't seem to update properly. Since they are quick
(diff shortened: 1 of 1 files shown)


acmorrow:
@laudrup - Nice catch! Actually, I think this *does* probably affect the correctness of the compilation database for MongoDB. We definitely have varying sub-environments for some sets of files.

The change itself looks good. A few requests before we can merge it though:
- Could you please file a ticket in the MongoDB server project: https://jira.mongodb.org/browse/SERVER/
- Then amend the commit so the commit message begins with the SERVER ticket (see the git history for examples).
- Please sign the MongoDB contributors agreement: https://www.mongodb.com/legal/contributor-agreement. I actually think this might be unnecessary since these files are Apache, so if this is any sort of problem for you we can try to work around it, but if you are willing to sign it then we can just take the commit immediately.

Thanks again for finding this issue and making a PR!

laudrup:
> Could you please file a ticket in the MongoDB server project

Done:
https://jira.mongodb.org/browse/SERVER-27628

> Then amend the commit so the commit message begins with the SERVER ticket

Done

> Please sign the MongoDB contributors agreement

Almost done. I'm still waiting for the email to "complete the transaction", but I don't mind "signing" it at all (what a strange definition of a digital signature BTW. :-))

> Thanks again for finding this issue and making a PR!

You are more than welcome. Thank you for making this project open source so I could use your code for generating a compilation database for my scons based project. Only happy to give something back.

acmorrow:
LGTM

PR #1138, opened 2017-03-04
Dynamic oplog sizing on WiredTiger

wolfkdy:
Jira Issue(https://jira.mongodb.org/browse/SERVER-22766)
I have signed the contribution agreement.
I have changed related unittests and made them passed.
I have run intergration-tests to ensure the correctness under high io-pressure.
The main idea is to reinit parameters related to oplogStones when command resizeOplog is applied.


wolfkdy:
I did not pass “resizeOplog” through the replication subsystem since oplogSize should be metadata of a single mongod rather than a repl-set.

milkie:
Hi,
We've scheduled the work to review your patch in our next sprint, starting March 27.  We hope to give you feedback soon!
-Eric

src/mongo/db/exec/geo_near.cpp
```diff
@@ -352,6 +352,14 @@ void GeoNear2DStage::DensityEstimator::buildIndexScan(OperationContext* txn,
     _children->emplace_back(_indexScan);
 }
 
+// NOTE(deyukong): DensityEstimator is used to calculate the radius of the first
```
GeertBosch: The changes in this file do not belong in the pull request for the oplog size change.

src/mongo/db/commands/resize_oplog.cpp
```diff
@@ -0,0 +1,111 @@
...
+class CmdResizeOplog : public Command {
+public:
+    virtual bool slaveOk() const {
+        return true;
+    }
+    virtual bool supportsWriteConcern(const BSONObj& cmd) const override {
+        return false;
+    }
+    virtual void help(stringstream& help) const {
+        help << "resize oplog size, only support wiredTiger";
```
GeertBosch: Please leave of the ", only support wiredTiger", as it is possible to compile MongoDB with different storage engines, and they may well decide to support changing oplog size.

wolfkdy:
@milkie , Almost 2 months past, do you guys have any plans for this pr?

src/mongo/db/commands/resize_oplog.cpp
```diff
@@ -0,0 +1,111 @@
...
+            return appendCommandStatus(result, Status(ErrorCodes::NamespaceNotFound, "ns does not exist"));
+        }
+        if (!coll->isCapped()) {
+            return appendCommandStatus(result, Status(ErrorCodes::InternalError, "ns does not exist"));
+        }
+        if (!jsobj["size"].isNumber()) {
+            return appendCommandStatus(result, Status(ErrorCodes::InvalidOptions, "invalid size field, size should be a number"));
+        }
+
+        long long size = jsobj["size"].numberLong();
```
GeertBosch: The minimum size is 990 MB (`990LL * 1024 * 1024` bytes, see `oplog.cpp`), so that should be checked. Would be even better to move that constant to `oplog.h`.

src/mongo/db/storage/wiredtiger/wiredtiger_record_store.cpp
```diff
@@ -220,6 +226,7 @@ void WiredTigerRecordStore::OplogStones::createNewStoneIfNeeded(RecordId lastRec
         return;
     }
 
+    LOG(1) << "create new oplogStone, current stones:" << _stones.size();
```
GeertBosch: At log level 1, we shouldn't really talk about "oplogStone", as that is not a concept that we expect users to understand, so move this to log level 2.

src/mongo/db/storage/wiredtiger/wiredtiger_record_store.cpp
```diff
@@ -277,15 +284,15 @@ void WiredTigerRecordStore::OplogStones::setMinBytesPerStone(int64_t size) {
 
 void WiredTigerRecordStore::OplogStones::setNumStonesToKeep(size_t numStones) {
     invariant(numStones > 0);
-
+    return;
```
GeertBosch: ?

src/mongo/db/storage/wiredtiger/wiredtiger_record_store_oplog_stones.h
```diff
@@ -129,12 +136,8 @@ class WiredTigerRecordStore::OplogStones {
     // database, and false otherwise.
     bool _isDead = false;
 
-    // Maximum number of stones to keep in the deque before the background reclaim thread should
-    // truncate the oldest ones. Does not include the stone currently being filled. This value
-    // should not be changed after initialization.
-    size_t _numStonesToKeep;
     // Minimum number of bytes the stone being filled should contain before it gets added to the
-    // deque of oplog stones. This value should not be changed after initialization.
```
GeertBosch: I'm still unclear what the repercussions are of changing this value to the overall logic. Can you give me an argument for why this is OK despite the original warning?

src/mongo/db/storage/wiredtiger/wiredtiger_record_store_test.cpp
```diff
@@ -1449,71 +1453,27 @@ TEST(WiredTigerRecordStoreTest, OplogStones_ReclaimStones) {
...
+        ASSERT_EQ(190, rs->dataSize(opCtx.get()));
+        ASSERT_EQ(1U, oplogStones->numStones());
         ASSERT_EQ(1, oplogStones->currentRecords());
         ASSERT_EQ(50, oplogStones->currentBytes());
     }
 }
 
-// Verify that oplog stones are not reclaimed even if the size of the record store exceeds
-// 'cappedMaxSize'.
-TEST(WiredTigerRecordStoreTest, OplogStones_ExceedCappedMaxSize) {
```
GeertBosch: Why is this test no longer applicable?

wolfkdy:
@GeertBosch  I've sent you an email to explain my modifications.
Based on your comments, I've updated my code and made a new commit.
Hoping for your review, thanks.

src/mongo/db/storage/wiredtiger/wiredtiger_record_store_test.cpp
wolfkdy: because in the original implementation, there are two benchmarks: oplogStoneNum  and cappedMaxSize.  In my 
new implementation, only cappedMaxSize is considered,  the oplogStoneNum is no longer used. I have made TEST(WiredTigerRecordStoreTest, OplogStones_ReclaimStones) passed, which in the new implementation is duplicated with TEST(WiredTigerRecordStoreTest, OplogStones_ReclaimStones)

src/mongo/db/storage/wiredtiger/wiredtiger_record_store_oplog_stones.h
wolfkdy: because in the original implementation, each oplog should have the same size which is calculated in the initialization of OplogStones, it is easy to check if oplogs exceeded the limit size by comparing _numStonesToKeep and currentOplogStones.size(). However, if oplogSize is changed, the new Stones should have a different size with the remaining ones, re-arranging remaining oplogStones costs too much so it can't be a possible way. So we have to iterate and sum size of each oplogStone, including the remaining ones and the newer ones, to check if oplog size exceeded.That's why I remove _numStonesToKeep and rewrite hasExcessStones into hasExcessStones_inlock

zydcom:
@wolfkdy @GeertBosch  I think we should use `collMod` command to support dynamic oplog resize, we just need to add a new option, as I mentioned in [pr-1069] (https://github.com/mongodb/mongo/pull/1069/commits) before.

src/mongo/db/commands/resize_oplog.cpp
```diff
@@ -0,0 +1,111 @@
...
+        }
+        WriteUnitOfWork wunit(opCtx);
+        Status status = coll->getRecordStore()->updateCappedSize(opCtx, size);
+        if (!status.isOK()) {
+            return appendCommandStatus(result, status);
+        }
+        CollectionCatalogEntry* entry = coll->getCatalogEntry();
+        entry->updateCappedSize(opCtx, size);
+        wunit.commit();
+        LOG(1) << "resizeOplog success, currentSize:" << size;
```
stbrody: This should probably be at the default log level, it's a rare but significant event

wolfkdy:
@stbrody 
I've carefully read your comments and refined the code.
a new action: resizeOplog is added and privilege is checked in resizeoplog command.
I've compiled with no warnings, run mongod, created users with different roles to check if everything works well.

src/mongo/db/commands/resize_oplog.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    virtual bool supportsWriteConcern(const BSONObj& cmd) const override {
+        return false;
+    }
+
+    virtual void help(stringstream& help) const {
+        help << "resize oplog size";
+    }
+
+    virtual void addRequiredPrivileges(const std::string& dbname,
```
stbrody: addRequiredPrivileges is deprecated, please use checkAuthForCommand instead.

wolfkdy:
@stbrody I refined my code, to update codes below:
1) use checkAuthForCommand instead of addRequiredPrivileges 
2) merge from upstream/master, resolve conflicts, use new api to get collection
3) do a fully new compile with no errors and check everything works ok.
I think it is the right time to merge the code. If time lasts longer, new conflicts will appear.

src/mongo/db/commands/resize_oplog.cpp
```diff
@@ -0,0 +1,111 @@
...
+        help << "resize oplog size";
+    }
+    CmdResizeOplog() : Command("resizeOplog") {}
+    bool run(OperationContext* opCtx,
+            const string& dbname,
+            BSONObj& jsobj,
+            int,
+            string& errmsg,
+            BSONObjBuilder& result) {
+        StringData dbName("local");
```
wolfkdy: A good suggestion
The uniform code style helps. I'll refine it.
You can also post your code style on mongo official site, maybe a cpplint. That helps contributors from community.

wolfkdy:
@GeertBosch 
Thanks for taking time.
I've read your comments and taken your suggestions.
You may take a look at my latest changes.
Hoping for your reply

src/mongo/db/commands/resize_oplog.cpp
```diff
@@ -0,0 +1,111 @@
...
+        AutoGetDb autoDb(opCtx, dbName, MODE_X);
```
zydcom: @GeertBosch 
`MODE_X` for `local database` blocks  all write requests,  would you please make a more detailed explain for using `GlobalWrite` lock?

src/mongo/db/commands/resize_oplog.cpp
zydcom: Use default log level and record the client ip adress would be better for trouble shooting.

src/mongo/db/commands/resize_oplog.cpp
zydcom: Agree with @wolfkdy , detailed code style will really helpful for contributors, `StringData` and `NameString` are both used in current code base to refer collection names.

GeertBosch:
This has now been merged. Thanks @wolfkdy for your contribution and patience!

wolfkdy:
@GeertBosch thanks for your time and help.

PR #1140, opened 2017-03-08
Verbosity bug

ryankall:
now the Explain command will not accept unknown command like verbos: 'queryPlanner'

### DIFF
--- a/jstests/core/explain_find_and_modify.js
+++ b/jstests/core/explain_find_and_modify.js
@@ -47,7 +47,8 @@
     assert.commandWorked(db.runCommand(explainUpsert));
     assertCollDoesNotExist(cName, "explaining an upsert should not create a new collection.");
 
-    assert.commandWorked(db.runCommand(Object.merge(explainUpsert, {fields: {x: 1}})));
+    explainUpsert['explain']['fields'] =  {x: 1};    
+    assert.commandWorked(db.runCommand(explainUpsert));
     assertCollDoesNotExist(cName, "explaining an upsert should not create a new collection.");
 
     // 3. Explaining findAndModify should not work with an invalid findAndModify command object.
--- a/jstests/core/explain_verbosity_unknown.js
+++ b/jstests/core/explain_verbosity_unknown.js
@@ -1,0 +1,22 @@
+//Test whether unknown commands are accepted in explain function
+
+(function() {
+	
+	"use strict";
+	
+	var unknownCollection = db["jstestDB"];
+
+	//insure the collection has an item
+	unknownCollection.insert({a:1});
+
+	//check with finding item that exist
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, verbosity: 'queryPlanner', unknown: 1}), "unknown argument command extra");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'queryPlanner'}), "unknown verbosity command");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'excutionStats'}), "unknown verbosity command");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'allPlansExecution'}), "unknown verbosity command");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'unknown'}), "unknown argument and command");
+
+	//empty collection
+	unknownCollection.drop();
+
+})();
--- a/src/mongo/db/query/explain_common.cpp
+++ b/src/mongo/db/query/explain_common.cpp
@@ -31,7 +31,7 @@
 #include "mongo/db/query/explain_common.h"
 
 #include "mongo/util/mongoutils/str.h"
-#include <string>
+
 namespace mongo {
 
 // static
@@ -59,7 +59,7 @@ Status ExplainCommon::parseCmdBSON(const BSONObj& cmdObj, ExplainCommon::Verbosi
     *verbosity = ExplainCommon::EXEC_ALL_PLANS;
     if (cmdObj.nFields() > 2) {
         return Status(ErrorCodes::BadValue, "explain allows only 2 objects: explain: {<command>}, verbosity: <excution_call>");
-    } else if (cmdObj.nFields() == 1 || cmdObj.nFields() == 0) {
+    } else if (cmdObj.nFields() == 1) {
         return Status::OK();
     } else if (!cmdObj["verbosity"].eoo()) { 
         const char* verbStr = cmdObj["verbosity"].valuestrsafe();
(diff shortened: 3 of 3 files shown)


jstests/core/explain_find_and_modify.js
```diff
@@ -47,7 +47,8 @@
     assert.commandWorked(db.runCommand(explainUpsert));
     assertCollDoesNotExist(cName, "explaining an upsert should not create a new collection.");
 
-    assert.commandWorked(db.runCommand(Object.merge(explainUpsert, {fields: {x: 1}})));
+    explainUpsert['explain']['fields'] =  {x: 1};
```
cswanson310: You can use the '.' syntax: `explainUpsert.explain.fields = {x: 1};`

jstests/core/explain_verbosity_unknown.js
```diff
@@ -0,0 +1,22 @@
+//Test whether unknown commands are accepted in explain function
+
+(function() {
+	
+	"use strict";
+	
+	var unknownCollection = db["jstestDB"];
```
cswanson310: We typically use a collection name that is identical to that of the test. In this case, I'd suggest calling the test 'jstests/core/explain_cmd_unknown_options.js', since it's a little more descriptive. You can also use '.' syntax here, and should prefer using 'let' over 'var', since 'let' has slightly better scoping rules: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let:

> let allows you to declare variables that are limited in scope to the block, statement, or expression on which it is used. This is unlike the var keyword, which defines a variable globally, or locally to an entire function regardless of block scope.

Thus, this line should read `let collection = db.explain_cmd_unknown_options;`

jstests/core/explain_verbosity_unknown.js
```diff
@@ -0,0 +1,22 @@
...
+
+	//insure the collection has an item
+	unknownCollection.insert({a:1});
+
+	//check with finding item that exist
```
cswanson310: How about `// Test that the explain command rejects unknown options when the collection exists.`?

jstests/core/explain_verbosity_unknown.js
```diff
@@ -0,0 +1,22 @@
...
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, verbosity: 'queryPlanner', unknown: 1}), "unknown argument command extra");
```
cswanson310: Sorry for misleading you earlier, but let's use the more-specific `assert.commandFailedWithCode()`: https://github.com/mongodb/mongo/blob/r3.5.3/src/mongo/shell/assert.js#L382

After my comment below, this should fail with the code FailedToParse: https://github.com/mongodb/mongo/blob/r3.5.3/src/mongo/base/error_codes.err

In javascript, you can access this by using `ErrorCodes.FailedToParse`, instead of using the number.

jstests/core/explain_verbosity_unknown.js
```diff
@@ -0,0 +1,22 @@
...
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'queryPlanner'}), "unknown verbosity command");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'excutionStats'}), "unknown verbosity command");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'allPlansExecution'}), "unknown verbosity command");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'unknown'}), "unknown argument and command");
+
+	//empty collection
+	unknownCollection.drop();
```
cswanson310: Move this line to the top of the file, to ensure the collection doesn't already exist when we start. This is more consistent with how we do it in other tests.

jstests/core/explain_verbosity_unknown.js
```diff
@@ -0,0 +1,22 @@
...
+	"use strict";
+	
+	var unknownCollection = db["jstestDB"];
+
+	//insure the collection has an item
+	unknownCollection.insert({a:1});
+
+	//check with finding item that exist
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, verbosity: 'queryPlanner', unknown: 1}), "unknown argument command extra");
+	assert.commandFailed(db.runCommand({explain: {find: 'a'}, unknownArg: 'queryPlanner'}), "unknown verbosity command");
```
cswanson310: Each of these are executing a query of the collection named 'a'. I think you meant to execute them on the collection mentioned on line 7 above, so you should use `collection.getName()` instead of `'a'`.

src/mongo/db/query/explain_common.cpp
```diff
@@ -51,12 +51,17 @@ const char* ExplainCommon::verbosityString(ExplainCommon::Verbosity verbosity) {
 
 // static
 Status ExplainCommon::parseCmdBSON(const BSONObj& cmdObj, ExplainCommon::Verbosity* verbosity) {
+
```
cswanson310: Remove extra newline.

src/mongo/db/query/explain_common.cpp
```diff
@@ -67,9 +72,11 @@ Status ExplainCommon::parseCmdBSON(const BSONObj& cmdObj, ExplainCommon::Verbosi
                           "verbosity string must be one of "
                           "{'queryPlanner', 'executionStats', 'allPlansExecution'}");
         }
+        return Status::OK();
+    } else {
+        return Status(ErrorCodes::BadValue, "verbosity invalid command");
```
cswanson310: Let's use `ErrorCodes::FailedToParse` instead.

src/mongo/db/query/explain_common.cpp
```diff
@@ -51,12 +51,17 @@ const char* ExplainCommon::verbosityString(ExplainCommon::Verbosity verbosity) {
...
 // static
 Status ExplainCommon::parseCmdBSON(const BSONObj& cmdObj, ExplainCommon::Verbosity* verbosity) {
+
     if (Object != cmdObj.firstElement().type()) {
         return Status(ErrorCodes::BadValue, "explain command requires a nested object");
     }
 
     *verbosity = ExplainCommon::EXEC_ALL_PLANS;
-    if (!cmdObj["verbosity"].eoo()) {
+    if (cmdObj.nFields() > 2) {
```
cswanson310: Instead of checking how many fields there are, let's loop over all of the fields and make sure that each one is recognized. I think this will be a little clearer. Here's some example code that does a similar thing for an aggregation command (https://docs.mongodb.com/manual/reference/command/aggregate/): https://github.com/mongodb/mongo/blob/r3.5.3/src/mongo/db/pipeline/aggregation_request.cpp#L91-L166

I think you will also have to do something similar and ignore the following field names, which will be parsed elsewhere: "explain", anything that starts with a "$", and that might be it. But I'll check with others here and get back to you.

PR #1141, opened 2017-03-30
SERVER-25980 Merge parallel vector to a single array.

ryankall:
I ran the auto formatter and many parts of the code got moved around. Unsure if it correct code style because of that formatter.


src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -66,7 +66,7 @@ DocumentSource::GetNextResult DocumentSourceGroup::getNext() {
         invariant(initializationResult.isEOF());
     }
 
-    for (auto&& accum : _currentAccumulators) {
+    for (auto && accum : _currentAccumulators) {
```
cswanson310: It does look like you've somehow got the wrong formatting. Let's chat about what you did to format the code.

src/mongo/db/pipeline/document_source_group.h
```diff
@@ -147,15 +147,21 @@ class DocumentSourceGroup final : public DocumentSource, public SplittableDocume
      */
     Value expandId(const Value& val);
 
+    struct Field {
+        std::string _fieldName;
```
cswanson310: Since this is a simple struct, and the member variables are public, the names should not start with an _.

jstests/core/explain_cmd_unknown_options.js
```diff
@@ -0,0 +1,18 @@
+//Test whether unknown commands are accepted in explain function
```
cswanson310: Let's talk about how to make sure these unrelated changes don't show up here.

cswanson310:
New pull request with just changes for SERVER-25980 (without patch for explain options) is here: https://github.com/mongodb/mongo/pull/1142. Closing.

PR #1142, opened 2017-04-03
SERVER-25980 Converted three parallel vectors of DocumentSourceGroup into a single…

ryankall:
… array of structs

Code style is possibly wrong base on my auto formatter.


ryankall:
Code style fixed!

PR #1143, opened 2017-04-03
SERVER-25980 Converted three parallel vectors of DocumentSourceGroup …

ryankall:
…into a single array of structs


src/mongo/db/pipeline/document_source_group.h
```diff
@@ -147,15 +147,21 @@ class DocumentSourceGroup final : public DocumentSource, public SplittableDocume
...
+        Accumulator::Factory accumulatorFactory;
+        boost::intrusive_ptr<Expression> expression;
+    };
+
     /**
-     * 'vFieldName' contains the field names for the result documents, 'vpAccumulatorFactory'
-     * contains the accumulator factories for the result documents, and 'vpExpression' contains the
-     * common expressions used by each instance of each accumulator in order to find the right-hand
-     * side of what gets added to the accumulator. These three vectors parallel each other.
+     * '_fieldName' contains the field names for the result documents, '_accumulatorFactory'
```
cswanson310: Let's split this comment into three, one about each data member, and put those three comments next to their corresponding fields above. Also, we now tend to prefer the `// Comment` syntax for comments describing member variables (I think), so if you could update the `/** comment */` to be `// comments`, that would be great.

Edit: we already have a class for this! Please move these comments to https://github.com/mongodb/mongo/blob/r3.5.3/src/mongo/db/pipeline/accumulation_statement.h, inline with 'fieldName', 'factor', and 'expression' there.

src/mongo/db/pipeline/document_source_group.h
```diff
@@ -147,15 +147,21 @@ class DocumentSourceGroup final : public DocumentSource, public SplittableDocume
      */
     Value expandId(const Value& val);
 
+    struct Field {
```
cswanson310: Let's call this `AccumulatedField`, to be more specific. `Field` is pretty generic.

Edit: we already have this struct: https://github.com/mongodb/mongo/blob/r3.5.3/src/mongo/db/pipeline/accumulation_statement.h! Let's just use that.

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -191,8 +191,8 @@ intrusive_ptr<DocumentSource> DocumentSourceGroup::optimize() {
         _idExpressions[i] = _idExpressions[i]->optimize();
     }
 
-    for (size_t i = 0; i < vFieldName.size(); i++) {
-        vpExpression[i] = vpExpression[i]->optimize();
+    for (size_t i = 0; i < _accumulatedFields.size(); i++) {
```
cswanson310: Similarly, you're only using `i` to index into `_accumulatedFields`, so convert this to a range-based loop: `for (auto&& accumulatedField : _acumulatedFields) {...}`

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -216,11 +216,12 @@ Value DocumentSourceGroup::serialize(boost::optional<ExplainOptions::Verbosity>
     }
 
     // Add the remaining fields.
-    const size_t n = vFieldName.size();
+    const size_t n = _accumulatedFields.size();
     for (size_t i = 0; i < n; ++i) {
```
cswanson310: Ditto rang-based for loop. You don't need `n` anymore either.

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -242,9 +243,9 @@ DocumentSource::GetDepsReturn DocumentSourceGroup::getDependencies(DepsTracker*
     }
 
     // add the rest
-    const size_t n = vFieldName.size();
+    const size_t n = _accumulatedFields.size();
     for (size_t i = 0; i < n; ++i) {
```
cswanson310: Same here.

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -320,7 +319,7 @@ void DocumentSourceGroup::setIdExpression(const boost::intrusive_ptr<Expression>
     if (auto object = dynamic_cast<ExpressionObject*>(idExpression.get())) {
         auto& childExpressions = object->getChildExpressions();
         invariant(!childExpressions.empty());  // We expect to have converted an empty object into a
-                                               // constant expression.
+        // constant expression.
```
cswanson310: This looks like it was formatted wrong?

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -465,7 +464,7 @@ void getFieldPathListForSpilled(ExpressionObject* expressionObj,
 }  // namespace
 
 DocumentSource::GetNextResult DocumentSourceGroup::initialize() {
-    const size_t numAccumulators = vpAccumulatorFactory.size();
```
cswanson310: Ditto not using `i` and `numAccumulators` throughout this method.

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -858,8 +858,10 @@ intrusive_ptr<DocumentSource> DocumentSourceGroup::getMergeSource() {
     /* the merger will use the same grouping key */
     pMerger->setIdExpression(ExpressionFieldPath::parse(pExpCtx, "$$ROOT._id", vps));
 
-    const size_t n = vFieldName.size();
+    const size_t n = _accumulatedFields.size();
     for (size_t i = 0; i < n; ++i) {
```
cswanson310: Ditto fixing this loop.

src/mongo/db/pipeline/accumulation_statement.h
```diff
@@ -93,9 +91,18 @@ class AccumulationStatement {
      */
     static Accumulator::Factory getFactory(StringData name);
 
+    // The field name is use to store the results of the accumulation in a result document.
```
cswanson310: used

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -494,7 +491,7 @@ DocumentSource::GetNextResult DocumentSourceGroup::initialize() {
         return DocumentSource::GetNextResult::makeEOF();
     }
 
-    dassert(numAccumulators == vpExpression.size());
+    dassert(numAccumulators == _accumulatedFields.size());
```
cswanson310: This doesn't do anything useful any more! We know these are equal! You declared it as such on line 465.

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -525,7 +522,7 @@ DocumentSource::GetNextResult DocumentSourceGroup::initialize() {
             // Add the accumulators
             group.reserve(numAccumulators);
             for (size_t i = 0; i < numAccumulators; i++) {
```
cswanson310: Can you convert this loop to a range-based loop, and the ones on 531, 582 as well?

src/mongo/db/pipeline/document_source_group.cpp
```diff
@@ -661,7 +659,7 @@ boost::optional<BSONObj> DocumentSourceGroup::findRelevantInputSort() const {
     // $group contained only FieldPaths or constants. Determine if this is the case, and extract
     // those FieldPaths if it is.
     DepsTracker deps(DepsTracker::MetadataAvailable::kNoMetadata);  // We don't support streaming
-                                                                    // based off a text score.
+    // based off a text score.
```
cswanson310: Looks like this one and the one on 709 had the same formatting problem.

PR #1147, opened 2017-04-20
Visual C++ workaround

solodon4:
Workaround to allow building MongoDB with /Zc:ternary on (implied by /permissive-) under existing STL sources. Explicit cast can be removed once Visual C++ std::list stops using inheritance between const and modifiable iterators.

Jira ticket: https://jira.mongodb.org/browse/SERVER-28890 

Visual C++ compiler team is improving conformance of conditional operator for the upcoming VS2017 Update 1 release (15.3). The improvements will be available under a switch, but they will also be implied by the /permissive- switch. Our QA team currently builds regularly MongoDB as a part of RWC suite and so far MongoDB been clean under /permissive-. There is one place right now that will fail to compile under upcoming /Zc:ternary:

mongo\src\mongo\util\lru_cache.h(134):
```C++
const_iterator cfind(const K& key) const {
    auto it = this->_map.find(key); 
    return (it == this->_map.end()) ? this->end() : it->second; // << Line 134
} 
```
The problem is not so much in the code itself as other compilers are able to compile it without problems, but in our STL implementation that currently uses inheritance between mutable and const iterators. Stephan acknowledged the library team is planning to get rid of this trick and replace it with conformant implementation in the next major breaking release, but it's not going to happen anytime soon, while as is the above statement is ambiguous (according to the standard; note other compilers would have complained as well should they've been using our STL implementation).
We would like to ask you guys to patch the above code with an explicit cast to resolve the ambiguity caused by our current iterator implementation as following (pull request is following):

```C++
const_iterator cfind(const K& key) const {
    auto it = this->_map.find(key); 
    return (it == this->_map.end()) ? this->end() : const_iterator(it->second); // << Line 134
} 
```
This would allow us to keep building MongoDB clean under /permissive- for validating compiler changes.

Thank you!
Yuriy

### DIFF
--- a/src/mongo/util/lru_cache.h
+++ b/src/mongo/util/lru_cache.h
@@ -131,7 +131,7 @@ public:
      */
     const_iterator cfind(const K& key) const {
         auto it = this->_map.find(key);
-        return (it == this->_map.end()) ? this->end() : it->second;
+        return (it == this->_map.end()) ? this->end() : const_iterator(it->second);
     }
 
     /**


acmorrow:
@samantharitter and @adamlsd - I've assigned you as reviewers, PTAL.

adamlsd:
Hello Yuriy,

Thank you for bringing this to our attention.  We are very excited to learn that MongoDB's code base is used as part of Microsoft's conformance suite.  We are happy to incorporate this change in order to keep MongoDB working with your in-development compiler.  I have indeed confirmed that the ambiguity in the ternary operator occurs in inheritance based scenarios -- I was able to reproduce the failing ambiguity by writing a mock-up of Microsoft's STL list iterator.  MongoDB's development teams are committed to excellence in modern C++; we are actively participating in C++'s standardization process.

The ambiguity issue raised is incredibly subtle -- we've all learned a thing or two about C++'s `?:` operator from this, particularly that `?:`'s conversion to base class is only via reference-widening, not via slicing.  We will certainly be on the lookout for any other places this might arise in our own code

Please note that our master branch is our primary development branch, so it is likely that future commits to our in-development codebase will cause problems in your development compiler.  We will be happy to fix those issues, should you bring them to our attention.  We are looking forward to Microsoft's next compiler release.

I have some minor changes that I'd like to incorporate to this pull-request:

* I want to include a `TODO` comment above the changed line, referencing the Jira ticket we've created to track this issue.
* I want to expand the commit message to use our Jira ticket number.
* I have a small write-up of the reasons behind this issue that I'd like to incorporate in the git commit message.

I've already incorporated these into an amended copy of your original commit.  You can find it on my github account at: "https://github.com/adamlsd/mongo".  Look at the head commit in the `msft-supplied-code-fix-from-yuriy` branch.  I will also mark in this github review where those changes should go.

Cheers,

-- 
ADAM David Alan Martin
Senior Software Engineer, MongoDB
Primary Representative & Head of Delegation for MongoDB,
ISO JTC1/SC22/WG21 (International C++ Standards Committee)

src/mongo/util/lru_cache.h
```diff
@@ -131,7 +131,7 @@ class LRUCache {
      */
     const_iterator cfind(const K& key) const {
         auto it = this->_map.find(key);
-        return (it == this->_map.end()) ? this->end() : it->second;
+        return (it == this->_map.end()) ? this->end() : const_iterator(it->second);
```
adamlsd: I'd like to add a comment above here:

// TODO(SERVER-28890): Remove the function-style cast when MSVC's
//`std::list< ... >::iterator` implementation doesn't conflict with their `/Zc:ternary`

solodon4:
Looks good to me, thank you Adam (or is it David? :), sorry couldn't figure out if ADAM was recursive acronym around your full name)!

adamlsd:
My name is "Adam".  People often do get confused by my capitalization to "ADAM".  Think "GNU's Not Unix".  :-)

PR #1148, opened 2017-04-26
SERVER-28986 Improve compression negotiation logging

bjori:


solangiirfanali:
How it work

src/mongo/transport/message_compressor_manager.cpp
```diff
@@ -249,6 +251,8 @@ void MessageCompressorManager::serverNegotiate(const BSONObj& input, BSONObjBuil
             sub.append(algo->getName());
         }
         sub.doneFast();
+    } else {
+        LOG(3) << "Could not agree on compressor to use";
```
bjori: It is subtly different now, in that you'll see the "zyz is not supported" for the `compressors: [foobar]` case, while no such entry in the `compressors: []` case.

I'll update this to check theirObj before entering the for loop to see if there is anything to negotiate for

src/mongo/transport/message_compressor_manager.cpp
```diff
@@ -218,6 +218,8 @@ void MessageCompressorManager::serverNegotiate(const BSONObj& input, BSONObjBuil
                 sub.append(algo->getName());
             }
             sub.doneFast();
+        } else {
+            LOG(3) << "No compression negotiated";
```
bjori: done

PR #1154, opened 2017-06-08
SERVER-9406 treat ObjectId type as Date in aggregation date expressions

asya999:


cswanson310:
@asya999 it looks like this somehow is missing the changes to `src/mongo/db/pipeline/value.cpp`?

PR #1155, opened 2017-06-16
[SERVER-29338] Fix. Switch to use Value node instead of base Nodes

bdbaddog:
Also using .write() and .read() methods of Value nodes to store values needed for compilation db for clang tools


Schubes:
merged with https://github.com/mongodb/mongo/commit/0019d153b92ddb6a588f12d70007ddae0b7b653c

PR #1164, opened 2017-07-16
Fix  fortify source warning

greenius:
Fixes problem with warnings when _FORTIFIED_SOURCE is already defined by the compiler by undefining it on the command line before defining it again.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -2480,8 +2480,20 @@ def doConfigure(myenv):
             context.Result(ret)
             return ret
 
+        def CheckForGlibcDefinesFortify(context):
+            test_body="""
+            #ifndef _FORTIFY_SOURCE
+            #error
+            #endif
+            """
+            context.Message('Checking for predefined _FORTIFY_SOURCE...')
+            ret = context.TryCompile(textwrap.dedent(test_body), ".c")
+            context.Result(ret)
+            return ret
+
         conf = Configure(myenv, help=False, custom_tests = {
             'CheckForFortify': CheckForGlibcKnownToSupportFortify,
+            'CheckForFortifyDefined': CheckForGlibcDefinesFortify,
         })
 
         # Fortify only possibly makes sense on POSIX systems, and we know that clang is not a valid
@@ -2490,6 +2502,10 @@ def doConfigure(myenv):
         # http://lists.llvm.org/pipermail/cfe-dev/2015-November/045852.html
         #
         if env.TargetOSIs('posix') and not env.ToolchainIs('clang') and conf.CheckForFortify():
+            if conf.CheckForFortifyDefined():
+                conf.env.Prepend(
+                    CPPFLAGS='-U_FORTIFY_SOURCE'
+                )
             conf.env.Append(
                 CPPDEFINES=[
                     ('_FORTIFY_SOURCE', 2),


greenius:
This also prevents the "Compiler must support a thread local storage class for trivially constructible types" error that you get on systems that automatically set _FORTIFIED_SOURCE without  using --disable-warnings-as-errors command line option.

acmorrow:
@greenius - Thanks for the PR. This was on my queue to fix (it is currently ticketed as https://jira.mongodb.org/browse/SERVER-29982), so I appreciate your sharing your fix with us. A few things we will want to do:

- Squash to one commit
- Prefix the commit message with SERVER-29982 (see the git commit log for formatting examples)
- If you have not already done so, please sign the contributor agreement: https://www.mongodb.com/legal/contributor-agreement

I'll take a closer look at the changes themselves later today or tomorrow.

greenius:
@acmorrow I have now squashed the commits, changed the message and signed the contributor agreement.

SConstruct
```diff
@@ -2490,6 +2502,10 @@ def doConfigure(myenv):
         # http://lists.llvm.org/pipermail/cfe-dev/2015-November/045852.html
         #
         if env.TargetOSIs('posix') and not env.ToolchainIs('clang') and conf.CheckForFortify():
+            if conf.CheckForFortifyDefined():
+                conf.env.Prepend(
+                    CPPFLAGS='-U_FORTIFY_SOURCE'
```
acmorrow: Rather than just having -U_FORTIFY_SOURCE -DFORTIFY_SOURCE, which feels a little strange, would it be better if we just didn't set our own _FORTIFY_SOURCE if the compiler has already done it?

In other words, just write:

```
if !conf.CheckForForifyDefined():
    conf.env.Append(
        CPPDEFINES=[
            ('_FORTIFY_SOURCE', 2)
```

greenius: That's what I did initially, but I think the default predefined value for _FORTIFY_SOURCE is 1 and this is setting it to 2, which enables more checking.

From http://man7.org/linux/man-pages/man7/feature_test_macros.7.html it says:

> If _FORTIFY_SOURCE is set to 1, with compiler optimization level 1 (gcc -O1) and above, checks that shouldn't change the behavior of conforming programs are performed.  With _FORTIFY_SOURCE set to 2, some more checking is added, but some conforming programs might fail.

acmorrow: Having thought about this a bit more, this really feels like a bug in the packaging of the compiler on this system. I'm fine with the distro toolchain wanting to default _FORTIFY_SOURCE=1, but it shouldn't cause a compilation error if I then chose to set _FORTIFY_SOURCE=2 on the command line, as I'm entirely entitled to do. That should simply cause the compiler to honor my choice rather than its default, not raise an error. While I'd be willing to pull this PR as a workaround, I think it might be better to try to first raise this as an issue with your distro toolchain maintainer.

acmorrow: @greenius - FYI I've added a similar comment to https://jira.mongodb.org/browse/SERVER-29982, if you want to follow along there.

Schubes:
I'm closing this pull request per [Drew's comment](https://jira.mongodb.org/browse/SERVER-29982?focusedCommentId=1650776&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1650776) on SERVER-29982

PR #1166, opened 2017-07-17
SERVER-30166 Replace string with explicit std::string. Enables compul…

greenius:
Replaces string with std::string in several files so it will build when using --using-system-pcre with libpcre 8.41.

This is currently preventing Mongo from building for Gentoo linux (See https://bugs.gentoo.org/show_bug.cgi?id=625148)

### DIFF
--- a/src/mongo/db/dbwebserver.cpp
+++ b/src/mongo/db/dbwebserver.cpp
@@ -67,6 +67,7 @@ namespace mongo {
 using std::map;
 using std::stringstream;
 using std::vector;
+using std::string;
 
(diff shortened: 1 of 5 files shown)


greenius:
This is the patch to libpcre that caused this problem to show itself... https://bugs.exim.org/show_bug.cgi?id=2084

Several using std directives (including "using std::string") were declared in global namespace in header file <pcre_stringpiece.h> which was included from <pcrecpp.h>. They have been moved into the pcrecpp namespace so as not to pollute the global namespace.

PR #1167, opened 2017-07-17
SERVER-30166 Replace string with explicit std::string. Enables compul…

greenius:
Replaces string with std::string in several files so it will build when using --using-system-pcre with libpcre 8.41.

This is the equivalent of Pull Request #1166 but applies to master instead of 3.4 branch.


markbenvenuto:
@greenius Thanks for the contribution

PR #1169, opened 2017-07-29
Server 25311

ryankall:


src/mongo/s/commands/cluster_count_cmd.cpp
```diff
@@ -47,9 +47,9 @@
 namespace mongo {
 namespace {
 
-class ClusterCountCmd : public ErrmsgCommandDeprecated {
+class ClusterCountCmd : public Command {
```
dstorch: I think this should inherit from `BasicCommand` rather than directly from `Command`. This is a `BasicCommand` in the sense that it does not support document sequences. (Batch write commands like insert, update, and delete are examples of commands that accept a sequence of documents as input.) Therefore, this could benefit from the simplified `run()` signature provided by `BasicCommand`

src/mongo/s/commands/cluster_count_cmd.cpp
```diff
@@ -71,29 +71,19 @@ class ClusterCountCmd : public ErrmsgCommandDeprecated {
...
-            if (skip < 0) {
-                errmsg = "skip value is negative in count query";
-                return false;
-            }
-        } else if (cmdObj["skip"].ok()) {
-            errmsg = "skip value is not a valid number";
-            return false;
+        const bool isExplain = true;
+        auto request = CountRequest::parseFromBSON(dbname, cmdObj, isExplain);
+        if (!request.isOK()) {
```
dstorch: Might as well just do

`uassertStatusOK(request.getStatus())`

We're trying to get a little more throw-y and a little less Status-y when it comes to propagating errors.

src/mongo/s/commands/cluster_count_cmd.cpp
```diff
@@ -71,29 +71,19 @@ class ClusterCountCmd : public ErrmsgCommandDeprecated {
...
+            return request.getStatus();
         }
 
+
```
dstorch: Commenting here because I can't comment on lines not changed in the diff...

There is still a lot of parsing code below for the "query", "collation", "limit", and "skip" fields that hasn't been removed. This means that we are currently double-parsing the request: first with `CountRequest::parseFromBSON()`, and next with `cmdObj["query"]`, `bsonExtractTyedField(cmdObj, "collation", ...)`, etc.

src/mongo/s/commands/cluster_count_cmd.cpp
```diff
@@ -184,11 +172,12 @@ class ClusterCountCmd : public ErrmsgCommandDeprecated {
                 resolvedView.asExpandedViewAggregation(aggRequestOnView.getValue());
             auto resolvedAggCmd = resolvedAggRequest.serializeToCommandObj().toBson();
 
-            BSONObj aggResult = Command::runCommandDirectly(
-                opCtx, OpMsgRequest::fromDBAndBody(dbname, std::move(resolvedAggCmd)));
+            BSONObjBuilder aggResult;
+            Command::findCommand("aggregate")
```
dstorch: Why did this line change? Is this just the result of merging with HEAD? Looks like a bad merge, as I think we should continue to use `Command::runCommandDirectly()`.

src/mongo/s/commands/cluster_count_cmd.cpp
```diff
@@ -270,16 +259,14 @@ class ClusterCountCmd : public ErrmsgCommandDeprecated {
...
-                                              dbname,
-                                              nss,
-                                              explainCmd,
-                                              ReadPreferenceSetting::get(opCtx),
-                                              ShardTargetingPolicy::UseRoutingTable,
-                                              targetingQuery,
-                                              targetingCollation,
-                                              true,  // do shard versioning
-                                              &viewDefinition);
+        auto swShardResponses = scatterGatherForNamespace(opCtx,
```
dstorch: Same here, maybe a bad merge?

dstorch:
Hi @ryankall ,

It looks like we haven't heard from you in a few months, so I'm going to close this pull request. If you're interested in addressing my few comments above and getting your code merged in the future, I'd be happy to re-open!

Best,
Dave

PR #1170, opened 2017-08-12
SERVER-8164 ISODate returns incorrect values for dates of than 100 years

a-h:
The ISODate incorrectly parses dates such as:

```
0001-01-01T00:00:00Z
0002-01-01T00:00:00Z
...
0099-01-01T00:00:00Z
```

If you insert data via the MongoDB shell:

```
> use test
switched to db test
> db.tests.insert({ d: ISODate('0001-01-01T00:00:00Z') })
> db.tests.findOne()
{
	"_id" : ObjectId("598ed923e9d74c3320ec1b39"),
	"d" : ISODate("1901-01-01T00:00:00Z")
}
```

The expectation is that `d` has a year value of 0001, not 1901.

### DIFF
--- a/jstests/core/isodate.js
+++ b/jstests/core/isodate.js
@@ -1,0 +1,48 @@
+var tests = 
+[
+    {
+        input: '0001-01-01T00:00:00Z',
+        expected: {
+            year: 1,
+            month: 0,
+            day: 1,
+            hour: 0,
+            minute: 0,
+            second: 0,
+        },
+    },
+    {
+        input: '2001-02-03T12:34:56',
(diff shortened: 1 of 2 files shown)


derickr:
We have already a patch in (internal) code review as the status for https://jira.mongodb.org/browse/SERVER-8164 indicates, which solves this a little more elegantly: https://github.com/mongodb/mongo/compare/master...derickr:SERVER-8164-isodate-shell-parsing

We expect that patch to be merged relatively soon, and as such, I am closing this PR. Thanks again for having a look at this!

PR #1172, opened 2017-08-31
SERVER-30914: Repair basic builds on OpenBSD

i80and:


markbenvenuto:
@i80and Thanks for the contribution

PR #1175, opened 2017-09-17
Make string.cpp compatible with boost 1.62

jelly:
Add a conditional for boost > 1.62 compatibility, boyer_moore_search
now returns an std::pair. Fixes https://jira.mongodb.org/browse/SERVER-31119


acmorrow:
Merged in 8a5656c58c18e7c23d5bec6f9b2d41013ee80a47

PR #1177, opened 2017-09-26
SERVER-31272: Include sys/resource.h on Unix

i80and:


src/mongo/transport/service_entry_point_utils.cpp
```diff
@@ -39,7 +39,7 @@
 #include "mongo/util/debug_util.h"
 #include "mongo/util/log.h"
 
-#ifdef __linux__  // TODO: consider making this ifndef _WIN32
+#ifndef _WIN32
```
acmorrow: Can you write this like the one below, using `#if defined` so they match up a little better?

i80and: Done!

markbenvenuto:
Thanks @i80and for the contribution.

PR #1186, opened 2017-10-26
Fix build.

judu:
Without these using `std::string`, mongodb does not build here.

I'm using gcc 6.4.0.

Please ask if you need more information about my build :)


acmorrow:
Hmm. I think this was already done in b7cebf82bc5e5572308cc3a3fb958a825db91498 (PR 1167). Could you please pull the newest version of master and see if this is still an issue for you?

PR #1192, opened 2017-12-11
SERVER-32344 fixed test crash on null pointer reference

apocarteres:
I recently checked out the master branch and noticed that ```dbtests``` do crash due to NPE
```
017-12-11T10:04:32.613+0300 I -        [testsuite] going to run suite: validate_tests
2017-12-11T10:04:32.613+0300 I -        [testsuite] 	 going to run test: ValidateTests::ValidateIdIndexCount<true, false>
ASAN:DEADLYSIGNAL
    #3 0x111b80c0c in mongo::unittest::Suite::run(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) unittest.cpp:354
    #4 0x10d21d821 in mongo::dbtests::runDbTests(int, char**) framework.cpp:95
    #5 0x10c71d5be in dbtestsMain(int, char**, char**) dbtests.cpp:162
    #6 0x10c71dc6a in main dbtests.cpp:178
    #7 0x7fff8d2c55ac in start (libdyld.dylib:x86_64+0x35ac)

==17794==Register values:
rax = 0x0000100000000002  rbx = 0x00007fff535759e0  rcx = 0x0000100000000000  rdx = 0x0000100000000000  
rdi = 0x0000000000000010  rsi = 0x0001100000011000  rbp = 0x00007fff53575a90  rsp = 0x00007fff535758e0  
 r8 = 0x00000000fffffffc   r9 = 0x0000613000421648  r10 = 0x00000000ffffffff  r11 = 0xffffffff00000000  
r12 = 0x00007fff53575940  r13 = 0x00007fff535758e0  r14 = 0x00007fff53575ac0  r15 = 0x00001fffea6aeb22  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV memory:2580 in ValidateTests::ValidateBase::ValidateBase(bool, bool)
==17794==ABORTING
2017-12-11T10:04:51.816+0300 F -        [testsuite] Got signal: 6 (Abort trap: 6).
U
```

so i ultimately found 2 places where bugs like NPE appeared. Please review

### DIFF
--- a/src/mongo/db/catalog/database_impl.cpp
+++ b/src/mongo/db/catalog/database_impl.cpp
@@ -600,10 +600,7 @@ Status DatabaseImpl::dropCollectionEvenIfSystem(OperationContext* opCtx,
 
     // Register this drop-pending namespace with DropPendingCollectionReaper to remove when the
     // committed optime reaches the drop optime.
-    auto &&pReaper = repl::DropPendingCollectionReaper::get(opCtx);
-    if (pReaper != nullptr) {
-        pReaper->addDropPendingNamespace(dropOpTime, dpns);
-    }
+    repl::DropPendingCollectionReaper::get(opCtx)->addDropPendingNamespace(dropOpTime, dpns);
     return Status::OK();
 }
 
--- a/src/mongo/db/catalog/database_impl.cpp
+++ b/src/mongo/db/catalog/database_impl.cpp
@@ -601,6 +601,7 @@ Status DatabaseImpl::dropCollectionEvenIfSystem(OperationContext* opCtx,
     // Register this drop-pending namespace with DropPendingCollectionReaper to remove when the
     // committed optime reaches the drop optime.
     repl::DropPendingCollectionReaper::get(opCtx)->addDropPendingNamespace(dropOpTime, dpns);
+
     return Status::OK();
 }
 
--- a/src/mongo/db/catalog/database_impl.cpp
+++ b/src/mongo/db/catalog/database_impl.cpp
@@ -600,8 +600,10 @@ Status DatabaseImpl::dropCollectionEvenIfSystem(OperationContext* opCtx,
 
     // Register this drop-pending namespace with DropPendingCollectionReaper to remove when the
     // committed optime reaches the drop optime.
-    repl::DropPendingCollectionReaper::get(opCtx)->addDropPendingNamespace(dropOpTime, dpns);
-
+    repl::DropPendingCollectionReaper *pReaper = repl::DropPendingCollectionReaper::get(opCtx);
+    if (pReaper != nullptr) {
+        pReaper->addDropPendingNamespace(dropOpTime, dpns);
+    }
     return Status::OK();
 }
 
--- a/src/mongo/db/catalog/database_impl.cpp
+++ b/src/mongo/db/catalog/database_impl.cpp
@@ -600,7 +600,7 @@ Status DatabaseImpl::dropCollectionEvenIfSystem(OperationContext* opCtx,
 
     // Register this drop-pending namespace with DropPendingCollectionReaper to remove when the
     // committed optime reaches the drop optime.
-    repl::DropPendingCollectionReaper *pReaper = repl::DropPendingCollectionReaper::get(opCtx);
+    auto &&pReaper = repl::DropPendingCollectionReaper::get(opCtx);
     if (pReaper != nullptr) {
         pReaper->addDropPendingNamespace(dropOpTime, dpns);
     }
--- a/src/mongo/dbtests/clienttests.cpp
+++ b/src/mongo/dbtests/clienttests.cpp
@@ -28,8 +28,6 @@
  *    then also delete it in the license file.
  */
 
-#include <mongo/db/repl/drop_pending_collection_reaper.h>
-#include <mongo/db/repl/storage_interface_mock.h>
 #include "mongo/platform/basic.h"
 
 #include "mongo/client/dbclientcursor.h"
@@ -49,14 +47,11 @@ using std::vector;
 
 class Base {
 public:
-    Base(string coll) : _ns("test." + coll), _storageInterface(std::make_unique<repl::StorageInterfaceMock>()) {
+    Base(string coll) : _ns("test." + coll) {
         const ServiceContext::UniqueOperationContext opCtxPtr = cc().makeOperationContext();
         OperationContext& opCtx = *opCtxPtr;
         DBDirectClient db(&opCtx);
-        repl::DropPendingCollectionReaper::set(
-                opCtxPtr.get()->getServiceContext(),
-                std::make_unique<repl::DropPendingCollectionReaper>(_storageInterface.get())
-        );
+
         db.dropDatabase("test");
     }
 
@@ -73,8 +68,6 @@ public:
     }
 
     const string _ns;
-
-    std::unique_ptr<repl::StorageInterface> _storageInterface;
 };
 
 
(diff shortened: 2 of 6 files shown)


src/mongo/db/catalog/database_impl.cpp
```diff
@@ -600,8 +600,10 @@ Status DatabaseImpl::dropCollectionEvenIfSystem(OperationContext* opCtx,
 
     // Register this drop-pending namespace with DropPendingCollectionReaper to remove when the
     // committed optime reaches the drop optime.
-    repl::DropPendingCollectionReaper::get(opCtx)->addDropPendingNamespace(dropOpTime, dpns);
-
+    auto &&pReaper = repl::DropPendingCollectionReaper::get(opCtx);
```
benety: A running mongod server must always have a value DropPendingCollectionReaper. Can you see which tests are failing to provide a valid DropPendingCollectionReaper instance?

src/mongo/dbtests/validate_tests.cpp
```diff
@@ -68,8 +68,10 @@ class ValidateBase {
         _client.createCollection(_ns);
         {
             AutoGetCollection autoGetCollection(&_opCtx, _nss, MODE_X);
-            _isInRecordIdOrder =
-                autoGetCollection.getCollection()->getRecordStore()->isInRecordIdOrder();
+            auto && col = autoGetCollection.getCollection();
+            if (col != nullptr) {
```
benety: Why is the collection "_nss" not always available? Can you check the DBClient::createCollection() result on line 68?

apocarteres:
@benety thanks for comments. will check it asap

src/mongo/dbtests/clienttests.cpp
```diff
@@ -51,7 +53,11 @@ class Base {
         const ServiceContext::UniqueOperationContext opCtxPtr = cc().makeOperationContext();
         OperationContext& opCtx = *opCtxPtr;
         DBDirectClient db(&opCtx);
-
+        auto storagePtr = std::make_unique<repl::StorageInterfaceMock>();
```
apocarteres: @benety please review this. As you mentioned NPE should not occur at all and test has to provide instance of ```DropPendingCollectionReaper``` by itself

benety: @apocarteres what's the lifetime of "storagePtr"? It looks like its scope is limited to this setup function. You might need to set the singleton for repl::StorageInterface. Look for usages of repl::StorageInterface::set().

benety:
https://jira.mongodb.org/browse/SERVER-32344

src/mongo/dbtests/validate_tests.cpp
benety: @apocarteres we can address this validate_tests.cpp change in another pull request if you don't have the time to investigate why we need this conditional.

src/mongo/dbtests/clienttests.cpp
apocarteres: sure. agree i'm going to update code asap

src/mongo/dbtests/validate_tests.cpp
apocarteres: sounds reasonable, i can exclude it from this PR for now

src/mongo/dbtests/clienttests.cpp
```diff
@@ -28,6 +28,8 @@
  *    then also delete it in the license file.
```
apocarteres: @benety please review this. I dropped extra changes and made storage mock a singleton

src/mongo/dbtests/dbtests.cpp
```diff
@@ -156,6 +158,11 @@ int dbtestsMain(int argc, char** argv, char** envp) {
         ->setFollowerMode(repl::MemberState::RS_PRIMARY)
         .ignore();
 
+    repl::DropPendingCollectionReaper::set(
+                service,
+                std::make_unique<repl::DropPendingCollectionReaper>(new repl::StorageInterfaceMock())
+    );
+
```
apocarteres: finally i moved storage mock to more suitable place (as far as i understand). @benety please review

src/mongo/dbtests/dbtests.cpp
```diff
@@ -45,6 +45,8 @@
 #include "mongo/db/logical_clock.h"
 #include "mongo/db/repl/replication_coordinator_global.h"
 #include "mongo/db/repl/replication_coordinator_mock.h"
+#include <mongo/db/repl/drop_pending_collection_reaper.h>
```
benety: Please use double quotes for #includes:

https://github.com/mongodb/mongo/wiki/Server-Code-Style#includes

src/mongo/dbtests/dbtests.cpp
```diff
@@ -156,6 +158,11 @@ int dbtestsMain(int argc, char** argv, char** envp) {
         ->setFollowerMode(repl::MemberState::RS_PRIMARY)
         .ignore();
 
+    repl::DropPendingCollectionReaper::set(
+                service,
+                std::make_unique<repl::DropPendingCollectionReaper>(new repl::StorageInterfaceMock())
```
benety: Please use stdx::make_unique() instead of std::make_unique(). FYI, there's a built-in linter you can run using the build system:

scons.py .... lint

apocarteres:
@benety fixed codestyle

src/mongo/dbtests/dbtests.cpp
```diff
@@ -156,6 +158,10 @@ int dbtestsMain(int argc, char** argv, char** envp) {
         ->setFollowerMode(repl::MemberState::RS_PRIMARY)
         .ignore();
 
+    repl::DropPendingCollectionReaper::set(
```
benety: This will leak an instance of StorageInterfaceMock. You have to set the singleton using StorageInterface::set(). Here's an example (change StorageInterfaceImpl to StorageInterfaceMock):

https://github.com/mongodb/mongo/blob/523531464b747daa2a25eea33ed42239db871622/src/mongo/dbtests/storage_timestamp_tests.cpp#L91

apocarteres: yes, i know, i just noticed that this is a ```main``` function, plus code above in this function doesn't care about similar leakage. Ok let me fix that. Thanks

src/mongo/dbtests/dbtests.cpp
```diff
@@ -156,6 +158,10 @@ int dbtestsMain(int argc, char** argv, char** envp) {
         ->setFollowerMode(repl::MemberState::RS_PRIMARY)
         .ignore();
 
+    auto storageMock = stdx::make_unique<repl::StorageInterfaceMock>();
```
apocarteres: @benety please review

src/mongo/dbtests/dbtests.cpp
benety: Are you referring to the call to "new repl::ReplicationCoordinatorMock(service, replSettings)"? repl::setGlobalReplicationCoordinator() creates a std::unique_ptr to manage the memory of the ReplicationCoordinator instance passed in.

apocarteres: @banker right, but wan't it leak in some specific case when ReplicationCoordinator is instantiated and unique pointer was unable to be constructed for some reason? Strictly speaking instance of ReplicationCoordinator spawns before unique pointer gets a control. I'm not a big C++ expert yet, but i'm trying to follow up :D

benety:
@apocarteres, your change has been merged. Thank you for your contribution!

apocarteres:
@benety thanks for reviewing!

PR #1196, opened 2017-12-22
SERVER-30283 Update topology_coordinator's .h and .cpp files

cramaechi:
PingStats, defined in topology_coordinator.cpp, is a data structure used
to track statistics about replication heartbeat messages. It has a
variable called _numFailuresSinceLastStart, which keeps track of failed
heartbeat attempts. When a good heartbeat response comes back, this
variable is set to std::numeric_limits<int>max(), presumably as an easy
way for this value to always be larger than others in certain
comparisons. In the case that we actually get back a good hearbeat
response (PingStats::hit) and then a failed heartbeat response from the
same node (PingStats::miss), we will actually overflow this
_numFailuresSinceLastStart variable. The integer max function should
not be used to determine if a good heartbeat response was received.

To remediate the issue, a field named _goodHeartbeatReceived was
added to PingStats to indicate whether a good heartbeat response
(PingStats::hit) was received since the last start
(PingStats::start). The assignment statement responsible for the
potential overflow is then replaced by a new assignment statement
that sets _goodHeartbeatReceived to true when a good heartbeat
response comes back.

Resolves: SERVER-30283

### DIFF
--- a/src/mongo/db/repl/topology_coordinator.h
+++ b/src/mongo/db/repl/topology_coordinator.h
@@ -52,6 +52,9 @@ class ReplSetConfig;
 class TagSubgroup;
 struct MemberState;
 
+// Maximum number of retries for a failed heartbeat.
+const int kMaxHeartbeatRetries = 2;
+
 /**
  * Replication Topology Coordinator
  *
@@ -1046,7 +1049,7 @@ public:
      * last call to start(). Otherwise, returns false.
      */
     bool hasFailed() const {
-        return (_numFailuresSinceLastStart > _kMaxHeartbeatRetries) ||
+        return (_numFailuresSinceLastStart > kMaxHeartbeatRetries) ||
             (_goodHeartbeatReceived == true);
     }
 
@@ -1054,14 +1057,12 @@ public:
      * Gets the number of retries left for a failed heartbeat.
      */
     int retriesLeft() const {
-        return _kMaxHeartbeatRetries - _numFailuresSinceLastStart;
+        return kMaxHeartbeatRetries - _numFailuresSinceLastStart;
     }
 
 private:
     static constexpr Milliseconds UninitializedPing{-1};
 
-    // Maximum number of retries for a failed heartbeat.
-    int _kMaxHeartbeatRetries = 2;
     unsigned int count = 0;
     Milliseconds value = UninitializedPing;
     Date_t _lastHeartbeatStartDate;
--- a/src/mongo/db/repl/topology_coordinator.h
+++ b/src/mongo/db/repl/topology_coordinator.h
@@ -1041,38 +1041,27 @@ public:
     }
 
     /**
-     * Gets the number of failures since start() was last called.
-     *
-     * This value is incremented by calls to miss() and cleared by calls to start().
-     * 
+     * Returns true if the number of failed heartbeats has exceeded the max number
+     * of heartbeat retries or if a good heartbeat has been received since the
+     * last call to start(). Otherwise, returns false.
      */
-    int getNumFailuresSinceLastStart() const {
-        return _numFailuresSinceLastStart;
+    bool hasFailed() const {
+        return (_numFailuresSinceLastStart > _kMaxHeartbeatRetries) ||
+            (_goodHeartbeatReceived == true);
     }
 
     /**
-     * Gets a flag that indicates whether the number of failed heartbeats has
-     * exceeded the max number of heartbeat retries or whether a good
-     * heartbeat has been received since the last call to start(). 
+     * Gets the number of retries left for a failed heartbeat.
      */
-    bool noMoreRetries() const {
-        return _numFailuresSinceLastStart > 2 || _goodHeartbeatReceived == true;
-    }
-	
-    /**
-     * Gets a flag that indicates whether a good heartbeat was received
-     * since the last call to start().
-     *
-     * This value is set to false on calls to start() and set to true on
-     * call to hit().
-     */
-    bool getGoodHeartbeatReceived() const {
-        return _goodHeartbeatReceived;
+    int retriesLeft() const {
+        return _kMaxHeartbeatRetries - _numFailuresSinceLastStart;
     }
 
 private:
     static constexpr Milliseconds UninitializedPing{-1};
 
+    // Maximum number of retries for a failed heartbeat.
+    int _kMaxHeartbeatRetries = 2;
     unsigned int count = 0;
     Milliseconds value = UninitializedPing;
     Date_t _lastHeartbeatStartDate;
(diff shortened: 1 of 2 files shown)


src/mongo/db/repl/topology_coordinator.h
```diff
@@ -1043,20 +1043,41 @@ class TopologyCoordinator::PingStats {
...
-     * set to the maximum possible value by calls to hit().
+     * This value is incremented by calls to miss() and cleared by calls to start().
+     * 
      */
     int getNumFailuresSinceLastStart() const {
         return _numFailuresSinceLastStart;
     }
 
+    /**
+     * Gets a flag that indicates whether the number of failed heartbeats has
```
will62794: When describing a function like this, it's a bit redundant to say "Gets a flag...". The fact that a function returns a boolean should be enough information about the nature of the return value. In a comment like this, it can be clearer to just concisely describe the case when the function returns `true`, and when the function returns `false`.

src/mongo/db/repl/topology_coordinator.h
```diff
@@ -1043,20 +1043,41 @@ class TopologyCoordinator::PingStats {
...
+     * exceeded the max number of heartbeat retries or whether a good
+     * heartbeat has been received since the last call to start(). 
+     */
+    bool noMoreRetries() const {
+        return _numFailuresSinceLastStart > 2 || _goodHeartbeatReceived == true;
+    }
+	
+    /**
+     * Gets a flag that indicates whether a good heartbeat was received
```
will62794: Same applies here as on line 1054.

cramaechi:
Hi will62794,

All the changes have been applied, as per your instructions.

I also added a new method called `PingStats::retriesLeft()`,  which will replace the arithmetic expression in the following line of code:

`src/mongo/db/repl/topology_coordinator.cpp`
screen shot 2018-01-03 at 4 09 17 pm

The reason for this particular change is that `kMaxHeartbeatRetries` will no longer be declared in src/mongo/db/repl/topology_coordinator.cpp, but instead in the .h file. So instead of incurring the overhead of calling the accessors to `kMaxHeartbeatRetries` and `getNumFailuresSinceLastStart`, I will now just be calling one method to handle the computation.

As for the C++ unit test, are there any comprehensive guides available to aid me in the design and implementation? I have already read the guide provided in the Wiki section of this repo, but are there any others?

- Chibuikem. A

will62794:
Sounds good. Once you add the unit test feel free to push the new changes and I will take another look at them.

The best guide for how to add a unit test is probably to look at an example i.e. `src/mongo/db/repl/topology_coordinator_test.cpp`. Since `PingStats` is a class inside the `TopologyCoordinator` file, you could probably add a small test case in `db/repl/topology_coordinator_test.cpp`.

cramaechi:
Hi will62794,

After building `src/mongo/db/repl/topology_coordinator_test.cpp` with my new unit test, I received a compilation error notifying me that `PingStats` was a private member class of `TopologyCoordinator`,   which made me realize that I was attempting to access a private member of a class.

Typically, unit tests only test the public interface, so are we allowed to test private members of a class in this context? If not, then we can still have the peace of mind that there are multiple tests in `src/mongo/db/repl/topology_coordinator_test.cpp` that are already exercising _all_ the methods of `PingStats` via public methods (e.g., `TopologyCoordinator::prepareHeartbeatRequest()`, `TopologyCoordinator::prepareHeartbeatResponse()`, `TopologyCoordinator::processHeartbeatResponse()`, etc..).

I will push my other changes for now and await your decision on the unit test part.

For documentation purposes, here's what my (unfinished) unit test would look like:

```
TEST_F(TopoCoordTest, UpdatePingStatsWhenHeartbeatRequestHasBeenProcessed) {
    // This test confirms that topology coordinator correctly updates heartbeat
    // stats for each heartbeat request processed.

    Date_t firstRequestDate = unittest::assertGet(dateFromISOString("2018-01-04T13:00Z"));
    TopologyCoordinator::PingStats hbStats;

    // Initial heartbeat request started at firstRequestDate.
    hbStats.start(firstRequestDate);
    ASSERT_EQUALS(hbStats.getLastHeartbeatStartDate(), firstRequestDate);
    ASSERT_EQUALS(hbStats.retriesLeft(), 2);
    ASSERT_EQUALS(hbStats.getCount(), 0);
    ASSERT_FALSE(hbStats.hasFailed());
    
    hbStats.miss(); 
    ASSERT_EQUALS(hbStats.retriesLeft(), 1);
    ASSERT_EQUALS(hbStats.getCount(), 0);
    ASSERT_FALSE(hbStats.hasFailed()); 

    hbStats.hit(Milliseconds(100))
    ASSERT_EQUALS(hbStats.retriesLeft(), 1);
    ASSERT_EQUALS(hbStats.getCount(), 1);
    ASSERT_TRUE(hbStats.hasFailed());

    hbStats.miss();
    ASSERT_EQUALS(hbStats.retriesLeft(), 0);
    ASSERT_EQUALS(hbStats.getCount(), 1);
    ASSERT_TRUE(hbStats.hasFailed());
}
```

cramaechi:
My new changes have now been pushed.

will62794:
Hi cramaechi,

You are right, I overlooked the fact that PingStats is private to `topology_coordinator.cpp`. As long as we are exercising PingStats via the public interface of `TopologyCoordinator`, that should be fine. Sorry about that confusion.

Also, it looks like your latest changes include many other unrelated changes, which makes it difficult to do another review pass. Did you perhaps rebase your branch against master and then push? That may be causing the noise. If you could try to clean that up and push your changes again, then I will take another look.

Thanks!

cramaechi:
Hi will62794 - I have opened a new pull request.

src/mongo/db/repl/topology_coordinator.h
```diff
@@ -1041,22 +1041,32 @@ class TopologyCoordinator::PingStats {
...
-        return _numFailuresSinceLastStart;
+    int retriesLeft() const {
+        return _kMaxHeartbeatRetries - _numFailuresSinceLastStart;
     }
 
 private:
     static constexpr Milliseconds UninitializedPing{-1};
 
+    // Maximum number of retries for a failed heartbeat.
+    int _kMaxHeartbeatRetries = 2;
```
will62794: Keep this defined as a constant in this file, instead of a member variable.

cramaechi:
Change has been made.

will62794:
https://jira.mongodb.org/browse/SERVER-30283

will62794:
Hi cramaechi,

This patch looks good, thanks! I have compiled it and run the appropriate tests and everything seems to look fine. I will let you know when I get this merged. Thank you for your work!

Will

cramaechi:
Hi will62794,

No problem. I look forward to contributing more!

Best Regards,
Chibuikem Amaechi

will62794:
Your changes have been merged at 6e3b0d!

cramaechi:
Great. Thank you!

PR #1197, opened 2017-12-26
SERVER-4785 maintain slowms at database level

zhihuiFan:
See issue at https://jira.mongodb.org/browse/SERVER-4785 and https://jira.mongodb.org/browse/SERVER-18946 

Here is the test result.

zhifan@zhifan-dev16:~/github/mongo$ ./mongo
> db.getProfilingStatus()
{ "was" : 0, "slowms" : 100, "sampleRate" : 1 }
> db.setProfilingLevel(0, 1)
{ "was" : 0, "slowms" : 100, "sampleRate" : 1, "ok" : 1 }
> db.getProfilingStatus()
{ "was" : 0, "slowms" : 1, "sampleRate" : 1 }
> for(var i =0; i<10000; i++) {db.test.save({'i': i})}
> use test2
switched to db test2
> db.getProfilingStatus()
{ "was" : 0, "slowms" : 100, "sampleRate" : 1 } . -- test 2: still 100
> for(var i =0; i<10000; i++) {db.test.save({'i': i})}

as a result:  we capture the slow queries from test database but not for test2 database.

`
2017-12-26T13:38:32.678+0800 I COMMAND  [conn1] command test.test appName: "MongoDB Shell" command: insert { insert: "test", ordered: true, $db: "test" } ninserted:1 keysInserted:1 numYields:0 reslen:29 locks:{ Global: { acquireCount: { r: 3, w: 1 } }, Database: { acquireCount: { w: 1, R: 1 } }, Collection: { acquireCount: { w: 1 } } } protocol:op_msg 3ms

`

### DIFF
--- a/jstests/core/hidden_index.js
+++ b/jstests/core/hidden_index.js
@@ -1,0 +1,45 @@
+var rsTest = new ReplSetTest({ nodes: 2 });
+rsTest.startSet({ oplogSize: 1000  });
+rsTest.initiate();
+rsTest.awaitSecondaryNodes();
+
+
+var primary = rsTest.getPrimary();
+var secondary = rsTest.getSecondary();
+secondary.setSlaveOk();
+
+var testtab = primary.getDB("test").hidden_index_tab;
+var testtabsec = secondary.getDB("test").hidden_index_tab;
+
+testtab.createIndex({a: 1});
+testtab.createIndex({b: 1}, {hidden: true});
+
+var res = testtab.find({a: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.inputStage.stage, "IXSCAN");
+res = testtab.find({b: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.stage, "COLLSCAN");
+sleep(1000);
+
+
+res = testtabsec.find({a: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.inputStage.stage, "IXSCAN");
+res = testtabsec.find({b: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.stage, "COLLSCAN");
+
+
+testtab.hiddenIndex("a_1");
+testtab.unhiddenIndex({b: 1});
+
+res = testtab.find({a: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.stage, "COLLSCAN");
+res = testtab.find({b: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.inputStage.stage, "IXSCAN");
+sleep(1000);
+
+res = testtabsec.find({a: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.stage, "COLLSCAN");
+res = testtabsec.find({b: 1}).explain()
+assert.eq(res.queryPlanner.winningPlan.inputStage.stage, "IXSCAN");
+
+
+rsTest.stopSet();
--- a/src/mongo/db/auth/auth_op_observer.cpp
+++ b/src/mongo/db/auth/auth_op_observer.cpp
@@ -114,11 +114,11 @@ void AuthOpObserver::onCollMod(OperationContext* opCtx,
                                OptionalCollectionUUID uuid,
                                const BSONObj& collModCmd,
                                const CollectionOptions& oldCollOptions,
-                               boost::optional<TTLCollModInfo> ttlInfo) {
+                               boost::optional<IndexCollModInfo> indexInfo) {
     const auto cmdNss = nss.getCommandNS();
 
     // Create the 'o' field object.
-    const auto cmdObj = makeCollModCmdObj(collModCmd, oldCollOptions, ttlInfo);
+    const auto cmdObj = makeCollModCmdObj(collModCmd, oldCollOptions, indexInfo);
 
     AuthorizationManager::get(opCtx->getServiceContext())
         ->logOp(opCtx, "c", cmdNss, cmdObj, nullptr);
--- a/src/mongo/db/auth/auth_op_observer.h
+++ b/src/mongo/db/auth/auth_op_observer.h
@@ -113,7 +113,7 @@ public:
                    OptionalCollectionUUID uuid,
                    const BSONObj& collModCmd,
                    const CollectionOptions& oldCollOptions,
-                   boost::optional<TTLCollModInfo> ttlInfo) final;
+                   boost::optional<IndexCollModInfo> indexInfo) final;
 
     void onDropDatabase(OperationContext* opCtx, const std::string& dbName) final;
 
--- a/src/mongo/db/catalog/coll_mod.cpp
+++ b/src/mongo/db/catalog/coll_mod.cpp
@@ -69,6 +69,7 @@ MONGO_FAIL_POINT_DEFINE(assertAfterIndexUpdate);
 struct CollModRequest {
     const IndexDescriptor* idx = nullptr;
     BSONElement indexExpireAfterSeconds = {};
+    BSONElement indexHidden = {};
     BSONElement viewPipeLine = {};
     std::string viewOn = {};
     BSONElement collValidator = {};
@@ -124,14 +125,19 @@ StatusWith<CollModRequest> parseCollModRequest(OperationContext* opCtx,
             }
 
             cmr.indexExpireAfterSeconds = indexObj["expireAfterSeconds"];
-            if (cmr.indexExpireAfterSeconds.eoo()) {
-                return Status(ErrorCodes::InvalidOptions, "no expireAfterSeconds field");
+            cmr.indexHidden = indexObj["hidden"];
+
+            if (cmr.indexExpireAfterSeconds.eoo() && cmr.indexHidden.eoo()) {
+                return Status(ErrorCodes::InvalidOptions, "no expireAfterSeconds or hidden field");
             }
-            if (!cmr.indexExpireAfterSeconds.isNumber()) {
+            if (!cmr.indexExpireAfterSeconds.eoo() && !cmr.indexExpireAfterSeconds.isNumber()) {
                 return Status(ErrorCodes::InvalidOptions,
                               "expireAfterSeconds field must be a number");
             }
-
+            if (!cmr.indexHidden.eoo() && !cmr.indexHidden.isBoolean()) {
+                return Status(ErrorCodes::InvalidOptions,
+                              "hidden field must be a boolean");
+            }
             if (!indexName.empty()) {
                 cmr.idx = coll->getIndexCatalog()->findIndexByName(opCtx, indexName);
                 if (!cmr.idx) {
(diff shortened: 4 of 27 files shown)


zhihuiFan:
pls ignore this pull request for now, looks there are more things need to thinking.  fixing the test case..

zhihuiFan:
now all the test passed except the following 3:
1. jstests/core/profile_sampling.js   the error is  
[js_test:profile_sampling.js] 2017-12-27T00:25:12.857+0800 2017-12-27T00:25:12.857+0800 E QUERY    [thread1] Error: command failed: {
[js_test:profile_sampling.js] 2017-12-27T00:25:12.857+0800 	"ok" : 0,
[js_test:profile_sampling.js] 2017-12-27T00:25:12.857+0800 	"errmsg" : "slowms has to be >=0",
[js_test:profile_sampling.js] 2017-12-27T00:25:12.857+0800 	"code" : 2,
[js_test:profile_sampling.js] 2017-12-27T00:25:12.857+0800 	"codeName" : "BadValue"
[js_test:profile_sampling.js] 2017-12-27T00:25:12.858+0800 } : undefined :
, but this looks a test case bug to me since slowms should be >= 0 in fact;

2. jstests/core/opcounters_write_cmd.js, the error is 
[js_test:opcounters_write_cmd] 2017-12-27T00:11:54.209+0800 2017-12-27T00:11:54.208+0800 E QUERY    [thread1] Error: [123044] != [123049] are not equal : undefined :
 
however,  if I run it again, it succeed. (when I get the error, the command is `python buildscripts/resmoke.py -j 4 --shellPort 27017 jstests/core/*.js`)

3. jstests/core/mr3.js
when I get the error, the command is `python buildscripts/resmoke.py -j 4 --shellPort 27017 jstests/core/*.js`
if I run into again, it succeeds.

could you help review the pull requests and the 3 issue above?

Thanks

zhihuiFan:
I will add a new test case under jstests/core/ for this feature soon(when I get a time).

zhihuiFan:
test case is added.

hi @amschwerin ,  could you or let someone review this request?   It would be great that this feature can be included in the next stable version, like 3.6.2 and our product will use 3.6.2 directly?

Thanks

amschwerin:
The SERVER ticket link in the summary is to the wrong ticket. Can you update it, so I can get some context on this change? Thanks!

hsasnmahamd:
[](url)

zhihuiFan:
hi @amschwerin:
   What will be the process after the pull request is approved?  

Thanks

amschwerin:
Once someone from the MongoDB approved committers has approved the request, he or she will merge it into the "master" branch. If it is deemed safe to back port to 3.6, we'll handle that process as well.

zhihuiFan:
Thank you @amschwerin  for the quick response all the time!

src/mongo/db/commands/dbcommands.cpp
```diff
@@ -371,7 +372,7 @@ class CmdProfile : public ErrmsgCommandDeprecated {
 
         const BSONElement slow = cmdObj["slowms"];
         if (slow.isNumber()) {
-            serverGlobalParams.slowMS = slow.numberInt();
+            uassertStatusOK(db->setSlowMS(slow.numberInt()));
```
amschwerin: How does an operator change the _global_ slowMS? We closed [SERVER-5266](https://jira.mongodb.org/browse/SERVER-5266) as "Works as designed", but maybe that could be revived as an approach.

We should also decide if it's OK to change the behavior of this command for existing users, or if we'll need to add new syntax or a new command for setting per-db slowMS values. @asya999 and I will try to spend some time on this question and get back to you.

zhihuiFan: the global slowMS will not be changed at more,  and the value of it is the "default" slowMS for all the database.  I know this request changed the default behavior,  I think you are already thinking if the behavior should be changed or not. 

please let me know your decision if you have made it since I have another feature called "invisible" index want to send...

zhihuiFan:
hi @amschwerin, any chance to move this quicker?

src/mongo/db/catalog/catalog_raii.cpp
```diff
@@ -29,6 +29,7 @@
 #include "mongo/platform/basic.h"
 
 #include "mongo/db/catalog/catalog_raii.h"
+#include "mongo/db/curop.h"
```
amschwerin: I don't know why the blank line below this was present before, but it was probably a mistake. If you delete the line below this one and sort the block of include directives, it will be more consistent with other places in our codebase.

zhihuiFan:
hi @amschwerin 
Thanks for the suggestion,  I have changed the code based on that.  here are some highlight:
1.  slowms change will be applied to the global parameter, which is the default value slowms value for all the database.
2. slowmsDb is only applied to a given database,  it has a high priority that global slowms.
3. if database A is changed with slowmsDb option,  then dba change the global slowms,  it following change will have no impacts on database A since it has been set with slowmsDb on purpose.
4.  I changed the test case `profile_sampling.js` which is there before my pull request.  the reason is it allow slowms < 0,  it doesn't make sense to me.  please correct me if I am wrong here. 
5.  I have run all the test case `python buildscripts/resmoke.py jstests/core/*.js --shellPort=27017`, all pass.
6. I have run `python buildscripts/clang_format.py format` for code format
7.  all the other format issues are handled manually.

src/mongo/db/commands/dbcommands.cpp
```diff
@@ -372,6 +373,13 @@ class CmdProfile : public ErrmsgCommandDeprecated {
         const BSONElement slow = cmdObj["slowms"];
         if (slow.isNumber()) {
             serverGlobalParams.slowMS = slow.numberInt();
+            uassertStatusOK(db->setSlowMS(slow.numberInt()));
```
amschwerin: I think this line should be removed. That way, the behavior of setting "slowms" is completely unchanged, and only setting "slowmsDb" changes the database-specific override.

src/mongo/db/service_entry_point_mongod.cpp
```diff
@@ -25,13 +25,15 @@
...
 
 #define MONGO_LOG_DEFAULT_COMPONENT ::mongo::logger::LogComponent::kCommand
 
-#include "mongo/platform/basic.h"
-
+#include "mongo/db/catalog/database_holder.h"
+#include "mongo/db/db_raii.h"
 #include "mongo/db/service_entry_point_mongod.h"
 
+
```
amschwerin: Since I'm asking you to update the patch one last time, anyhow, please remove this introduced blank line.

src/mongo/db/service_entry_point_mongod.cpp
```diff
@@ -1132,8 +1134,8 @@ DbResponse ServiceEntryPointMongod::handleRequest(OperationContext* opCtx, const
     }
 
     OpDebug& debug = currentOp.debug();
+    long long logThresholdMs;
```
amschwerin: This variable is now unused, I believe. Go ahead and remove it.

zhihuiFan:
hi @amschwerin  I have changed based on your suggestion and added ';' into db_level_slowms.js file and run the format.  

please let me know if any more changes needed.

thanks for your help on this case.

jstests/core/db_level_slowms.js
```diff
@@ -0,0 +1,24 @@
+(function() {
+    "use strict";
+
+    let SlowDB = db.getSiblingDB('slowdb20171229');
+    assert.commandWorked(SlowDB.dropDatabase());
+    SlowDB.setProfilingLevel(1, {slowmsDb: 0});
+    assert(SlowDB.system.profile.count() < 2, "db.system.profile is not empty as the beginging");
```
asya999: What is this error message supposed to be?

zhihuiFan: hi @asya999 :  If we set the `slowmsDb` to 0, suppose every query should be logged.  so I count the number of slow query at the beginning s1, run the  10 test queries, count the number of slow query again s2.

then s2 >= 10.  

I think all of us has spent on this small feature too long time,  do you have any suggestion to make it faster?

asya999:
I added a note to the related server ticket.  There are some questions in my mind about how the interaction between different global and dB-level slowms should be handled. 

In addition we are designing how logging slow operations should happen on mongos and we want to make sure there is no conflict between this work and that design.

zhihuiFan:
Hi @asya999 :
  Thanks for your quick response. I'd like to keep the discussion in one place if you don't mind.  Here is the summary of your concerns and the answer from my side. 
  
  1. we should consider non-backwards breaking syntax since support which is currently assumed to have global effect.
  >> I followed the suggestion from Andy by adding a slowmsDb option to set the db level. so it will not broke the "slowms" option.
  
  2. Does the "db-only" flag apply to slowms and sampleRate?
  >> With the "slowmsDb" option, the sampleRate is not impacted by this change. 
  
  3. What if global setting is later changed, does the global setting override earlier set DB only setting?
  >> this is a problem I realized also.  Per current change,  the db level setting has a higher proirity than system level.  so suppose at the beginging, the global setting is 100ms,  then we can set the DB X slowmsDb to 50ms.  then user set the global ms to 10ms. then the DB x still have the 50ms.  I admit that the changes broke the original way of "slowms" in this sistuation. but I think it is reasonable personally. But if you insist on the original behavior and you have a discussion internally, I can accept your final decission. 


  4. Since slowms applies to profiling as well as logging, should per-db change also apply to both or to profiling only?
  >>> yes, this change still follow this way.
  
  5. we are designing how logging slow operations should happen on mongos and we want to make sure there is no conflict between this work and that design.
  >>> I'm not familar with that part, but they should be not irrelevant.  Do please let me know if I need to pay more attention to this part.

zhihuiFan:
hi @asya999 :    Any update?

amschwerin:
Sorry for the long delay; I've been slammed, here. Your current update looks pretty good, so I'll put it through our internal testing and get it merged. It might still take awhile, but unless something explodes, I think we're all set.

zhihuiFan:
Thanks @amschwerin !

amschwerin:
Automated testing revealed a snag that I didn't notice in review. In particular, the database-specific value of slowms is always used, even if the user never sets slowmsDb. What this means is that if the user tries to use the profile command to change slowms (the global value) it will not cause any already-existing databases to see the new value.

If you're feeling patient, I'll spend some time to fix up this bug and get the patch merged over the next month or so. If you want to do it yourself, I'd welcome a proposal. If you do submit a proposal, I suggest also merging up to the current master.

zhihuiFan:
Thanks @amschwerin .   The fix I can think out now is to add a `bool _dbLevelSlowms` into `Database::Impl` and then make the `Database::Impl::getSlowMS()` always check this flag before return the value.  if (!_dbLevelSlowms),   it should return the current global slowms.  do you think it will be ok? or you have other suggestion?

I'd like to submit a clear PR for this case, so I prefer to fix this issue.

zhihuiFan:
hi @amschwerin   Any update on this?

dstorch:
Since this change was proposed, we have added an implementation for the `profile` command on mongos. This was done in support of logging slow queries on mongos under [SERVER-14900](https://jira.mongodb.org/browse/SERVER-14900) (see commit [40d59f4bc](https://github.com/mongodb/mongo/commit/40d59f4bc8eec0fc585edec0f2c8833f09a4a853)).

Mongos will not support per-database `slowms` configuration unless further work is done as part of this patch. In fact, I would recommend _not_ pursuing per-database `slowms` configuration on mongos, instead leaving this as future work. This means, however, that this patch will have to update the mongos `profile` command to fail with an appropriate message before it can be merged.

zhihuiFan:
hi @dstorch :  now if user want to change slowmsDb on a database level, it will get error see [here](https://github.com/mongodb/mongo/pull/1197/files#diff-4e7282df69b7c562da6767be9a1b439aR61)

please let me know if any other changes I need.

zhihuiFan:
closed.

PR #1202, opened 2018-01-17
SERVER-32576: Merge ARM64 support to use 64-bit optimized code path

JunHe77:
Change-Id: Icd2e0f9378952bbd811b54fe60a7f3895c960e92
Signed-off-by: Jun He <jun.he@arm.com>

### DIFF
--- a/src/third_party/snappy-1.1.3/snappy-internal.h
+++ b/src/third_party/snappy-1.1.3/snappy-internal.h
@@ -79,9 +79,8 @@ char* CompressFragment(const char* input,
 // Does not read *(s1 + (s2_limit - s2)) or beyond.
 // Requires that s2_limit >= s2.
 //
-// Separate implementation for x86_64, for speed.  Uses the fact that
-// x86_64 is little endian.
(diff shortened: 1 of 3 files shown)


acmorrow:
Is this just a backport of changes that have landed in the newer version of snappy upstream? If so, would we be better off just upgrading our vendored snappy to 1.1.7?

JunHe77:
Hi, @acmorrow , yes, this is just a backport. If upgrade snappy to 1.1.7 this PR would not be necessary.
I found there is a request for snappy upgrade(https://jira.mongodb.org/browse/SERVER-28073). Is it going to be done in 3.7?

acmorrow:
SERVER-28073 is currently listed as 3.7 desired, which means we want to do it, but no promises. I'd prefer though that we fix this via that mechanism, whenever it happens, so unless you have a strong objection I propose that we close this PR.

JunHe77:
Sure. It would be best to upgrade vendored snappy to 1.1.7. I've updated SERVER-28073 with related JIRA link.

PR #1203, opened 2018-01-17
[SERVER-32259] Update wiredtiger_util.cpp

cramaechi:
WiredTigerUtil::fetchTypeAndSourceURI() is using
invariant(colgroupResult.isOK()) instead of
invariantOK(colgroupResult.getStatus()). This means that if the
invariant ever fires, we don't know what the bad Status actually was.


IanWhalen:
Resolved by merging c765de210434a6a49fa62c8c0602876fec628bfc.

Thanks @cramaechi !

cramaechi:
Thanks @IanWhalen!

PR #1204, opened 2018-01-20
SERVER-31942 Truncate test lifecycle JIRA issue description

cramaechi:
When the update_test_lifecycle task sees changes in unreliable tests, it
creates a JIRA issue before committing the changes.The description of
the issue contains the tag changes and if there are many of them the
description text size can exceed the allowed limit for the field
(32,767 characters).

When that happens the task fails and the changes are not committed.

Instead we should truncate the description when its size exceeds the
allowed limit.

### DIFF
--- a/buildscripts/tests/test_update_test_lifecycle.py
+++ b/buildscripts/tests/test_update_test_lifecycle.py
@@ -1010,14 +1010,14 @@ class TestJiraIssueCreator(unittest.TestCase):
         self.assertEqual(expected, desc)
 
     def test_truncate_description(self):
-        desc = "a" * (JiraIssueCreator._MAX_DESCRIPTION_SIZE - 1)
-        self.assertTrue(desc == JiraIssueCreator._truncate_description(desc))
+        desc = "a" * (update_test_lifecycle.JiraIssueCreator._MAX_DESCRIPTION_SIZE - 1)
+        self.assertTrue(desc == update_test_lifecycle.JiraIssueCreator._truncate_description(desc))
        
         desc += "a"
-        self.assertTrue(desc == JiraIssueCreator._truncate_description(desc))
+        self.assertTrue(desc == update_test_lifecycle.JiraIssueCreator._truncate_description(desc))
 
         desc += "a"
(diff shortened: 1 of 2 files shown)


buildscripts/tests/test_update_test_lifecycle.py
```diff
@@ -1009,6 +1009,15 @@ def test_clean_up_description_empty(self):
         expected = "_None_"
         self.assertEqual(expected, desc)
 
+    def test_truncate_description(self):
+        desc = "a" * (JiraIssueCreator._MAX_DESCRIPTION_SIZE - 1)
+        self.assertTrue(desc == JiraIssueCreator._truncate_description(desc))
```
syev: `JiraIssueCreator` is not defined here, you'll need to reference the class using `update_test_lifecycle.JiraIssueCreator`.

Same on lines 1017 and 1020.

buildscripts/tests/test_update_test_lifecycle.py
```diff
@@ -1009,6 +1009,15 @@ def test_clean_up_description_empty(self):
...
+       
+        desc += "a"
+        self.assertTrue(desc == JiraIssueCreator._truncate_description(desc))
+
+        desc += "a"
+        self.assertTrue(len(desc) <= len(JiraIssueCreator._truncate_description(desc)))
```
syev: This expression will be `False`.
I believe what you want to check is that `len(update_test_lifecycle.JiraIssueCreator._truncate_description(desc))` is inferior or equal to `update_test_lifecycle.JiraIssueCreator._MAX_DESCRIPTION_SIZE)`.

syev:
Hi Chibuikem,
I have pushed your changes in https://github.com/mongodb/mongo/commit/ad0ab27807a29e025b36a82ede139c975ad65cfb.
Thanks for your contribution!

cramaechi:
Hi Yves,

No problem. Thank you for all your helpful notes!

I look forward to contributing more in the future!

Best Regards,
Chibuikem Amaechi

PR #1206, opened 2018-01-24
SERVER-32148 Update oplog.cpp & sync_tail_test.cpp

cramaechi:
We should make "NameSpaceNotFound" an accetable error for "emptyCapped"
and "convertToCapped".

### DIFF
--- a/src/mongo/db/repl/sync_tail_test.cpp
+++ b/src/mongo/db/repl/sync_tail_test.cpp
@@ -1815,34 +1815,37 @@ TEST_F(SyncTailTest, DropDatabaseSucceedsInRecovering) {
 }
 
 TEST_F(IdempotencyTest, EmptyCappedNamespaceNotFound) {
-    // Create a dummy namespace string.
-    auto ns = NamespaceString("foo.bar");
+    // Create a BSON "emptycapped" command.
+    auto emptyCappedCmd = BSON("emptycapped" << nss.coll());
 
-    AutoGetCollectionForReadCommand autoColl(_opCtx.get(), ns);
+    // Create an "emptycapped" oplog entry.
+    auto emptyCappedOp = makeCommandOplogEntry(nextOpTime(), nss, emptyCappedCmd);
+
+    // Ensure that the returned status code signals a failure.
+    ASSERT_NOT_OK(runOpInitialSync(emptyCappedOp));
+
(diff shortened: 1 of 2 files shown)


cramaechi:
Hi @visualzhou - I have updated the two test cases, but the `convertToCapped` test case is still failing on line 1842. Are both operations (`emptyCappedOp` and `convertToCappedOp`) supposed to run successfully given the namespace we are using? Am I constructing `emptyCappedCmd` and `convertToCappedCmd` correctly? 

I have tried answering these questions on my own, but there doesn't seem to be a lot of code examples that are similar to the test cases I am writing.

src/mongo/db/repl/sync_tail_test.cpp
```diff
@@ -1814,6 +1814,40 @@ TEST_F(SyncTailTest, DropDatabaseSucceedsInRecovering) {
...
 
+TEST_F(IdempotencyTest, EmptyCappedNamespaceNotFound) {
+    // Create a BSON "emptycapped" command.
+    auto emptyCappedCmd = BSON("emptycapped" << nss.coll());
+
+    // Create an "emptycapped" oplog entry.
+    auto emptyCappedOp = makeCommandOplogEntry(nextOpTime(), nss, emptyCappedCmd);
+
+    // Ensure that the returned status code signals a failure.
+    ASSERT_NOT_OK(runOpInitialSync(emptyCappedOp));
```
visualzhou: Both of these should be ASSERT_OK.

src/mongo/db/repl/sync_tail_test.cpp
```diff
@@ -1814,6 +1814,40 @@ TEST_F(SyncTailTest, DropDatabaseSucceedsInRecovering) {
...
+
+TEST_F(IdempotencyTest, ConvertToCappedNamespaceNotFound) {
+    // Create a BSON "convertToCapped" command.
+    auto convertToCappedCmd = BSON("convertToCapped" << nss.coll());
+
+    // Create a "convertToCapped" oplog entry.
+    auto convertToCappedOp = makeCommandOplogEntry(nextOpTime(), nss, convertToCappedCmd);
+
+    // Ensure that the returned status code signals a failure.
+    ASSERT_NOT_OK(runOpInitialSync(convertToCappedOp));
```
visualzhou: Ditto.

PR #1209, opened 2018-01-31
V3.4

jinxufei:
image

as the screenshots,i don't know why the .wt file is so big .But i had't do some special operator.


ramonfm:
Looks like this pull request may have been an error, so I'm closing it.

PR #1211, opened 2018-02-07
SERVER-32873 Fixed Incorrect handling of NumberLong in $pow function

KevinCybura:
When the user uses the $pow expression and the base is -1 and the exponent is negative as well the expression returns 1 when it should potentially return -1 for example if base = -1 and exponent = -5  the result should be -1 not 1.
       This was happening because the for loop that finds the power has the condition  for(int i= 0; i < expLong; i++) since the expLong is already < 0 this loop doesnt run and by default the result is 1 so 1 is returned.

### DIFF
--- a/src/mongo/db/pipeline/expression.cpp
+++ b/src/mongo/db/pipeline/expression.cpp
@@ -3146,8 +3146,9 @@ Value ExpressionPow::evaluate(const Document& root) const {
     long long result = 1;
     // Use repeated multiplication, since pow() casts args to doubles which could result in loss of
     // precision if arguments are very large.
+    
     for (int i = 0; i < expLong; i++) {
-        result *= baseLong;
+            result *= baseLong;
     }
 
     if (baseType == NumberLong || expType == NumberLong) {
--- a/src/mongo/db/pipeline/expression.cpp
+++ b/src/mongo/db/pipeline/expression.cpp
@@ -74,8 +74,7 @@ static Value serializeConstant(Value val) {
 string Expression::removeFieldPrefix(const string& prefixedField) {
     uassert(16419,
             str::stream() << "field path must not contain embedded null characters"
-                          << prefixedField.find("\0")
-                          << ",",
+                          << prefixedField.find("\0") << ",",
             prefixedField.find('\0') == string::npos);
 
     const char* pPrefixedField = prefixedField.c_str();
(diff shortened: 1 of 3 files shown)


src/mongo/db/pipeline/expression.cpp
```diff
@@ -3144,8 +3144,18 @@ Value ExpressionPow::evaluate(const Document& root) const {
     }
 
     long long result = 1;
+
+    // When baseLong == -1 and expLong is < 0 the following for loop will never run because expLong
+    // will always be less than 0 so result will always be 1 but the result can potentialy be -1
+    // ex baselong = -1 expLong = -5  then result should be -1
+
```
cswanson310: Remove extra newline, and on 3158. Generally we prefer to keep the comments directly above the line(s) of code they're referencing.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -3144,8 +3144,18 @@ Value ExpressionPow::evaluate(const Document& root) const {
     }
 
     long long result = 1;
+
+    // When baseLong == -1 and expLong is < 0 the following for loop will never run because expLong
```
cswanson310: We tend to put variable names in single quotes within comments, to make it very obvious you're referring to the variable. Here it's not confusing, but sometimes variables have generic names like 'input' or 'it'.

So, here the comment should look like:
`// When 'baseLong' is -1 and 'expLong' is negative, the following for loop will never run because 'expLong' ...`

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -27,7 +27,6 @@
  */
 
 #include "mongo/platform/basic.h"
-
```
cswanson310: This newline is actually intentional, believe it or not. See our style guide here:
https://github.com/mongodb/mongo/wiki/Server-Code-Style#for-cpp-files

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2179,6 +2178,10 @@ TEST(ExpressionFromAccumulators, StdDevSamp) {
          {{}, Value(BSONNULL)}});
 }
 
+TEST(ExpressionFromAccumulators, Pow) {
+    assertExpectedResults("$pow", {{{Value(-1LL), Value(-5LL)}, Value(-1LL)}});
```
cswanson310: Please add at least one case for a negative even exponent. In fact, it should be easy to add some for everything up to 5, just to show the pattern:
```
assertExpectedResults("$pow", {{{Value(-1LL), Value(0LL)}, Value(1LL)}});
assertExpectedResults("$pow", {{{Value(-1LL), Value(-1LL)}, Value(-1LL)}});
assertExpectedResults("$pow", {{{Value(-1LL), Value(-2LL)}, Value(1LL)}});
assertExpectedResults("$pow", {{{Value(-1LL), Value(-3LL)}, Value(-1LL)}});
assertExpectedResults("$pow", {{{Value(-1LL), Value(-4LL)}, Value(1LL)}});
assertExpectedResults("$pow", {{{Value(-1LL), Value(-5LL)}, Value(-1LL)}});
```

It also would be good to include some cases with regular integers (as opposed to `long long`s):
`assertExpectedResults("$pow", {{{Value(-1), Value(-5)}, Value(-1)}});`
`assertExpectedResults("$pow", {{{Value(-1), Value(-4)}, Value(1)}});`

Maybe put those above the ones with the `LL`s, since they're slightly easier to read.

KevinCybura:
I just pushed the changes I created the new test suite and changed the test
so that it just doing what the name says and add back in that newline

PR #1212, opened 2018-02-13
SERVER-33154 upconvert {query: "foo"} correctly

dtbartle:
I believe the `jsCore_compatibility` evergreen task should exercise the buggy legacy path.

### DIFF
--- a/jstests/core/profile_find.js
+++ b/jstests/core/profile_find.js
@@ -171,4 +171,11 @@
     profileObj = getLatestProfilerEntry(testDB, profileEntryFilter);
     assert.eq((typeof profileObj.command.$truncated), "string", tojson(profileObj));
     assert.eq(profileObj.command.comment, "profile_find", tojson(profileObj));
+
+    //
+    // Confirm {query: "foo"} appears correctly in the profiler (SERVER-33154)
+    //
+    coll.find({query: "foo"}).itcount();
+    profileObj = getLatestProfilerEntry(testDB, profileEntryFilter);
+    assert.eq(profileObj.command.filter, {query: "foo"}, tojson(profileObj));
 })();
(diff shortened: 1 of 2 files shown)


jstests/core/profile_find.js
```diff
@@ -171,4 +171,11 @@
     profileObj = getLatestProfilerEntry(testDB, profileEntryFilter);
     assert.eq((typeof profileObj.command.$truncated), "string", tojson(profileObj));
     assert.eq(profileObj.command.comment, "profile_find", tojson(profileObj));
+
+    //
+    // Confirm {query: "foo"} appears correctly in the profiler (SERVER-33154)
```
gormanb: This comment should be more explicit about what the test is doing, and the `{query: "foo"}` filter is misleading since the query will only appear in that format when the test is run under `jsCore_compatibility`. We also (generally) don't refer to specific SERVER tickets in test comments; if it's necessary to find the commit where a particular test was added, we can easily determine that without including an in-code reference here.

Could you change this to something like the following: "Confirm that a query whose filter contains a field named 'query' appears as expected in the profiler. This test ensures that upconverting a legacy query correctly identifies this as a user field rather than a wrapped filter spec."

jstests/core/profile_find.js
```diff
@@ -171,4 +171,11 @@
...
+    //
+    coll.find({query: "foo"}).itcount();
```
gormanb: In addition to testing the profiler output here, could you also add the same coverage to the other places where upconversion is relevant, i.e. `currentOp` and the output logs? Creating some new test cases in the following files should cover everything:

- `jstests/noPassthrough/log_find_getmore.js`
- `jstests/noPassthrough/currentop_query.js`

Note that both of the above are *not* run in a `_compatibility` suite; they run each test for `command` and `legacy` readModes internally.

gormanb:
Hi @dtbartle - given that there was some additional CurOp work queuing up behind this, I've pushed your fix to master along with a followup patch to add the tests and clarifications mentioned in my review comments. You can view the commit here:

https://github.com/mongodb/mongo/commit/e4f1475e9cdf8da3958c14a92b9362e1ce3520af

Thanks again for your contribution!

PR #1213, opened 2018-02-14
SERVER-27100 ExpressionObject does not optimize itself to a constant expression

KevinCybura:
Added optimization to ExpressionObject if it contains all constant values optimize to ExpressionConstant.


src/mongo/db/pipeline/expression.cpp
```diff
@@ -1729,9 +1729,18 @@ intrusive_ptr<ExpressionObject> ExpressionObject::parse(
 }
 
 intrusive_ptr<Expression> ExpressionObject::optimize() {
+    size_t countConstants = 0;
```
cswanson310: You don't really need to count them all, do you? A simple boolean should suffice:
`bool allValuesConstant = true`

src/mongo/db/pipeline/expression.cpp
```diff
@@ -1729,9 +1729,18 @@ intrusive_ptr<ExpressionObject> ExpressionObject::parse(
...
     for (auto&& pair : _expressions) {
+        // Check if Expression Constant
```
cswanson310: Use full sentences in comments. This one seems like it's not adding much that the next line doesn't already say, so I'd actually suggest removing it.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -1729,9 +1729,18 @@ intrusive_ptr<ExpressionObject> ExpressionObject::parse(
...
+        if (dynamic_cast<ExpressionConstant*>(pair.second.get())) {
+            countConstants++;
+        }
         pair.second = pair.second->optimize();
     }
+    // If all values in ExpressionObject are constant evaluate to ExpressionConstant
```
cswanson310: End comments in a period. I also found this slightly hard to read - how about 
`// If all values in the object are constant, we can replace this ExpressionObject with a constant value.`

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2864,6 +2864,24 @@ TEST(ExpressionObjectOptimizations, OptimizingAnObjectShouldOptimizeSubExpressio
     ASSERT_VALUE_EQ(expConstant->evaluate(Document()), Value(3));
 };
 
+TEST(ExpressionObjectOptimizations, OptimizingAnObjectWithAllConstantsShouldOptimizeToExpressionConstant) {
+
+    intrusive_ptr<ExpressionContextForTest> expCtx(new ExpressionContextForTest());
+    VariablesParseState vps = expCtx->variablesParseState;
+
+    //All constants should optimize to ExpressionConstant
```
cswanson310: It looks like you haven't run clang-format on this code. Comments should be separated from the `//` by a space. Also, use a full sentence here, and end with a period.
`// Test that an object with constant values for every field will optimize to a constant value.`

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2864,6 +2864,24 @@ TEST(ExpressionObjectOptimizations, OptimizingAnObjectShouldOptimizeSubExpressio
...
+    auto object1 =  ExpressionObject::parse(expCtx, BSON("b" << 1 << "c" << 1),vps);
```
cswanson310: Don't use numbers in your variable names. If these tests are different, split them out into separate tests, otherwise, feel free to re-use the variable names and replace their values.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2864,6 +2864,24 @@ TEST(ExpressionObjectOptimizations, OptimizingAnObjectShouldOptimizeSubExpressio
...
+    auto optimized1 = object1->optimize();
+    auto constant1 = dynamic_cast<ExpressionConstant*>(optimized1.get());
+    ASSERT_TRUE(constant1);
+   
+   //Not all constants should not optimize to ExpressionConstant
+    auto object2=  ExpressionObject::parse(expCtx, BSON("b"<< 1 << "input" << "$inputField"),vps);
+    auto optimized2 = object2->optimize();
+    auto constant2 = dynamic_cast<ExpressionConstant*>(optimized2.get());
+    ASSERT_FALSE(constant2);
```
cswanson310: Please add a test case that shows that we optimize the inner expressions before checking if they are constant. Specifically, if one field is an expression which evaluates to a constant, and the rest of the fields are constant, then the entire object should optimize to a constant.

Something like `BSON("willBeConstant" << BSON("$add" << BSON_ARRAY(1 << 2)) << "alreadyConstant" << "string"));`

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2864,6 +2864,41 @@ TEST(ExpressionObjectOptimizations, OptimizingAnObjectShouldOptimizeSubExpressio
     ASSERT_VALUE_EQ(expConstant->evaluate(Document()), Value(3));
```
cswanson310: I think you can adapt this test to suit your needs.

Instead of asserting that the inner thing is a constant, assert that the 'optimized' variable itself is an ExpressionConstant, and that it has value `Value(Document{{"a", 3}})`

src/mongo/db/pipeline/expression.cpp
```diff
@@ -1729,9 +1727,17 @@ intrusive_ptr<ExpressionObject> ExpressionObject::parse(
 }
 
 intrusive_ptr<Expression> ExpressionObject::optimize() {
+    bool allValuesConstant = true;
     for (auto&& pair : _expressions) {
+        if (!dynamic_cast<ExpressionConstant*>(pair.second.get())) {
+            allValuesConstant = false;
+        }
         pair.second = pair.second->optimize();
```
cswanson310: I think this line needs to go above the check of whether it is a constant value in order for your tests to work.

KevinCybura:
I believe I fixed the test case and hopefully formatted everything
correctly and for the other tickets i will just work on them and make a
pull request for them next time we me and enjoy your honeymoon.

PR #1214, opened 2018-02-15
SERVER-33345 Update mongodb.conf to match current bindIpAll and bindIp behavior

juanpaulo:
Resolves: SERVER-33345


juanpaulo:
Applied requested changes. Thanks for reviewing @Schubes !

PR #1215, opened 2018-02-16
SERVER-9306 support Invisible index in MongoDB

zhihuiFan:
Hi:
   This pull request is about  https://jira.mongodb.org/browse/SERVER-9306 and https://jira.mongodb.org/browse/SERVER-26589
   
   The user story of this feature is something like this:
1. DBA want to drop an index
2. DBA is not confident enough to know if the index is not need at all.
3. The table is huge so recreating the index after drop will take a long time.
 
  In this situation,  we can make the index `invisible` for the following few days,   which means mongodb will still maintain index for the following data change but the optimizer will not use it any more.
If we found the index is still needed,  we can simple rollback our change quickly by making it visible again.   If we found the index is not needed for the following N days,  we can drop it safely.
 
The reason why “indexStats” command in MongoDB 3.2 is not perfect for this user case is that the optimizer may choose a wrong index, so it will still show the index is in use but in fact it can be and should be dropped.
 
Here is an overview for this feature:
 ```
> for(var i = 0; i<10000; i++) { db.i.save({a: i})}
> db.i.createIndex({a: 1})
{
                "createdCollectionAutomatically" : false,
                "numIndexesBefore" : 1,
                "numIndexesAfter" : 2,
                "ok" : 1
}
 
> db.i.getIndices()
[
                {
                                "v" : 2,
                                "key" : {
                                                "_id" : 1
                                },
                                "name" : "_id_",
                                "ns" : "zhifan.i",
                                "invisible" : false
                },
                {
                                "v" : 2,
                                "key" : {
                                                "a" : 1
                                },
                                "name" : "a_1",
                                "ns" : "zhifan.i",
                                "invisible" : false
                }
]
 
> db.i.find({a: 100}).explain()
{
                "queryPlanner" : {
                                "plannerVersion" : 1,
                                "namespace" : "zhifan.i",
                                "indexFilterSet" : false,
                                "parsedQuery" : {
                                                "a" : {
                                                                "$eq" : 100
                                                }
                                },
                                "winningPlan" : {
                                                "stage" : "FETCH",
                                                "inputStage" : {
                                                                "stage" : "IXSCAN",
                                                                "keyPattern" : {
                                                                                "a" : 1
                                                                },
                                                                "indexName" : "a_1",
                                                                "isMultiKey" : false,
                                                                "multiKeyPaths" : {
                                                                                "a" : [ ]
                                                                },
                                                                "isUnique" : false,
                                                                "isSparse" : false,
                                                                "isPartial" : false,
                                                                "indexVersion" : 2,
                                                                "direction" : "forward",
                                                                "indexBounds" : {
                                                                                "a" : [
                                                                                                "[100.0, 100.0]"
                                                                                ]
                                                                }
                                                }
                                },
                                "rejectedPlans" : [ ]
                },
                "serverInfo" : {
                                "host" : "zhifan-dev16",
                                "port" : 27017,
                                "version" : "3.7.0-353-g2307b7a",
                                "gitVersion" : "2307b7ae2495b4b1c0caa0a83d7be1323b975fe4"
                },
                "ok" : 1
}
 
> db.runCommand({"collMod": "i", "index": {"name": "a_1", "invisible": true}})
{ "invisible_old" : false, "invisible_new" : true, "ok" : 1 }
> db.i.find({a: 100}).explain()
{
                "queryPlanner" : {
                                "plannerVersion" : 1,
                                "namespace" : "zhifan.i",
                                "indexFilterSet" : false,
                                "parsedQuery" : {
                                                "a" : {
                                                                "$eq" : 100
                                                }
                                },
                                "winningPlan" : {
                                                "stage" : "COLLSCAN",
                                                "filter" : {
                                                                "a" : {
                                                                                "$eq" : 100
                                                                }
                                                },
                                                "direction" : "forward"
                                },
                                "rejectedPlans" : [ ]
                },
                "serverInfo" : {
                                "host" : "zhifan-dev16",
                                "port" : 27017,
                                "version" : "3.7.0-353-g2307b7a",
                                "gitVersion" : "2307b7ae2495b4b1c0caa0a83d7be1323b975fe4"
                },
                "ok" : 1
}
> db.runCommand({"collMod": "i", "index": {"name": "a_1", "invisible": false}})
{ "invisible_old" : true, "invisible_new" : false, "ok" : 1 }
> db.i.find({a: 100}).explain()
{
                "queryPlanner" : {
                                "plannerVersion" : 1,
                                "namespace" : "zhifan.i",
                                "indexFilterSet" : false,
                                "parsedQuery" : {
                                                "a" : {
                                                                "$eq" : 100
                                                }
                                },
                                "winningPlan" : {
                                                "stage" : "FETCH",
                                                "inputStage" : {
                                                                "stage" : "IXSCAN",
                                                                "keyPattern" : {
                                                                                "a" : 1
                                                                },
                                                                "indexName" : "a_1",
                                                                "isMultiKey" : false,
                                                                "multiKeyPaths" : {
                                                                                "a" : [ ]
                                                                },
                                                                "isUnique" : false,
                                                                "isSparse" : false,
                                                                "isPartial" : false,
                                                                "indexVersion" : 2,
                                                                "direction" : "forward",
                                                                "indexBounds" : {
                                                                                "a" : [
                                                                                                "[100.0, 100.0]"
                                                                                ]
                                                                }
                                                }
                                },
                                "rejectedPlans" : [ ]
                },
                "serverInfo" : {
                                "host" : "zhifan-dev16",
                                "port" : 27017,
                                "version" : "3.7.0-353-g2307b7a",
                                "gitVersion" : "2307b7ae2495b4b1c0caa0a83d7be1323b975fe4"
                },
                "ok" : 1
}
 ```
all the existing test case are passed  with `python buildscripts/resmoke.py jstests/core/*.js` and code is formated with `python buildscripts/clang_format.py format`

### DIFF
--- a/jstests/core/hidden_index.js
+++ b/jstests/core/hidden_index.js
@@ -1,0 +1,114 @@
+// Test expected behavior for hidden indexes. A hidden index is invisible to the query planner so
+// it will not be used in planning. It is handled in the same way as other indexes by the index
+// catalog and for TTL purposes.
+
+(function () {
+    'use strict';
+    load("jstests/libs/analyze_plan.js");  // For getPlanStages.
+    const testtab = db["hidden_index"];
+
+    function usedIndex(query) {
+	const explain = assert.commandWorked(testtab.find(query).explain());
+	const ixScans = getPlanStages(explain.queryPlanner.winningPlan, "IXSCAN");
+	return ixScans.length === 1;
+    }
+
+    function HiddenCheckTest(query, index_type, wildcard) {
+	let index_name;
+	if (wildcard)
+	    index_name = 'a.$**_' + index_type;
+	else
+	    index_name = 'a_' + index_type;
+
+	testtab.dropIndexes();
+	if (wildcard)
+	    assert.commandWorked(testtab.createIndex({"a.$**": index_type}));
+	else
+	    assert.commandWorked(testtab.createIndex({"a": index_type}));
+	
+	assert.eq(testtab.getIndexes()[1].hidden, undefined);
+	assert(usedIndex(query));
+	
+	assert.commandWorked(testtab.hideIndex(index_name));
+	assert(testtab.getIndexes()[1].hidden);
+	if (index_type === "text") {
+	    // errmsg: error processing query: ns=test.hidden_indexTree: TEXT : query=java,
+	    // language=english, caseSensitive=0, diacriticSensitive=0,
+	    // tag=NULL\nSort: {}\nProj: {}\n planner returned error ::
+	    // caused by :: need exactly one text index for $text query
+	    assert.throws(() => {testtab.find(query).explain();});
+	    return;
+	}
+	assert(!usedIndex(query));
+	
+	assert.commandWorked(testtab.unhideIndex(index_name));
+	assert.eq(testtab.getIndexes()[1].hidden, undefined);
+	assert(usedIndex(query));
+	
+	assert.commandWorked(testtab.dropIndex(index_name));
+
+	if (wildcard)
+	    assert.commandWorked(testtab.createIndex({"a.$**": index_type}, {hidden: true}));
+	else
+	    assert.commandWorked(testtab.createIndex({"a": index_type}, {hidden: true}));
+
+	assert(testtab.getIndexes()[1].hidden);
+	assert(!usedIndex(query));
+    }
+
+    
+    // Normal index testing.
+    HiddenCheckTest({a: 1}, 1);
+    // GEO index testing.
+    HiddenCheckTest({a: { $geoWithin :
+			  { $geometry :
+			    { type : "Polygon" ,
+                              coordinates : [ [
+                                  [ 0 , 0 ] ,
+                                  [ 3 , 6 ] ,
+                                  [ 6 , 1 ] ,
+                                  [ 0 , 0 ]
+                              ] ]
+			    } } } }, "2dsphere");
+
+    // Fts index.
+    HiddenCheckTest({ $text: { $search: "java" } }, "text");
+
+    // Wildcard index.
+    HiddenCheckTest({"a.f": 1}, 1, true);
+
+    // Hidden index on capped collection.
+    assert(testtab.drop());
+    assert.commandWorked(db.createCollection("hidden_index", {capped: true, size: 100}));
+    HiddenCheckTest({a: 1}, 1);
+    
+    // Can't hide index on system index.
+    const systemtab = db.getSiblingDB('admin').system.version;
+    assert.commandWorked(systemtab.createIndex({a: 1}));
+    assert.commandFailedWithCode(systemtab.hideIndex("a_1"), 2);
+    assert.commandWorked(systemtab.dropIndex("a_1"));
+    assert.commandFailedWithCode(systemtab.createIndex({a: 1}, {hidden: true}), 2);
+
+    // Can't hide _id index.
+    assert.commandFailed(testtab.hideIndex("_id_"));
+
+    // We can change ttl index and hide info at the same time.
+    assert.commandWorked(testtab.dropIndexes());
+    assert.commandWorked(testtab.createIndex({"tm": 1}, {expireAfterSeconds: 10}));
+    assert.eq(testtab.getIndexes()[1].hidden, undefined);
+    assert.eq(testtab.getIndexes()[1].expireAfterSeconds, 10);
+
+    db.runCommand({"collMod": testtab.getName(), "index": {"name": "tm_1",
+							   "expireAfterSeconds": 1,
+							   "hidden": true
+							  }});
+    assert(testtab.getIndexes()[1].hidden);
+    assert.eq(testtab.getIndexes()[1].expireAfterSeconds, 1);
+
+    // Hidden index is only allowed with FCV 4.4.
+    assert.commandWorked(db.adminCommand({setFeatureCompatibilityVersion: "4.2"}));
+    assert.commandFailedWithCode(testtab.createIndex({b: 1}, {hidden: true}), 31449);
+    assert.commandWorked(testtab.createIndex({b: 1}));
+    assert.commandFailedWithCode(testtab.hideIndex("b_1"), 2);
+    assert.commandWorked(testtab.unhideIndex("tm_1"));
+})();
--- a/jstests/noPassthrough/ttl_hidden_index.js
+++ b/jstests/noPassthrough/ttl_hidden_index.js
@@ -1,0 +1,26 @@
+// Make sure the TTL index still work after we hide it
+(function() {
+"use strict";
+let runner = MongoRunner.runMongod({setParameter: "ttlMonitorSleepSecs=1"});
+let coll = runner.getDB("test").ttl_hiddenl_index;
+coll.drop();
+
+// Create TTL index.
+assert.commandWorked(coll.ensureIndex({x: 1}, {expireAfterSeconds: 0}));
+let now = new Date();
+assert.commandWorked(coll.insert({x: now}));
+assert.commandWorked(coll.insert({x: now}));
+
+assert.commandWorked(coll.hide_index("x_1"));
+// Wait for the TTL monitor to run at least twice (in case we weren't finished setting up our
+// collection when it ran the first time).
+var ttlPass = coll.getDB().serverStatus().metrics.ttl.passes;
+assert.soon(function() {
+    return coll.getDB().serverStatus().metrics.ttl.passes >= ttlPass + 2;
+}, "TTL monitor didn't run before timing out.");
+
+assert.eq(coll.count(), 0, "We should get 0 documents after TTL monitor run");
+    
+MongoRunner.stopMongod(runner);
+    
+});
--- a/src/mongo/db/catalog/coll_mod.cpp
+++ b/src/mongo/db/catalog/coll_mod.cpp
@@ -69,6 +69,7 @@ MONGO_FAIL_POINT_DEFINE(assertAfterIndexUpdate);
 struct CollModRequest {
     const IndexDescriptor* idx = nullptr;
     BSONElement indexExpireAfterSeconds = {};
+    BSONElement indexHidden = {};
     BSONElement viewPipeLine = {};
     std::string viewOn = {};
     BSONElement collValidator = {};
@@ -125,14 +126,19 @@ StatusWith<CollModRequest> parseCollModRequest(OperationContext* opCtx,
             }
 
             cmr.indexExpireAfterSeconds = indexObj["expireAfterSeconds"];
-            if (cmr.indexExpireAfterSeconds.eoo()) {
-                return Status(ErrorCodes::InvalidOptions, "no expireAfterSeconds field");
+            cmr.indexHidden = indexObj["hidden"];
+
+            if (cmr.indexExpireAfterSeconds.eoo() && cmr.indexHidden.eoo()) {
+                return Status(ErrorCodes::InvalidOptions, "no expireAfterSeconds or hidden field");
             }
-            if (!cmr.indexExpireAfterSeconds.isNumber()) {
+            if (!cmr.indexExpireAfterSeconds.eoo() && !cmr.indexExpireAfterSeconds.isNumber()) {
                 return Status(ErrorCodes::InvalidOptions,
                               "expireAfterSeconds field must be a number");
             }
-
+            if (!cmr.indexHidden.eoo() && !cmr.indexHidden.isBoolean()) {
+                return Status(ErrorCodes::InvalidOptions,
+                              "hidden field must be a boolean");
+            }
             if (!indexName.empty()) {
                 cmr.idx = coll->getIndexCatalog()->findIndexByName(opCtx, indexName);
                 if (!cmr.idx) {
@@ -161,15 +167,16 @@ StatusWith<CollModRequest> parseCollModRequest(OperationContext* opCtx,
                 cmr.idx = indexes[0];
             }
 
-            BSONElement oldExpireSecs = cmr.idx->infoObj().getField("expireAfterSeconds");
-            if (oldExpireSecs.eoo()) {
-                return Status(ErrorCodes::InvalidOptions, "no expireAfterSeconds field to update");
+            if (!cmr.indexExpireAfterSeconds.eoo()) {
+                BSONElement oldExpireSecs = cmr.idx->infoObj().getField("expireAfterSeconds");
+                if (oldExpireSecs.eoo()) {
+                    return Status(ErrorCodes::InvalidOptions, "no expireAfterSeconds field to update");
+                }
+                if (!oldExpireSecs.isNumber()) {
+                    return Status(ErrorCodes::InvalidOptions,
+                                  "existing expireAfterSeconds field is not a number");
+                }
             }
-            if (!oldExpireSecs.isNumber()) {
-                return Status(ErrorCodes::InvalidOptions,
-                              "existing expireAfterSeconds field is not a number");
-            }
-
         } else if (fieldName == "validator" && !isView) {
             // Save this to a variable to avoid reading the atomic variable multiple times.
             const auto currentFCV = serverGlobalParams.featureCompatibility.getVersion();
@@ -250,13 +257,25 @@ class CollModResultChange : public RecoveryUnit::Change {
 public:
     CollModResultChange(const BSONElement& oldExpireSecs,
                         const BSONElement& newExpireSecs,
+                        const BSONElement& oldHidden,
+                        const BSONElement& newHidden,
                         BSONObjBuilder* result)
-        : _oldExpireSecs(oldExpireSecs), _newExpireSecs(newExpireSecs), _result(result) {}
+        : _oldExpireSecs(oldExpireSecs), _newExpireSecs(newExpireSecs),
+          _oldHidden(oldHidden), _newHidden(newHidden),
+          _result(result) {
+    }
 
     void commit(boost::optional<Timestamp>) override {
         // add the fields to BSONObjBuilder result
-        _result->appendAs(_oldExpireSecs, "expireAfterSeconds_old");
-        _result->appendAs(_newExpireSecs, "expireAfterSeconds_new");
+        if (!_oldExpireSecs.eoo()) {
+            _result->appendAs(_oldExpireSecs, "expireAfterSeconds_old");
+            _result->appendAs(_newExpireSecs, "expireAfterSeconds_new");
+        }
+        if (!_newHidden.eoo()) {
+            bool oldValue = _oldHidden.eoo() ? false : _oldHidden.booleanSafe();
+            _result->append("hidden_old", oldValue);
+            _result->appendAs(_newHidden, "hidden_new");
+        }
     }
 
     void rollback() override {}
@@ -264,6 +283,8 @@ public:
 private:
     const BSONElement _oldExpireSecs;
     const BSONElement _newExpireSecs;
+    const BSONElement _oldHidden;
+    const BSONElement _newHidden;
     BSONObjBuilder* _result;
 };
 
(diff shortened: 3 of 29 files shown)


zhihuiFan:
hi @amschwerin  and @asya999,  could you help on this case as well?  Thanks

dstorch:
Hi @zhihuiFan ,

This looks great, thanks for contributing. Very nice work. Before I do a complete review, and we go through all the details, I am going to summarize the proposed change and pass it through our internal feature design review process. This might take a week or two, so stay tuned! The main question I want to bring to the design review is about high-level implementation approach: Should we take this approach in which the `invisible` flag persists and is replicated, or should we instead implement this with some in-memory state that can be configured per-process? The former approach has obvious advantages, but the latter approach is simpler to implement and has the advantage of being separately configurable on each mongod in the cluster.

The other complication here has to do with upgrade/downgrade scenarios. The changes in this patch will be released as part of 4.0. In order to upgrade to 4.0, users will have mixed version replica sets consisting of both 3.6 and 4.0 nodes. For upgrade and downgrade to function correctly, 3.6/4.0 mixed version replica sets must be fully supported.

The difficulty is that a 3.6 secondary syncing from a 4.0 primary will not correctly handle the `invisible` flag, since invisible indexes are not supported by 3.6. In fact, a 3.6 secondary will hit a process-fatal assertion when encountering an index spec field is does not recognize:

https://github.com/mongodb/mongo/blob/r3.6.2/src/mongo/db/catalog/index_catalog_impl.cpp#L532-L537

Thus, users should only be able to create an invisible index when the cluster has been fully upgraded to 4.0. This is handled by upgrading all of the binaries and then setting the `featureCompatibilityVersion` to "4.0". The 3.6 <=> 4.0 upgrade/downgrade hasn't yet been documented, since 4.0 is still under development, but here is documentation about how things worked for the 3.4 => 3.6 upgrade of a replica set:

https://docs.mongodb.com/manual/release-notes/3.6-upgrade-replica-set/

In particular, note the step of enabling backwards incompatible features by setting the `featureCompatibilityVersion` (FCV): https://docs.mongodb.com/manual/release-notes/3.6-upgrade-replica-set/#enable-backwards-incompatible-newversion-features. Invisible indexes are an example of a 3.6 backwards incompatible feature. It should be illegal to create such an index unless the FCV is fully upgraded to "4.0". Luckily, the index spec validation code already has the FCV threaded through to it:

https://github.com/mongodb/mongo/blob/d1ba78a5a83b58275aed657d71b58a57847e7027/src/mongo/db/catalog/index_key_validate.cpp#L217

So, there will be some work to do around handling and testing multiversion scenarios. Let me know if anything here doesn't make sense. I'll get back to you in a bit about the results of our internal design review discussion.

Best,
Dave

zhihuiFan:
Thanks @dstorch  for the review, your explanation make senses and the solution for the incompatible issue is amazing - clear, simple, clean and works!

As a DBA,  I dislike some in-memory features personally since we need extra effort to make them still work after restarting (like IndexFilter),  if we ignore that,  we may have some short time production issue.

Looks forward your suggestion and I will think/test more upgrade scenarios.

zhihuiFan:
hi @dstorch :  Any update on this?

dstorch:
Hi @zhihuiFan ,

My apologies for the delay—things have been busy here due to our work finishing up the 4.0 release. Unfortunately, the MongoDB Query Team is just entering a stabilization period in which we will not be adding any new features, but rather will be working only on bug fixes, performance work, and testing. We expect it to be a few months until we branch, which will permit us to begin feature development work for version 4.2.

Feel free to reach out with any questions!

Best,
Dave

zhihuiFan:
hi @dstorch ,
Thanks for your information,  that arrangement does make sense!   There are still some features I want from MongoDB.  Like
1.  Persist index filter  https://jira.mongodb.org/browse/SERVER-17625         
2. MongoDB user resource limits     https://jira.mongodb.org/browse/SERVER-15072  

could I just develop them, open the PRs and wait for MongoDB 4.2 feature branch?

src/mongo/db/catalog/coll_mod.cpp
```diff
@@ -164,14 +168,17 @@ StatusWith<CollModRequest> parseCollModRequest(OperationContext* opCtx,
 
                 cmr.idx = indexes[0];
             }
+            if (!cmr.indexExpireAfterSeconds.eoo()) {
 
-            BSONElement oldExpireSecs = cmr.idx->infoObj().getField("expireAfterSeconds");
```
eternalme: The  deletion and addition are same except few line breaks.

asya999:
@zhihuiFan,
 
We have released 4.0 and are working on 4.2 features now.   As part of that process we create scope documents for all proposed new features.   

I want to clarify something about our process.  When a new feature or functionality is considered, if it has user facing API or parts that interact with other components of the system, we have a process to decide and agree what the scope and syntax of the API or command will be and how exactly the new functionality will interact with existing features, with replication, with sharding, version upgrades, tools, drivers, etc.

Unfortunately the tickets that you flagged are definitely blocked by there being no consensus on their exact functionality, nor syntax.  In general if the SERVER ticket has "fixVersion" like 'Needs Further Definition' or 'features we are not sure of' it means that we aren't sure we want to implement the feature in the server (at least not until there is more clarity on exactly how this feature would work).

For features that are in "fixVersion" 'Backlog' it means we want the feature implemented at some point, but if it has syntax that's user facing then we would need to agree on syntax before it makes sense for you or anyone else to work on it.   We are currently working on how we can expose this information so you know when you see a ticket whether or not it's "ready for work".   

I'm very sorry this is taking a long time, much longer than I would want, but we want to make sure that the process is clear and ultimately maximize the chance community contributions make it into the code base.

Specifically for this ticket (SERVER-9306), we need to schedule internal review for invisible indexes feature - part of that review will be to decide on exact scope, whether it's a persistent feature that replicates and needs to support feature compatibility level for mixed version clusters during upgrade, or not, and what the actual syntax for it will be.

Asya Kamsky

zhihuiFan:
Thanks @asya999 for your clarify.   

When I send a PR,  actually I want to learn something from author's review & suggestion (at the mean time, contribute some my effort).   All the processes you just mentioned totally matches my expectation, except the long time response:(.     If the time can be shorter, that would be perfect!

As for this PR,  I will not do any changes before the internal review completed.   After that,  I can do the changes accordingly. 

Thanks for your time!

funny-falcon-at-joomcode:
Aaaaahhhhhhh.... It is super useful feature! Why hasn't it merged yet? It will simplify maintaining a lot.

jameswahlin:
Hi @zhihuiFan,

Our apologies on the delay in getting back to you. We are starting work on scope and syntax for this feature soon. If you are still interested in delivering this feature, we can provide scope/syntax to you once completed and work with you through the pull request process. We would like to include this in our next major release, MongoDB 4.4. In order to do so development would need to be completed by the end of February.

Please let me know if you would like to continue working with us on this.

Best,
James Wahlin

zhihuiFan:
Hi @jameswahlin : 

Glad to hear that you are starting to work on this.  sure,  let's finish this task cleanly.  

Thanks

jameswahlin:
@zhihuiFan - I am in the process of reviewing your commit and hope to have feedback for you tomorrow. In the mean time can you sign our contributor agreement? This is required by us in order to accept your commit. It looks like we have one on file for you, but for a former company. 

https://www.mongodb.com/legal/contributor-agreement

zhihuiFan:
> @zhihuiFan - I am in the process of reviewing your commit and hope to have feedback for you tomorrow. In the mean time can you sign our contributor agreement? This is required by us in order to accept your commit. It looks like we have one on file for you, but for a former company.
> 
> https://www.mongodb.com/legal/contributor-agreement

I resigned with my current company

zhihuiFan:
Thanks for your detailed explanation.  I am agreed with most of them,  but the following need some more discussion. 

> * Test that confirms that index hiding is only possible in FCV 4.4 and that unhiding is possible in both FCV 4.4 and 4.2.

If we have this command supported,  we are sure the binary can handle the hidden index as well.   unless we would say the hidden index can't work even with this binary,  if we design like this,  I think we need to check FCV during handling  index-selection as well.  

Basically I prefer to not add this limitation. 


> * Test that confirms a binary 4.2 mongod will fail to start gracefully when the index catalog has a hidden index.
> * Test that a 4.2 binary initial-syncing from a binary 4.4 / FCV 4.2 node will fail gracefully when encountering a hidden index.
> * Test that a 4.2 binary that receives an index create or collMod oplog entry with a hidden index flag fails gracefully.

Looks this need a changes on the 4.2 binary rather than this one.  this is a work can't be merged into this pull request.

jstests/core/hidden_index.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+var primary = rsTest.getPrimary();
+var secondary = rsTest.getSecondary();
+secondary.setSlaveOk();
+
+var testtab = primary.getDB("test").hidden_index_tab;
+var testtabsec = secondary.getDB("test").hidden_index_tab;
+
+testtab.createIndex({a: 1});
```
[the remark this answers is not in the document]
zhihuiFan: done

zhihuiFan:
> Hi Zhihui,
> 
> Nice work on this commit! In addition to the comments left on this patch there are also tests that I would like to see prior to merge. As discussed earlier, it may be easier for MongoDB to write some or all of these tests (and can do so) given there would be a learning curve with our test framework and as an external contributor you do not have access to test via our continuous integration environment. That said, if you would like to take any of them on you are welcome and I can provide guidance. The testing I have identified so far is as follows:

I have tested most of of them with js testcase.  include 2,4, 6,7, 8.   Following is the reason why I didn't add them for the rest 3. 

> 1. Confirm existence of the "hidden" flag in db.currentOp() collMod and createIndexes entries and the profiler.

In JS framework,  I don't know how to parallel 2 operations, then I have no idea how the db.currentOp() can be tested.  

> 3. Confirm that TTL document removal continues to work after an index is hidden
I have tested it manually, but the time when the TTL thread will be waked up to remove the old data is out of control in js framework and I don't want to sleep too long time in js test case.

> 5. Confirm that an index backing a shard key cannot be dropped via mongos.

I guess you mean "hide" rather than "drop". I found we can drop the index backing a shard key currently. what is the purpose of this check for hidden index?

I think most part of this feature has been completed.  and you are free to do some changes on this patch if you think that will be more efficient.  

Thanks

jstests/core/hidden_index.js
```diff
@@ -0,0 +1,113 @@
...
+
+    // hidden index on capped collection
+    assert(testtab.drop());
+    assert.commandWorked(db.createCollection(jsTest.name(), {capped: true, size: 100}));
+    HiddenCheckTest({a: 1}, 1);
+    
+    // can't hide index on system index
+    const systemtab = db.getSiblingDB('admin').system.version;
+    assert.commandWorked(systemtab.createIndex({a: 1}));
+    assert.commandFailed(systemtab.hideIndex("a_1"));
```
jameswahlin: Can you substitute the assert.commandFailed() calls in this section with assert.commandFailedWithCode()? You would then pass the error code as a second argument. This is useful in that it will alert us if the reason a test fails has changed.

jstests/core/hidden_index.js
```diff
@@ -0,0 +1,113 @@
...
+	    assert.commandWorked(testtab.createIndex({"a.$**": index_type}, {hidden: true}));
+	else
+	    assert.commandWorked(testtab.createIndex({"a": index_type}, {hidden: true}));
+
+	assert(testtab.getIndexes()[1].hidden);
+	assert(!usedIndex(query));
+    }
+
+    
+    // normal index
```
jameswahlin: Please start these comments with a capital letter and end with a period.

jstests/core/hidden_index.js
```diff
@@ -0,0 +1,113 @@
...
+    const systemtab = db.getSiblingDB('admin').system.version;
+    assert.commandWorked(systemtab.createIndex({a: 1}));
+    assert.commandFailed(systemtab.hideIndex("a_1"));
+    assert.commandWorked(systemtab.dropIndex("a_1"));
+    assert.commandFailed(systemtab.createIndex({a: 1}, {hidden: true}));
+
+    // can't hide _id index
+    assert.commandFailed(testtab.hideIndex("_id_"));
+
+    // we can change ttl index and hide info at the same timex
```
jameswahlin: Typo: timex

jstests/core/hidden_index.js
```diff
@@ -0,0 +1,113 @@
...
+    assert.eq(testtab.getIndexes()[1].expireAfterSeconds, 10);
+
+    db.runCommand({"collMod": testtab.getName(), "index": {"name": "tm_1",
+							   "expireAfterSeconds": 1,
+							   "hidden": true
+							  }});
+    assert(testtab.getIndexes()[1].hidden);
+    assert.eq(testtab.getIndexes()[1].expireAfterSeconds, 1);
+
+    // after we hide the index, the ttl still work as well
```
jameswahlin: Was this comment meant to be paired with an additional test? I think confirming that documents are deleted via TTL on a hidden index should be a separate test and probably one with its own test fixture. An example of a similar test is: jstests/noPassthrough/ttl_partial_index.js

jstests/core/hidden_index.js
```diff
@@ -0,0 +1,113 @@
+// test the feature of hidden index,
```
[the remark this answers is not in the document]
zhihuiFan: done

zhihuiFan:
> Please see my comments for additional changes requested.
> 
Done. 

> Also, can you confirm whether you would like MongoDB to write the tests listed in the following comment? [#1215 (review)](https://github.com/mongodb/mongo/pull/1215#pullrequestreview-356844354)

I just added the test case for item 3.  Currently  item  1,5,9 are left.   9 looks impossible without your internal testing framework.   I also replied you for my question about 1 & 5. 
I would prefer to leave the 3 items for MongoDB since that would be more effective.  I have marked this pull request as "allow the edit from the maintainer".

jstests/noPassthrough/ttl_hidden_index.js
```diff
@@ -0,0 +1,26 @@
+// Make sure the TTL index still work after we hide it
+(function() {
+"use strict";
+var runner = MongoRunner.runMongod({setParameter: "ttlMonitorSleepSecs=1"});
```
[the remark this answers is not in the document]
zhihuiFan: done

src/mongo/db/catalog/coll_mod.cpp
```diff
@@ -360,40 +393,66 @@ Status _collModInternal(OperationContext* opCtx,
...
+                if (SimpleBSONElementComparator::kInstance.evaluate(oldExpireSecs != newExpireSecs)) {
+                    // Change the value of "expireAfterSeconds" on disk.
+                    DurableCatalog::get(opCtx)->updateTTLSetting(opCtx,
+                                                                 coll->getCatalogId(),
+                                                                 cmrOld.idx->indexName(),
+                                                                 newExpireSecs.safeNumberLong());
+                }
+            }
+
+            // User wants to hide this index
```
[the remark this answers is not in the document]
zhihuiFan: done

src/mongo/db/storage/bson_collection_catalog_entry.cpp
```diff
@@ -120,6 +120,26 @@ void BSONCollectionCatalogEntry::IndexMetaData::updateTTLSetting(long long newEx
     spec = b.obj();
 }
 
+
+void BSONCollectionCatalogEntry::IndexMetaData::updateHiddenSetting(bool hidden) {
+    // if hidden == false, we remove this field from catalog rather than
```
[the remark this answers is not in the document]
zhihuiFan: done

zhihuiFan:
> 3. Add upgrade / downgrade tests:
>    While we hope that changes to the 4.2 release will not be required to satisfy upgrade/downgrade testing, it is not certain that this will be the case. Also worth noting is that our fuzzer tests are not public so not modifiable via pull request. Given that I agree that it makes sense for MongoDB to write these tests.

Hi @jamesfcohan : 
Sorry for the long response,  I have to say I doesn't work on MongoDB projects any more (as you may have noticed,  this patch was submitted 2 year ago),   so I have very limited time on this.  then Thanks, I would leave it for MongoDB to write these test case. 

for others my latest push has fixed them.

jameswahlin:
> Hi @jamesfcohan :
> Sorry for the long response, I have to say I doesn't work on MongoDB projects any more (as you may have noticed, this patch was submitted 2 year ago), so I have very limited time on this. then Thanks, I would leave it for MongoDB to write these test case.

I understand completely. MongoDB will write the update/downgrade tests required for this feature.

src/mongo/db/catalog/index_key_validate.cpp
```diff
@@ -269,6 +271,12 @@ StatusWith<BSONObj> validateIndexSpec(
 
     boost::optional<IndexVersion> resolvedIndexVersion;
 
+    // Allow compound hashed index only if FCV is 4.4.
+    const auto isFeatureDisabled =
+                (featureCompatibility.isVersionInitialized() &&
```
dgottlieb: This call to `isVersionInitialized` also seems off to me. If this code can be called prior to FCV initialization, I would expect the `isFeatureDisabled` boolean to be `true`. If the code cannot be called prior to FCV initialization, leaving it as an embedded invariant inside `getVersion` seems best.

If I'm mistaken on both fronts, a comment describing the scenario would be beneficial.

dgottlieb: That being said, I now see this is moved from below. @jameswahlin can you verify that logic is actually intended?

src/mongo/db/catalog/index_key_validate.cpp
```diff
@@ -330,6 +334,19 @@ StatusWith<BSONObj> validateIndexSpec(
...
+                return {ErrorCodes::Error(31449),
+                        "Hidden indexes can only be created with FCV 4.4"};
+            }
+            if (indexSpecElem.type() != BSONType::Bool) {
+                return {ErrorCodes::TypeMismatch,
+                        str::stream()
+                            << "The field '" << IndexDescriptor::kIndexNameFieldName
+                            << "' must be a bool, but got " << typeName(indexSpecElem.type())};
+            }
+            hasHiddenField = true;
```
dgottlieb: I see this is a pattern in this function, but it doesn't seem to have a side-effect? Is that expected?

src/mongo/db/catalog/index_key_validate.cpp
jameswahlin: I believe this is the intended logic. @dgottlieb - your right, this is a bug in the current code. I have informed the author of compound hashed indexes to take a look.

@zhihuiFan, can you replace this with the following? This will correctly disallow hidden indexes if FCV has not been initialized:
const auto isFeatureDisabled =
                (!featureCompatibility.isVersionInitialized() ||
                 featureCompatibility.getVersion() <
                     ServerGlobalParams::FeatureCompatibility::Version::kFullyUpgradedTo44);

src/mongo/db/catalog/index_key_validate.cpp
zhihuiFan: Yes，that can be removed.

src/mongo/db/catalog/index_key_validate.cpp
zhihuiFan: Hi @jameswahlin :    I'm agreed with this change. Can I handle over this to MongoDB team to change it during the merge.  do we have any other comments?

jameswahlin: Hi @zhihuiFan - Dan Gottlieb posted a few additional comments as part of his review on March 9th. This was prior to my comment regarding FCV. Would you like MongoDB to address?

zhihuiFan: Yes, Thanks @jameswahlin

jameswahlin: We will preserve your latest commit and will complete work in additional commits so that you get credit for your work. I will leave this pull request open until we commit your changes so that we can document the commit hash here.

Thanks for your contribution @zhihuiFan! We appreciate your work and are excited to have this feature in MongoDB.

monkey101:
@zhihuiFan I just saw that @RuoxinXu just closed SERVER-9306 after merging your changes. Many thanks for your contribution!!

jameswahlin:
Thanks for working with us on this @zhihuiFan! This change was included under the following commit: https://github.com/mongodb/mongo/commit/bad7c538e7efbc996a6089e1569681edf24e6b33

lemontree-mxh:
Thanks you @jameswahlin and @monkey101 !

zhihuiFan:
Thanks you @jameswahlin and @monkey101 !

PR #1216, opened 2018-02-19
SERVER-32295 Support python 3

omron93:
Add python3 support for scons build system and `resmoke.py` tool to run tests. Python2 support shouldn't be affected.

To note: other packaging scripts are not changed, so they support only python2.

@acmorrow @bmanojlovic Please take a look. Sorry that it took so long.

(not tested for master branch, tested only for mongodb 3.6.2)

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -867,7 +867,7 @@ SConsignFile(str(sconsDataDir.File('sconsign')))
 def printLocalInfo():
     import sys, SCons
     print( "scons version: " + SCons.__version__ )
-    print( "python version: " + " ".join( [ repr(i) for i in sys.version_info ] ) )
+    print( "python version: " + " ".join( [ str(i) for i in sys.version_info ] ) )
 
 printLocalInfo()
 
@@ -920,7 +920,7 @@ envDict = dict(BUILD_ROOT=buildDir,
                MODULE_INJECTORS=dict(),
                ARCHIVE_ADDITION_DIR_MAP={},
                ARCHIVE_ADDITIONS=[],
-               PYTHON="/usr/bin/python2",
+               PYTHON=utils.find_python(),
                SERVER_ARCHIVE='${SERVER_DIST_BASENAME}${DIST_ARCHIVE_SUFFIX}',
                UNITTEST_ALIAS='unittests',
                # TODO: Move unittests.txt to $BUILD_DIR, but that requires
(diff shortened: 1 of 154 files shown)


acmorrow:
@omron93 - Thanks for the PR.

It does look like there are a few conflicts that have crept in over the past few weeks. My apologies, I was OOO. But it would be helpful if you could rebase this work to the head of master and force push to the PR.

Also, I presume that you actually need this work for 3.6 not master, so we will need to backport it. I think we should be able to take care of that from our side.

acmorrow:
Actually, I just realized a serious obstacle for our being able to accept this. Currently we rely on the built-in SCons 2.5.1 that we vendor to src/third_party, so that we have a consistent version of SCons across all of our CI systems.

That version of SCons doesn't support python3.

That means that I can't take this PR and then switch over to using python3 in our CI environment, even though that environment *does* have python3 everywhere (because we build it from source), because we don't also have a python3 capable SCons installed.

So we will have no way to ensure that python3 support doesn't regress, at least until we upgrade our vendored python3 to SCons 3.0 or newer.

bmanojlovic:
In openSUSE we are using system scons for python3 build and python2 internal for older releases , so it is not a problem to have it "vendored" and use python3 in same time depending on what you are building. 
https://build.opensuse.org/package/show/server:database/mongodb

acmorrow:
@bmanojlovic - Right, but my point was that even if we merge this, we won't be able to build with python3 in our CI system until we upgrade our vendored SCons. Without the ability to build with python3 in our CI system, the python3 support will bit-rot because engineers will check in py3 incompatible changes.

omron93:
> but my point was that even if we merge this, we won't be able to build with python3 in our CI system until we upgrade our vendored SCons. Without the ability to build with python3 in our CI system, the python3 support will bit-rot because engineers will check in py3 incompatible changes.

@acmorrow I understand. But from MPOV, better something than nothing. Even if you don't use it and test it right now, we'll do it in Fedora and as I understand @bmanojlovic use python3 also in openSUSE. And if something broke by some change, we can report it.

I know that you prefer to merge in upstream only code you use. But if you plan to use python3 anyway, adding python3 support now (and "finishing" it later when you update bundled scons) won't hurt you. On the contrary our downstream patches will be smaller and you'll get earlier feedback for python3.

jamespharvey20:
Ping?  python2 retires in 4 months and 22 days, and several distributions have removed python2, or at least banished it to unofficially supported repositories.

acmorrow:
Python 3 support was added per https://jira.mongodb.org/browse/SERVER-32295, so I'm closing this review.

PR #1219, opened 2018-02-24
SERVER-32300 Aggregation shell helper should not modify options document

Dandanlin0702:


jstests/aggregation/bugs/server32300.js
```diff
@@ -0,0 +1,19 @@
+// Core ServerSERVER-32300 Aggregation shell helper should not modify options document
+
+load("src/mongo/shell/explainable.js")
+
+let collation = {collation: {locale: "fr", backwards: true}};
+
+// Computes correct result of the aggragation
+// Computes an empty array
+jsTestLog("desired output: " + tojson(db.collection.aggregate([], collation).toArray()));
```
cswanson310: Instead of just logging here, we should make assertions on the behavior. In this case, I would recommend inserting some data into a collection, then asserting that the results of an aggregation are the same before and after an explain. Something like:
```
const coll = db.explain_options;
coll.drop();

for (let i = 0; i < 10; ++i) {
    assert.writeOK(coll.insert({_id: i}));
}

const firstResults = coll.aggregate([{$sort: {_id: 1}}], collation).toArray();
// Here we issue an explain in order to verify that 'collation' is not modified to include the explain flag. We do not care about the result of the explain.
assert.commandWorked(coll.explain().aggregate([], collation));
const secondResults = coll.aggregate([{$sort: {_id: 1}}], collation).toArray();
assert.eq(firstResults, secondResults);
```

src/mongo/shell/explainable.js
```diff
@@ -100,14 +100,14 @@ var Explainable = (function() {
...
             // we use the explain option to the aggregate command. Otherwise we issue an explain
             // command wrapping the agg command, which is supported by newer versions of the server.
             if (this._verbosity === "queryPlanner") {
-                extraOpts.explain = true;
-                return this._collection.aggregate(pipeline, extraOpts);
+                extraOptsCopy.explain = true;
+                return this._collection.aggregate(pipeline, extraOptsCopy);
             } else {
                 // The aggregate command requires a cursor field.
                 if (!extraOpts.hasOwnProperty("cursor")) {
```
cswanson310: Please use 'extraOptsCopy' throughout this whole function, in order to prevent a similar bug from creeping in in the future.

ksuarz:
Thanks for the fix @Dandanlin0702, this bug was really annoying :)

PR #1221, opened 2018-02-28
SERVER-33149 CreateIndexes fails to report an error when index is not created with the specified name

Dandanlin0702:


jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,20 @@
+(function() {
+    "use strict";
+
+    // For arrayEq.
+    load("jstests/aggregation/extras/utils.js");
+
+    let indexDB = db.getSiblingDB("create_indexess_db");
```
cswanson310: Use const for variables which do not change.

jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,20 @@
...
+    assert.commandWorked(indexDB.dropDatabase());
+
+    // Create an index with key: {x : 1}, name : "x_1
+    assert.commandWorked(indexDB.runCommand({ createIndexes : "test", indexes : [ { key : { x : 1 }, name : "x_1" } ] }));
```
cswanson310: All of the code in our repository is formatted by clang-format, including our javascript tests. It looks like this code needs to be formatted by clang-format. Please see here for a guide of that: https://github.com/mongodb/mongo/wiki/Server-Code-Style#clang-format

This applies to the C++ code as well - thanks.

jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,20 @@
...
+
+    // Create a duplicate index will not result an error
+    assert.commandWorked(indexDB.runCommand({ createIndexes : "test", indexes : [ { key : { x : 1 }, name : "x_1"} ] }), ErrorCodes.IndexAlreadyExists, "Identical index already exists ");
```
cswanson310: It looks like you're passing extra options to 'commandWorked' here - they are probably being ignored? If so, just delete them.
https://github.com/mongodb/mongo/blob/b7b58a38dd34ac09ed56eb1f306ea6d64f8b2d88/src/mongo/shell/assert.js#L588

jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,20 @@
...
+
+    // Create an index that has same key pattern but with different name will fail
+    assert.commandFailedWithCode(indexDB.runCommand({ createIndexes : "test", indexes : [ { key : { x : 1 }, name : "x_2"} ] }), ErrorCodes.CannotCreateIndex, "Index with the same key pattern already exists with a different name");
```
cswanson310: The third argument here is what will be printed if the assertion fails. I'd say you could leave it off, or say something like "expected not to be able to create index with different name and same options"

src/mongo/db/catalog/index_catalog_impl.cpp
```diff
@@ -813,8 +814,8 @@ Status IndexCatalogImpl::_doesSpecConflictWithExisting(OperationContext* opCtx,
                                             << " already exists with different options: "
                                             << desc->infoObj());
 
-            return Status(ErrorCodes::IndexAlreadyExists,
-                          str::stream() << "index already exists with different name: " << name);
+            return Status(ErrorCodes::CannotCreateIndex,
+                          str::stream() << "Index with the same keys but a diferent name already exists.");
```
cswanson310: Please use 'options' instead of 'keys' here. To me, 'keys' implies the object had the same field names, but not necessarily the same values for those field names.

jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,29 @@
+(function() {
+    "use strict";
+
+    // For arrayEq.
```
cswanson310: It doesn't look like you're using this?

jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,29 @@
...
+    load("jstests/aggregation/extras/utils.js");
+
+    const indexDB = db.getSiblingDB("create_indexess_db");
```
cswanson310: Extra 's'/

jstests/aggregation/bugs/server33149.js
```diff
@@ -0,0 +1,29 @@
...
+    assert.commandWorked(indexDB.dropDatabase());
+
+    // Create an index with key: {x : 1}, name : "x_1
```
cswanson310: This is a total nitpick, sorry.

For the comments in this file, please be sure to use full sentences, and end with period. Also, avoid making statements like "Duplicate index cannot be created" about what will or should happen, instead phrasing it with "Test that a duplicate index cannot be created".

Also, please wrap at 100 characters, and only use two lines if the thoughts/comments are very distinct. In your case, I think all the comments should be on one line.

src/mongo/db/catalog/index_catalog_impl.cpp
```diff
@@ -813,8 +813,9 @@ Status IndexCatalogImpl::_doesSpecConflictWithExisting(OperationContext* opCtx,
                                             << " already exists with different options: "
                                             << desc->infoObj());
 
-            return Status(ErrorCodes::IndexAlreadyExists,
-                          str::stream() << "index already exists with different name: " << name);
+            return Status(ErrorCodes::CannotCreateIndex,
```
cswanson310: I'm curious, and couldn't quite figure it out myself - do you how exactly the IndexAlreadyExists code was being swallowed and ignored?

Dandanlin0702: I believe it is because of this block of code in index_builder.cpp (line 150-160):

 try {
                status = indexer.init(_index).getStatus();
                if (status == ErrorCodes::IndexAlreadyExists ||
                    (status == ErrorCodes::IndexOptionsConflict && _relaxConstraints)) {
                    LOG(1) << "Ignoring indexing error: " << redact(status);
                    if (allowBackgroundBuilding) {
                        // Must set this in case anyone is waiting for this build.
                        _setBgIndexStarting();
                    }
                    return Status::OK();
                }

cswanson310: Oh, that totally makes sense. Thanks.

As discussed, this looks good to go, so I'll run it through the tests and see if everything passes.

src/mongo/db/catalog/index_catalog_impl.cpp
```diff
@@ -813,8 +813,9 @@ Status IndexCatalogImpl::_doesSpecConflictWithExisting(OperationContext* opCtx,
...
+                          str::stream()
+                              << "Index with the same options but a diferent name already exists.");
```
Parad0X: Typo in `different`

kelly-cs:
Fixed in 4.1.10

PR #1223, opened 2018-03-12
SERVER-25957 Optimize $filter + $arrayElemAt to avoid scanning entire array

KevinCybura:
Create a separate evaluate function called filterTillNthValue that takes a long long that will be the less than or equal to size of the vector that will contain our filtered values.


src/mongo/db/pipeline/expression.cpp
```diff
@@ -490,8 +490,15 @@ const char* ExpressionArray::getOpName() const {
 /* ------------------------- ExpressionArrayElemAt -------------------------- */
 
 Value ExpressionArrayElemAt::evaluate(const Document& root) const {
-    const Value array = vpOperand[0]->evaluate(root);
     const Value indexArg = vpOperand[1]->evaluate(root);
+    long long i = indexArg.coerceToLong();
```
cswanson310: This coerceToLong with throw an error if it's not a long, and that error message will not have as much context as the error that used to get thrown on line 511. Please move the assert on line 511 up before this, then use getLong instead: https://github.com/mongodb/mongo/blob/d50f8ba3c8f3f9380876361def45f8531247f8aa/src/mongo/db/pipeline/value.h#L203

src/mongo/db/pipeline/expression.cpp
```diff
@@ -490,8 +490,15 @@ const char* ExpressionArray::getOpName() const {
...
+
+    // If "vpOperand[0]" is an ExpressionFilter given index is greater than 0 call
+    // filterTillNthValue
+    const Value array = dynamic_cast<ExpressionFilter*>(vpOperand[0].get()) && (i >= 0)
```
cswanson310: It's a little strange to do a dynamic cast each and every time you evaluate this. The answer to this won't change, so I'd recommend doing the dynamic cast during optimize, and remembering via a boolean "_firstArgumentIsFilter" or something like that. I'm not sure how slow a dynamic cast is, but certainly not as slow as checking a boolean :)

src/mongo/db/pipeline/expression.cpp
```diff
@@ -490,8 +490,15 @@ const char* ExpressionArray::getOpName() const {
 /* ------------------------- ExpressionArrayElemAt -------------------------- */
 
 Value ExpressionArrayElemAt::evaluate(const Document& root) const {
-    const Value array = vpOperand[0]->evaluate(root);
     const Value indexArg = vpOperand[1]->evaluate(root);
+    long long i = indexArg.coerceToLong();
+
+    // If "vpOperand[0]" is an ExpressionFilter given index is greater than 0 call
```
cswanson310: why only do this if index is greater than 0?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -490,8 +490,15 @@ const char* ExpressionArray::getOpName() const {
...
+    // filterTillNthValue
+    const Value array = dynamic_cast<ExpressionFilter*>(vpOperand[0].get()) && (i >= 0)
+        ? dynamic_cast<ExpressionFilter*>(vpOperand[0].get())->filterTillNthValue(root, i)
+        : vpOperand[0]->evaluate(root);
+
 
     if (array.nullish() || indexArg.nullish()) {
         return Value(BSONNULL);
```
cswanson310: Oh, this check should also move before you call 'coerceToLong' or 'getLong'.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -503,16 +510,13 @@ Value ExpressionArrayElemAt::evaluate(const Document& root) const {
...
             str::stream() << getOpName() << "'s second argument must be a numeric value,"
-                          << " but is "
-                          << typeName(indexArg.getType()),
+                          << " but is " << typeName(indexArg.getType()),
             indexArg.numeric());
     uassert(28691,
             str::stream() << getOpName() << "'s second argument must be representable as"
-                          << " a 32-bit integer: "
-                          << indexArg.coerceToDouble(),
+                          << " a 32-bit integer: " << indexArg.coerceToDouble(),
```
cswanson310: If your version of clang-format is correct, the formatting here shouldn't change... Mine is version 3.8.0, what's yours?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2212,6 +2216,38 @@ Value ExpressionFilter::evaluate(const Document& root) const {
     return Value(std::move(output));
 }
 
+Value ExpressionFilter::filterTillNthValue(const Document& root, long long n) const {
```
cswanson310: Let's try to de-duplicate some of the code between ExpressionFilter::evaluate and ExpressionFilter::filterTillNthValue.

How about if the second argument here was a boost::optional<long long>, and ExpressionArrayElemAt called this function by passing a long long there (not boost::none), then the contents of ExpressionFilter::evaluate() can entirely be replaced with `filterTillNthValue(root, boost::none)`? Of course, you'd have to change line 2242 a bit, and the name might be a little misleading.

How does that sound?

KevinCybura:
The reason we need to check if the index is greater than zero is because it
can be negative for example if the index is -n we want the element n
indecies from the end of the array but we can't determine that unless we
know the end of the array so it should filter should evaluate as normal. A
similar issue arises with $slice but with $slice you can have and negative
start and end index.

You mention that I should check for $filter in the optimize stage of
$arrayElemAt so i guess I should do the same thing of $slice and then I
thought about the other ticket with the $indexOfArray (which could also
optimize for filter) and I though maybe I could write a class
OptimizeFilterExpression for when an expression is chained with $filter
similar for how you want me to write a class to optimize $indexOfArray with
a unordered_map in the next ticket. This is because you mention I should
make a private variable for $arrayElemAt called _firstArgumentIsFilter but
I would also need to the same for $slice and $indexOfArray so instead in
their optimize functions I can replace their FilterExpressions with
OptimizeFilterExpression and pass in vpOperand to the new Expression so
that could handle these cases. This could potentially clean up the code but
I'm not sure if should I try to do it this way it might overcomplicate
things if not ill just continue with what I was doing.

When running clang-format --version this is my out put

clang-format NPM version 1.2.2 at
/home/kevin-cybura/.nvm/versions/node/v8.9.4/lib/node_modules/clang-format/index.js
clang-format version 6.0.0 (trunk 318836) (llvm/trunk 318838)

Since im using VS code to format with clang-format it might be using the
NPM version.

PR #1225, opened 2018-03-16
SERVER-32191 DocumentSourceSkip can overflow a long long when combining with adjacent skips

KevinCybura:


src/mongo/db/pipeline/document_source_skip.cpp
```diff
@@ -84,11 +84,12 @@ Pipeline::SourceContainer::iterator DocumentSourceSkip::doOptimizeAt(
     invariant(*itr == this);
 
     auto nextSkip = dynamic_cast<DocumentSourceSkip*>((*std::next(itr)).get());
-
     if (nextSkip) {
-        _nToSkip += nextSkip->getSkip();
-        container->erase(std::next(itr));
-        return itr;
+        if (std::numeric_limits<long long>::max() - _nToSkip - nextSkip->getSkip() >= 0) {
```
cswanson310: Please add a comment here describing that we want to avoid overflowing '_nToSkip'. This only works because we know the skips are non-negative, right? If _nToSkip were negative, it looks like we'd overflow.

src/mongo/db/pipeline/document_source_skip_test.cpp
```diff
@@ -65,5 +65,28 @@ TEST_F(DocumentSourceSkipTest, ShouldPropagatePauses) {
     ASSERT_TRUE(skip->getNext().isEOF());
 }
 
+TEST_F(DocumentSourceSkipTest, SkipsChainedTogetherShouldNotOverFlowWhenOptimizing) {
+    // $skip should not optimize if overflow
```
cswanson310: 'if overflow' doesn't really make sense. How about 'if combining the two values of skips would overflow a long long'. Ditto for the comment below.

src/mongo/db/pipeline/document_source_skip_test.cpp
```diff
@@ -65,5 +65,28 @@ TEST_F(DocumentSourceSkipTest, ShouldPropagatePauses) {
...
+    auto skipShort = DocumentSourceSkip::create(getExpCtx(), 1);
+    auto skipLong = DocumentSourceSkip::create(getExpCtx(), std::numeric_limits<long long>::max());
```
cswanson310: Can you add a test to make sure that two $skips with `std::numeric_limits<long long>::max()` will not combine?

src/mongo/db/pipeline/document_source_skip_test.cpp
```diff
@@ -65,5 +65,28 @@ TEST_F(DocumentSourceSkipTest, ShouldPropagatePauses) {
...
+    Pipeline::SourceContainer overFlowContainer;
```
cswanson310: total nitpick, but if overflow is one word, this should be 'overflowContainer' not 'overFlowContainer'.

src/mongo/db/pipeline/document_source_skip.cpp
```diff
@@ -84,11 +84,13 @@ Pipeline::SourceContainer::iterator DocumentSourceSkip::doOptimizeAt(
     invariant(*itr == this);
 
     auto nextSkip = dynamic_cast<DocumentSourceSkip*>((*std::next(itr)).get());
-
     if (nextSkip) {
-        _nToSkip += nextSkip->getSkip();
-        container->erase(std::next(itr));
-        return itr;
+        // '_nToSkip' can potentially overflow causing it to become negative and skip nothing
```
cswanson310: All the changes look good here, thanks! Can you be sure to end all comments in periods: https://github.com/mongodb/mongo/wiki/Style-Guidelines#general-guidelines-for-commenting-code

Other than that this looks ready to go!

PR #1227, opened 2018-03-23
Scons build fixes for a smaller build

xnox:
I'm trying to perform a no-js engine build, and no mongo command. I've hit two issues:

serveronly target was underlinked, failing to find a reference to local_sharding_info.

core target referenced mongo executable, when it was not going to be built.

Below fixes both of the above issues on the v3.4 branch.

### DIFF
--- a/src/mongo/SConscript
+++ b/src/mongo/SConscript
@@ -513,10 +513,13 @@ env.Alias("tools", "#/" + add_exe("mongobridge"))
 installBinary( env, "mongod" )
 installBinary( env, "mongos" )
 
+core_alias_binaries = [ add_exe( "mongod" ), add_exe( "mongos" ) ]
+
 if shellEnv is not None:
     installBinary( shellEnv, "mongo" )
+    core_alias_binaries += [ add_exe( "mongo" ) ]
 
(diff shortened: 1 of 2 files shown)


Schubes:
Hi @xnox,

Thank you for the pull request. So we can review these changes, would you please be sure to sign the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement) and open a [SERVER ticket](https://jira.mongodb.org/browse/SERVER/) so we can track the fix?

Thanks again,
Kelsey

xnox:
Actually, this is a cherrypick request to fix https://jira.mongodb.org/browse/SERVER-29726 on the v3.4 branch

Fixed in v3.5 via https://github.com/mongodb/mongo/commit/b798e8d10ee53e1448a6ef0542d9d022037d2ea7

acmorrow:
Thanks for clarifying that this is a partial backport request. Given that, I think it would be best to split out the library dependency change into a separate PR, since that isn't really a backport.

acmorrow:
Also, in the other PR, can you provide some details on how you are building, what target fails to link, the link error, and your platform/toolchain?

PR #1229, opened 2018-03-23
SERVER-26751 $indexOfArray can optimize for constant array

KevinCybura:


src/mongo/db/pipeline/expression.cpp
```diff
@@ -26,7 +26,6 @@
  * it in the license file.
  */
 
-
```
cswanson310: Please avoid stray whitespace changes. It makes it harder to review and complicates the git blame unnecessarily. Not generally a big deal, but please avoid it going forward.

src/mongo/db/pipeline/expression.h
```diff
@@ -1218,13 +1219,19 @@ class ExpressionIn final : public ExpressionFixedArity<ExpressionIn, 2> {
...
 
 
-class ExpressionIndexOfArray final : public ExpressionRangedArity<ExpressionIndexOfArray, 2, 4> {
+class ExpressionIndexOfArray  : public ExpressionRangedArity<ExpressionIndexOfArray, 2, 4> {
 public:
     explicit ExpressionIndexOfArray(const boost::intrusive_ptr<ExpressionContext>& expCtx)
         : ExpressionRangedArity<ExpressionIndexOfArray, 2, 4>(expCtx) {}
 
-    Value evaluate(const Document& root) const final;
+    Value evaluate(const Document& root) const ;
```
cswanson310: extra space after 'const'.

src/mongo/db/pipeline/expression.h
```diff
@@ -1218,13 +1219,19 @@ class ExpressionIn final : public ExpressionFixedArity<ExpressionIn, 2> {
...
+    std::vector<Value> parseDeps(const Document& root,
```
cswanson310: Please add a comment here describing what this does. Also - parseDeps doesn't really give me a clue what this method does, so a more descriptive name like 'evaluateAndValidateArguments' might be more appropriate. Finally, if this is only used within IndexOfArray and IndexOfArray::Optimized, can you move it to the private or protected section?

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,45 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
                           });
 }
 
+TEST(ExpressionArrayTest, ExpressionArrayWithALlConstantValuesShouldOptimizeToExpressionConstant) {
```
cswanson310: Extra capital in WithALlConstant

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+        endIndex = std::min(arrayLength, static_cast<size_t>(endIndexArg.coerceToInt()));
+    }
+    deps.push_back(Value(endIndex));
+    return deps;
+}
+/**
+ *  This class handles the case where IndexOfArray is given an ExpressionConstant
+ *  instead of using a vector and searching through it we can use a unordered_map
+ *  for O(1) lookup time.
+ **/
```
cswanson310: extra `*` in close, just one here.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
         uassertIfNotIntegralAndNonNegative(endIndexArg, getOpName(), "ending index");
         // Don't let 'endIndex' exceed the length of the array.
-        endIndex = std::min(array.size(), static_cast<size_t>(endIndexArg.coerceToInt()));
+        endIndex = std::min(arrayLength, static_cast<size_t>(endIndexArg.coerceToInt()));
+    }
+    deps.push_back(Value(endIndex));
+    return deps;
+}
+/**
+ *  This class handles the case where IndexOfArray is given an ExpressionConstant
```
cswanson310: Only one space between the * and the words.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
         }
     }
 
-    return Value(-1);
+private:
+    const ValueUnorderedMap<int> _indexMap;
+};
+
+intrusive_ptr<Expression> ExpressionIndexOfArray::optimize() {
+    ExpressionNary::optimize();
```
cswanson310: Add a comment here indicating that this will optimize the arguments to this expression. In particular, if all arguments are constant, this should immediately evaluate to a constant. Can you add tests for that like you did for ExpressionArray?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+
+    // If the input arr is an ExpressionConstant we can optimize using a unordered_map instead of an
```
cswanson310: Don't abbreviate - it just makes it harder to understand, which is counter to the point of the comment. arr -> array.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+    // Array
+    if (dynamic_cast<ExpressionConstant*>(vpOperand[0].get())) {
+        ExpressionConstant* ec = dynamic_cast<ExpressionConstant*>(vpOperand[0].get());
```
cswanson310: Avoid short variable names like `ec`. Also, you can combine this with the if statement above:
`if (auto constantArray = dynamic_cast<ExpressionConstant*>(vpOperand[0].get())) {`

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+        const Value valueArray = ec->getValue();
+        if (valueArray.nullish()) {
+            return this;
```
cswanson310: We know what the answer is in this case, just return `ExpressionConstant::create(getExpressionContext(), Value(BSONNULL))` to save time. Also add a test case for this please :)

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2730,102 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+                               << typeName(valueArray.getType()),
+                valueArray.isArray());
+
+        auto arr = valueArray.getArray();
+
+        ValueUnorderedMap<int> indexMap =
+            getExpressionContext()->getValueComparator().makeUnorderedValueMap<int>();
+
+        for (int i = 0; i < int(arr.size()); i++) {
+            auto pair = std::make_pair(arr[i], i);
```
cswanson310: optional, re-write this as
`indexMap.emplace(arr[i], i);` or if that doesn't work, `indexMap[arr[i]] = i;`? I'm curious why you made a pair here.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2819,8 +2820,8 @@ intrusive_ptr<Expression> ExpressionIndexOfArray::optimize() {
             getExpressionContext()->getValueComparator().makeUnorderedValueMap<int>();
 
         for (int i = 0; i < int(arr.size()); i++) {
-            auto pair = std::make_pair(arr[i], i);
-            indexMap.insert(pair);
+            // auto pair = std::make_pair(arr[i], i);
```
cswanson310: Remove dead code.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2797,15 +2798,15 @@ class ExpressionIndexOfArray::Optimized : public ExpressionIndexOfArray {
...
+    // This is optimizes all arguments to this expression
     ExpressionNary::optimize();
 
-    // If the input arr is an ExpressionConstant we can optimize using a unordered_map instead of an
-    // Array
-    if (dynamic_cast<ExpressionConstant*>(vpOperand[0].get())) {
-        ExpressionConstant* ec = dynamic_cast<ExpressionConstant*>(vpOperand[0].get());
-        const Value valueArray = ec->getValue();
+    // If the input array is an ExpressionConstant we can optimize using a unordered_map instead of an
+    // array
```
cswanson310: End comments in periods - also lines 2801 and in the header.

src/mongo/db/pipeline/expression.h
```diff
@@ -1219,17 +1219,23 @@ class ExpressionIn final : public ExpressionFixedArity<ExpressionIn, 2> {
...
-                                  const ExpressionVector& operands,
-                                  size_t arrayLength) const;
+    Value evaluate(const Document& root) const;
     boost::intrusive_ptr<Expression> optimize() final;
     const char* getOpName() const final;
+
+protected:
+    // Evaluates and validates the value that is being searched for and if given the start index and
+    // the end index passed to ExpressionIndexOfArray returns arguments in the form of a
+    // vector<Value>
```
cswanson310:
```
/**
 * Use this syntax for comments documenting the behavior of a function or method.
 */
```

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2740,7 +2741,7 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
     return Value(-1);
 }
 
-vector<Value> ExpressionIndexOfArray::parseDeps(const Document& root,
+vector<Value> ExpressionIndexOfArray::evaluateAndValidateArguments(const Document& root,
                                                 const ExpressionVector& operands,
                                                 size_t arrayLength) const {
```
cswanson310: `arrayLength` is unused?

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2740,7 +2741,7 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
     std::vector<Value> deps;
```
cswanson310: `deps` is a strange name for this - how about `validatedArguments`?

src/mongo/db/pipeline/expression.h
cswanson310: Also - can we re-word this comment to make it more clear what the vector of values will be? How about something like
`When given 'operands' which correspond to the arguments to $indexOfArray, evaluates and validates the target value, starting index, and ending index arguments and returns their values in that order. The starting index and ending index are optional, so the returned vector will have a length between 1 and 3. Throws a UserException if the values are found to be invalid in some way, e.g. if the indexes are not numbers.`

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2247,6 +2247,45 @@ TEST(ExpressionArrayTest, ExpressionArrayShouldOptimizeSubExpressionToExpression
     ASSERT_TRUE(constantExpression);
 }
 
+TEST(ExpressionIndexOfArray, ExpressionIndexOfArrayShouldOptimizeArguments) {
+    intrusive_ptr<ExpressionContextForTest> expCtx(new ExpressionContextForTest());
+
+    auto expIndexOfArray = Expression::parseExpression(
+        expCtx,
+        BSON("$indexOfArray" << BSON_ARRAY(BSON_ARRAY(BSON("$add" << BSON_ARRAY(1 << 1)) << 1 << 1)
+                                           // Start index
```
cswanson310: I don't think this is the start index, I think this is the value you're searching for?

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2247,6 +2247,45 @@ TEST(ExpressionArrayTest, ExpressionArrayShouldOptimizeSubExpressionToExpression
...
+                                           << BSON("$add" << BSON_ARRAY(1 << 1))
+                                           // End index
+                                           << BSON("$add" << BSON_ARRAY(1 << 1)))),
+        expCtx->variablesParseState);
+    auto argsOptimizedToConstants = expIndexOfArray->optimize();
+    auto shouldBeNary = dynamic_cast<ExpressionNary*>(argsOptimizedToConstants.get());
+    ASSERT_TRUE(shouldBeNary);
+
+    auto optimizedArgs = shouldBeNary->getOperandList();
+    // All arguments should be ExpressionConstants.
```
cswanson310: Oh, interesting - I would have expected it to fully optimize to a constant in this case... I wonder why it didn't? Can you make sure it does?

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2247,6 +2247,45 @@ TEST(ExpressionArrayTest, ExpressionArrayShouldOptimizeSubExpressionToExpression
...
+    auto isExpIndexOfArray = dynamic_cast<ExpressionIndexOfArray*>(expIndex.get());
+    ASSERT_TRUE(isExpIndexOfArray);
+
+    auto nullishValueOptimizedToExpConstant = isExpIndexOfArray->optimize();
+    auto shouldBeExpressionConstant = dynamic_cast<ExpressionConstant*>(nullishValueOptimizedToExpConstant.get());
+    ASSERT_TRUE(shouldBeExpressionConstant);
+    // Nullish input array should become a Value(BSONNULL).
+    ASSERT_VALUE_EQ(Value(BSONNULL), shouldBeExpressionConstant->getValue());
+}
+
```
cswanson310: Since we aren't adding any coverage of actually evaluating $indexOfArray when it can be optimized, can you either add some tests here which exercise the new code (create an `IndexOfArray` with a constant array and various other arguments, making sure to call `optimize()` before `evaluate()`), or find some that are already written to prove that it will be exercised?

src/mongo/db/pipeline/expression_test.cpp
KevinCybura: I think this is because we are optimizing an ExpresssionIndexOfArray not evaluating so it should still be an ExpressionIndexOfArray.

Im casting it to an ExpressionNary because I need to be able to call getOperandList() to determine if the arguments optimized. I think this is where the confusion is to clearify I will cast it to an ExpressionIndexOfArray

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,101 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
         }
     }
 
-    return Value(-1);
+private:
+    const ValueUnorderedMap<int> _indexMap;
+};
+
+intrusive_ptr<Expression> ExpressionIndexOfArray::optimize() {
+    // This is optimize all arguments to this expression.
```
cswanson310: This *will* optimize ...

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,101 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+                               << typeName(valueArray.getType()),
+                valueArray.isArray());
+
+        auto arr = valueArray.getArray();
+
+        ValueUnorderedMap<int> indexMap =
+            getExpressionContext()->getValueComparator().makeUnorderedValueMap<int>();
+
+        for (int i = 0; i < int(arr.size()); i++) {
+            indexMap.emplace(arr[i], i);
```
cswanson310: As discussed, we should not add a value if it already occurred so that we only ever return the first occurrence.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,101 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
-    for (size_t i = startIndex; i < endIndex; i++) {
-        if (getExpressionContext()->getValueComparator().evaluate(array[i] == searchItem)) {
-            return Value(static_cast<int>(i));
+    virtual Value evaluate(const Document& root) const {
+        std::vector<Value> operands = evaluateAndValidateArguments(root, vpOperand, _indexMap.size());
+        auto index = _indexMap.find(operands[0]);
+        auto comparator = getExpressionContext()->getValueComparator();
+        Value vIndex = Value(index->second);
+        if (index != _indexMap.end() && comparator.evaluate(Value(index->second) >= operands[1]) &&
+            comparator.evaluate(Value(index->second) < operands[2])) {
```
cswanson310: Oh this is pretty awkward. You have to keep using `comparator.evaluate(...)` because `operands[1]` and `operands[2]` are of type `Value`. In reality, we know they will be of type `int`, right? Or even `size_t`?

Can we change the signature of `evaluateAndValidateArguments` to return a struct, something like
```
struct Arguments {
    Arguments(Value targetOfSearch, int startIndex, boost::optional<int> endIndex) : 
        targetOfSearch(targetOfSearch), startIndex(startIndex), endIndex(endIndex) {}

    Value targetOfSearch;
    int startIndex;
    boost::optional<int> endIndex;
};
```

It might be more complicated in the parsing code, but I think it would make this much easier to read.

cswanson310: Also, how do you know that operands[2] is valid? Couldn't the user have left off the end index?

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,132 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+    auto notOptimized = expressionArrNotConstant->optimize();
+    auto notExprConstant = dynamic_cast<ExpressionConstant*>(notOptimized.get());
+    ASSERT_FALSE(notExprConstant);
+}
+
+TEST(ExpressionArrayTest, ExpressionArrayShouldOptimizeSubExpressionToExpressionConstant) {
+    intrusive_ptr<ExpressionContextForTest> expCtx(new ExpressionContextForTest());
+    VariablesParseState vps = expCtx->variablesParseState;
+
+
```
cswanson310: delete extra newline.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,132 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+                                           << BSON("$add" << BSON_ARRAY(1 << 1))
+                                           // End index.
+                                           << BSON("$add" << BSON_ARRAY(1 << 1)))),
+        expCtx->variablesParseState);
+    auto argsOptimizedToConstants = expIndexOfArray->optimize();
+    auto shouldBeIndexOfArray = dynamic_cast<ExpressionIndexOfArray*>(argsOptimizedToConstants.get());
+    ASSERT_TRUE(shouldBeIndexOfArray);
+
+    auto optimizedArgs = shouldBeIndexOfArray->getOperandList();
+    // All arguments should be ExpressionConstants.
```
cswanson310: I think if you change the optimize implementation to return what `ExpressionNary::optimize()` returns then the whole expression should evaluate to a constant.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,132 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+
+    auto nullishValueOptimizedToExpConstant = isExpIndexOfArray->optimize();
+    auto shouldBeExpressionConstant = dynamic_cast<ExpressionConstant*>(nullishValueOptimizedToExpConstant.get());
+    ASSERT_TRUE(shouldBeExpressionConstant);
+    // Nullish input array should become a Value(BSONNULL).
+    ASSERT_VALUE_EQ(Value(BSONNULL), shouldBeExpressionConstant->getValue());
+}
+
+TEST(ExpressionIndexOfArray,
+     OptimizedExpressionIndexOfArrayWithConstantArgumentsShouldEvaluateProperly) {
```
cswanson310: This is a good test, but because you're using all constant values, I think you will hit the case where all arguments are constant instead of when just the array is constant. If you change the search values (e.g. `2` and `10`) to be something like `$x`, then use `Document{{"x", 2}}` as the argument to evaluate(), you'll be testing what you want to test.

cswanson310: Same thing for the test below.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,132 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+    auto IndexNotFoundInRange =
+        Expression::parseExpression(expCtx,
+                                    // Search for 0 between 3(inclusive) and 5 (exclusive).
+                                    fromjson("{ $indexOfArray : [ [0, 1, 2, 3, 4, 5] , 0, 3, 5] }"),
+                                    expCtx->variablesParseState);
+    auto optimizedIndexNotFoundInRange = IndexNotFoundInRange->optimize();
+    // Should evaluate to -1 if not found in range.
+    ASSERT_VALUE_EQ(Value(-1), optimizedIndexNotFoundInRange->evaluate(Document()));
+}
+
```
cswanson310: Can you add a test where there are duplicate values in the array? Such a test should have caught the bug you mentioned with inserting into the unordered_map twice.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,100 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
-        if (getExpressionContext()->getValueComparator().evaluate(array[i] == searchItem)) {
-            return Value(static_cast<int>(i));
+    virtual Value evaluate(const Document& root) const {
+        auto args = evaluateAndValidateArguments(root, vpOperand, _indexMap.size());
+        auto index = _indexMap.find(args.targetOfSearch);
+        if (index != _indexMap.end() && index->second >= args.startIndex &&
+            index->second < args.endIndex) {
+            return Value(index->second);
+        } else {
+            // If the item we are searching is not found or if not found in given range return -1.
```
cswanson310: Hmm I actually think this won't work. Consider the following:
```
{$indexOfArray: [[0, 1, "a", 0, 1, 1, "a"], "a", 3, 8]}
```
I think that the map here will be populated with `0-> 0, 1-> 1, "a" -> 2`, but then it has lost the information about where the second or third occurrences of those values are. I think we need to map a value to a `vector<int>` of all the positions so that we can rule out those which are out of range.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+ExpressionIndexOfArray::Arguments ExpressionIndexOfArray::evaluateAndValidateArguments(const Document& root,
+                                                const ExpressionVector& operands,
+                                                size_t arrayLength) const {
+    
+    int startIndex = 0;
+    if (operands.size() > 2) {
+        Value startIndexArg = operands[2]->evaluate(root);
         uassertIfNotIntegralAndNonNegative(startIndexArg, getOpName(), "starting index");
-        startIndex = static_cast<size_t>(startIndexArg.coerceToInt());
+        startIndex = static_cast<int>(startIndexArg.coerceToInt());
```
cswanson310: I don't think you need a static_cast here.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
     }
+
```
cswanson310: extra newline.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
-    size_t endIndex = array.size();
-    if (vpOperand.size() > 3) {
-        Value endIndexArg = vpOperand[3]->evaluate(root);
+    int endIndex = arrayLength;
+    if (operands.size() > 3) {
+        Value endIndexArg = operands[3]->evaluate(root);
         uassertIfNotIntegralAndNonNegative(endIndexArg, getOpName(), "ending index");
         // Don't let 'endIndex' exceed the length of the array.
-        endIndex = std::min(array.size(), static_cast<size_t>(endIndexArg.coerceToInt()));
+        endIndex = std::min(arrayLength, static_cast<size_t>(endIndexArg.coerceToInt()));
```
cswanson310: I don't think you need a static_cast here.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
     }
 
-    for (size_t i = startIndex; i < endIndex; i++) {
-        if (getExpressionContext()->getValueComparator().evaluate(array[i] == searchItem)) {
-            return Value(static_cast<int>(i));
+    return ExpressionIndexOfArray::Arguments(vpOperand[1]->evaluate(root), startIndex, endIndex);
```
cswanson310: I think the complier knows the return types, so you could just say
`return {vpOperand[1]->evaluate(root), startIndex, endIndex}`

Or, if you prefer to keep the name, this method is defined on ExpressionIndexOfArray, so you don't need to include the `ExpressionIndexOfArray::` prefix here.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+              const ValueUnorderedMap<vector<int>>& indexMap,
+              const ExpressionVector& operands)
+        : ExpressionIndexOfArray(expCtx), _indexMap(std::move(indexMap)) {
+        vpOperand = operands;
+    }
+
+    virtual Value evaluate(const Document& root) const {
+        auto args = evaluateAndValidateArguments(root, vpOperand, _indexMap.size());
+        auto indexVec = _indexMap.find(args.targetOfSearch);
+        if (indexVec != _indexMap.end()) {
```
cswanson310: I think this would be easier to read as
```
if (indexVec == _indexMap.end()) {
    return Value(-1);
}

// Search through the vector of indices for the first index in our range.
...  // Un-indented code.
```

cswanson310: Alternatively, you could just leave off the else, and move the `return Value(-1);` to always be the last thing that happens, whether you get into the if statement or not.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+            // Search through the vector of indecies for first index in our range.
+            for (auto index : indexVec->second) {
+                if (index >= args.startIndex && index < args.endIndex) {
+                    return Value(index);
+                }
+            }
+            // the value we are searching for exists but is not in our range.
```
cswanson310: Capital `The`.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+            return Value(-1);
+        } else {
+            return Value(-1);
         }
     }
 
-    return Value(-1);
+private:
+    const ValueUnorderedMap<vector<int>> _indexMap;
```
cswanson310: Can you add a comment here like the following?
```
// Maps the values in the array to the positions at which they occur. We need to remember the positions 
// so that we can verify they are in the appropriate range.
```

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+        }
+        uassert(50749,
+                str::stream() << "First operand of $indexOfArray must be an array. First "
+                              << "argument is of type: "
+                               << typeName(valueArray.getType()),
+                valueArray.isArray());
+
+        auto arr = valueArray.getArray();
+        // To handle the case of duplicate values the values need to map to a vector of indecies.
+        ValueUnorderedMap<vector<int>> indexMap =
```
cswanson310: You should use 'auto' here, since the type is declared later on this line when you create it.

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+            getExpressionContext()->getValueComparator().makeUnorderedValueMap<vector<int>>();
+
+        for (int i = 0; i < int(arr.size()); i++) {
+            if(indexMap.find(arr[i]) == indexMap.end()){
+               indexMap.emplace(arr[i], vector<int>());
+               indexMap[arr[i]].push_back(i);
```
cswanson310: Since you're doing this in both the if and the else, I'd recommend re-writing lines 2826-2832 as
```
if (indexMap.find(arr[i]) == indexMap.end()) {
    indexMap.emplace(arr[i], vector<int>());
}
indexMap[arr[i]].push_back(i);
```

src/mongo/db/pipeline/expression.cpp
```diff
@@ -2716,31 +2731,110 @@ Value ExpressionIndexOfArray::evaluate(const Document& root) const {
...
+        for (int i = 0; i < int(arr.size()); i++) {
+            if(indexMap.find(arr[i]) == indexMap.end()){
+               indexMap.emplace(arr[i], vector<int>());
+               indexMap[arr[i]].push_back(i); 
+            }
+            else {
+                indexMap[arr[i]].push_back(i);
+            }
+        }
+        return intrusive_ptr<Expression>(
```
cswanson310: I think you can leave off the `intrusive_ptr<Expression>` and the compiler will figure it out.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,153 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+TEST(ExpressionIndexOfArray,
+     OptimizedExpressionIndexOfArrayWithConstantArgumentsShouldEvaluateProperly) {
+
+    intrusive_ptr<ExpressionContextForTest> expCtx(new ExpressionContextForTest());
+
+    auto expIndexOfArray =
+        Expression::parseExpression(expCtx,
+                                    // Search for 2.
```
cswanson310: Search for '$x' you mean?

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,153 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+                                    fromjson("{ $indexOfArray : [ [0, 1, 2, 3, 4, 5] , '$x'] }"),
+                                    expCtx->variablesParseState);
+    auto optimizedIndexOfArray = expIndexOfArray->optimize();
+    ASSERT_VALUE_EQ(Value(1), optimizedIndexOfArray->evaluate(Document{{"x", 1}}));
```
cswanson310: Might as well add a bunch of these:
```
ASSERT_VALUE_EQ(Value(0), optimizedIndexOfArray->evaluate(Document{{"x", 0}}));
ASSERT_VALUE_EQ(Value(1), optimizedIndexOfArray->evaluate(Document{{"x", 1}}));
ASSERT_VALUE_EQ(Value(2), optimizedIndexOfArray->evaluate(Document{{"x", 2}}));
ASSERT_VALUE_EQ(Value(3), optimizedIndexOfArray->evaluate(Document{{"x", 3}}));
ASSERT_VALUE_EQ(Value(4), optimizedIndexOfArray->evaluate(Document{{"x", 4}}));
ASSERT_VALUE_EQ(Value(5), optimizedIndexOfArray->evaluate(Document{{"x", 5}}));
```

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,153 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+    auto IndexNotFound =
```
cswanson310: I don't think you need to re-parse or optimize, you can use the same one.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,153 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+
+    auto expIndexOfArray = Expression::parseExpression(
+        expCtx,
+        // Search for 4 between 3 and 5.
+        fromjson("{ $indexOfArray : [ [0, 1, 2, 3, 4, 5] , '$x', 3, 5] }"),
+        expCtx->variablesParseState);
+    auto optimizedIndexOfArray = expIndexOfArray->optimize();
+    ASSERT_VALUE_EQ(Value(4), optimizedIndexOfArray->evaluate(Document{{"x", 4}}));
+
+    auto IndexNotFoundInRange = Expression::parseExpression(
```
cswanson310: I think you can re-use `optimizedIndexOfArray`

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,153 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+    auto expIndexOfArrayWithDuplicateValues =
+        Expression::parseExpression(expCtx,
+                                    // Search for 4 between 3 and 5.
+                                    fromjson("{ $indexOfArray : [ [0, 1, 2, 2, 3, 4, 5] , '$x'] }"),
+                                    expCtx->variablesParseState);
+    auto optimizedIndexOfArrayWithDuplicateValues = expIndexOfArrayWithDuplicateValues->optimize();
+    ASSERT_VALUE_EQ(Value(2),
+                    optimizedIndexOfArrayWithDuplicateValues->evaluate(Document{{"x", 2}}));
+
+    auto expIndexInRangeWithhDuplicateValues = Expression::parseExpression(
```
cswanson310: Again, I think you can re-use the expression from above.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2208,6 +2208,153 @@ TEST(ExpressionPowTest, NegativeOneRaisedToNegativeOddExponentShouldOutPutNegati
...
+        // Search for 2 between 4 and 6.
+        fromjson("{ $indexOfArray : [ [0, 1, 2, 2, 2, 2, 4, 5] , '$x', 4, 6] }"),
+        expCtx->variablesParseState);
+    auto optimizedIndexInRangeWithDuplcateValues = expIndexInRangeWithhDuplicateValues->optimize();
+    // Should evaluate to 4.
+    ASSERT_VALUE_EQ(Value(4),
+                    optimizedIndexInRangeWithDuplcateValues->evaluate(Document{{"x", 2}}));
+}
+
+
```
cswanson310: Extra newline.

PR #1230, opened 2018-03-27
Add missing error code in central list of error codes

fparrel:
Just got an error code that is not on the main list. I think it should be added.

### DIFF
--- a/src/mongo/base/error_codes.err
+++ b/src/mongo/base/error_codes.err
@@ -269,6 +269,7 @@ error_code("ShardKeyTooBig", 13334);
 error_code("StaleConfig", 13388, extra="StaleConfigInfo");
 error_code("DatabaseDifferCase", 13297);
 error_code("OBSOLETE_PrepareConfigsFailed", 13104);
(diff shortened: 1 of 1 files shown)


ramonfm:
Thanks for taking the time to create a pull request @fparrel. Please take a look at the [Contributing to MongoDB](https://github.com/mongodb/mongo/wiki) section of the wiki for some prerequisites.

Note also that when adding a new name to error_codes.err this error code should be referenced by name when used.

Thanks,
Ramón.

fparrel:
Thanks Ramón, contributor's agreement signed.
I don't have reference to a JIRA ticket but to an internal support ticket: https://support.mongodb.com/case/00487973
I preparing a new commit with replacement of reference by name instead of literal integer, and including other unlisted error codes.
Regards
Frédéric

jseyster:
Frédéric,

Thanks for this contribution! We recently changed the error codes file to be a YAML document, so I went ahead and added your suggested error code to that as part of SERVER-42659. The new error code is slightly more general: it also applies to $group stages that run out of memory. I hope that's helpful, and thanks again for getting this improvement started.

Justin

PR #1231, opened 2018-03-27
SERVER-32721 Explain output should indicate when a backup plan is used

d-d-ddl:


jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,28 @@
+// Test that the explain will use backup plan if the original winning plan ran out of memory in the
+// "executionStats" mode
+// This test was designed to reproduce SERVER-32721"
+(function() {
+    "use strict";
+
+    db.foo.drop() let bulk = db.foo.initializeUnorderedBulkOp();
```
cswanson310: Looks like you're missing a semi-colon after `.drop()` here.

jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,28 @@
...
+    for (let i = 0; i < 100000; ++i) {
```
cswanson310: This test is great, but we tend to favor tests that don't need as much data to be inserted, since they will execute faster. In this case you configure the maxBlockingSortBytes, so do you need this many documents?

jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,28 @@
...
+        bulk.insert({_id: i, x: i, y: i});
+    }
+
+    bulk.execute();
+    db.foo
+        .ensureIndex({x: 1})
```
cswanson310: Missing semi-colon, also on line 23 and 27.

src/mongo/db/exec/multi_plan.cpp
```diff
@@ -245,6 +247,7 @@ Status MultiPlanStage::pickBestPlan(PlanYieldPolicy* yieldPolicy) {
     _backupPlanIdx = kNoSuchPlan;
     if (bestSolution->hasBlockingStage && (0 == alreadyProduced.size())) {
         LOG(5) << "Winner has blocking stage, looking for backup plan...";
+
```
cswanson310: Not a big deal, but try not to include unnecessary changes like this in your patch. This one isn't particularly offensive, but I'd still recommend reverting this change.

src/mongo/db/exec/multi_plan.h
```diff
@@ -132,6 +132,11 @@ class MultiPlanStage final : public PlanStage {
     /** Return the index of the best plan chosen, for testing */
     int bestPlanIdx() const;
 
+    /** Return the index of the backup plan chosen, for testing */
```
cswanson310: I know you just copied this from above, but in new code we tend to always prefer
```
/**
 * This kind of syntax.
 */
```
instead of 
```
/** This kind of syntax */
```

Also, these new methods are used for explain output, not just for testing. I'm pretty sure `bestPlanIdx()` is as well, so you can remove that part of it.

Please update these, and any others in this file while you're at it.

src/mongo/db/exec/multi_plan.h
```diff
@@ -198,10 +203,14 @@ class MultiPlanStage final : public PlanStage {
     // uses -1 / kNoSuchPlan when best plan is not (yet) known
     int _bestPlanIdx;
 
-    // index into _candidates, of the backup plan for sort
+    // index into _candidates, of the backup of the plan competition
```
cswanson310: This still doesn't read particularly well. How about re-phrasing this to
` The index within '_candidates' of the non-blocking backup plan which can be used if a blocking plan fails. This is set to 'kNoSuchPlan' if there is no backup plan, or when it is not yet known.`? I'd recommend a similar re-wording for the comment on '_originalWinningPlanIndex' below, with a mention that this is used to remember which plan one in the case that a backup plan took over.

src/mongo/db/query/explain.cpp
```diff
@@ -289,6 +289,15 @@ unique_ptr<PlanStageStats> getWinningPlanStatsTree(const PlanExecutor* exec) {
                : std::move(exec->getRootStage()->getStats());
 }
 
+
+/**
+ * Get PlanExecutor's original winning plan stats tree.
```
cswanson310: Please elaborate here that this will only be different from the actual winning plan if a non-blocking backup plan was used. How about `Returns the root of the original winning plan used by 'exec'. This may be different than the final winning plan in the case that the MultiPlanStage selected a blocking plan which failed, and fell back to a non-blocking plan instead. If there is no MultiPlanStage in the tree, returns the root stage of 'exec'.`

src/mongo/db/query/explain.cpp
```diff
@@ -644,6 +653,16 @@ void Explain::generatePlannerInfo(PlanExecutor* exec,
     BSONObjBuilder plannerBob(out->subobjStart("queryPlanner"));
 
     plannerBob.append("plannerVersion", QueryPlanner::kPlannerVersion);
+
+    const auto mps = getMultiPlanStage(exec->getRootStage());
+    int originaWinningPlanIdx = static_cast<size_t>(mps->originalWinningPlanIdx());
+
+    if (originaWinningPlanIdx > -1) {
```
cswanson310: It looks like you never use the actual index, so how about changing line 658 to be
`const bool usedBackupPlan = (static_cast<size_t>(mps->originalWinningPlanIdx()) > -1);`? That way you can collapse lines 660-664 into one line.

jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,28 @@
...
+
+        // Configure log level and lower the sort bytes limit.
+        db.setLogLevel(5, "query");
+    db.adminCommand({setParameter: 1, internalQueryExecMaxBlockingSortBytes: 100});
+
+    // This query will not use the backup plan, hence it generates only two stages: winningPlan and
+    // rejectedPlans
+    assert
+        .commandWorked(db.foo.find({x: {$gte: 90}}).sort({_id: 1}).explain(true))
```
cswanson310: Your comment here mentions what we expect, but you just assert that the command "worked", which just means it returned a document with an `ok` field with a truthy value (usually `1`). 

Check out the helper functions defined here: https://github.com/mongodb/mongo/blob/r3.7.2/jstests/libs/analyze_plan.js

I think you should add a helper there which gets the 'originalWinningPlan', and assert that it is null here, and non-null below.

cswanson310 (changes requested):
Lots of little changes requested, but overall this looks really good! The code is looking very clean and organized. You happened to copy some older code in the MultiPlanStage, so I understand some of the comments may be a surprise - sorry about that. We have different style standards today and the code you copied hadn't been updated in a while. I hope they most of them should be pretty easy to address.

Thanks for doing this!

jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,30 @@
+// Test that the explain will use backup plan if the original winning plan ran out of memory in the
+// "executionStats" mode
+// This test was designed to reproduce SERVER-32721"
+
+load("jstests/libs/analyze_plan.js");
```
cswanson310: Please put the contents of this test in an immediately-invoked-function-expression (https://en.wikipedia.org/wiki/Immediately-invoked_function_expression)

jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,30 @@
...
+
+"use strict";
+
+db.foo.drop();
```
cswanson310: Use a more unique collection name, how about 
```
const coll = db.explain_backup_plan;
coll.drop();

// Replace usages of 'db.foo' with 'coll' below.
```

jstests/core/explain_use_backup_plan.js
```diff
@@ -0,0 +1,30 @@
...
+
+for (let i = 0; i < 100000; ++i) {
+    bulk.insert({_id: i, x: i, y: i});
+}
+
+bulk.execute();
+db.foo.ensureIndex({x: 1});
+
+// Configure log level and lower the sort bytes limit.
+db.setLogLevel(5, "query");
```
cswanson310: You don't need to set the log level for the test, that was just for your personal info. Please remove this line so the test doesn't generate lots of logs. This will also change the setting for subsequent tests, so it would get quite noisy.

src/mongo/db/exec/multi_plan.h
cswanson310: It looks like this still hasn't been addressed?

src/mongo/db/exec/multi_plan.h
```diff
@@ -198,10 +210,14 @@ class MultiPlanStage final : public PlanStage {
...
     // uses -1 / kNoSuchPlan when best plan is not (yet) known
     int _backupPlanIdx;
 
+    // Index into _candidates, of the original winner of the plan which can be used if a blocking plan .fails
```
cswanson310: Use single-quotes to refer to the variable name: `Index into '_candidates', ...`

cswanson310: Also, it looks like you don't need the comma here, and you should move the "." before 'fails' to be after it.

src/mongo/db/query/explain.cpp
```diff
@@ -62,6 +62,7 @@ using std::string;
 using std::unique_ptr;
 using std::vector;
 
+bool backupPlanUsed = false;
```
cswanson310: This should be passed as a parameter to the function. These helpers can be called by multiple threads, which would induce a race condition writing to the variable here.

src/mongo/db/query/explain.cpp
```diff
@@ -280,6 +281,30 @@ std::vector<std::unique_ptr<PlanStageStats>> getRejectedPlansTrialStats(PlanExec
     return res;
 }
 
+/**
+ * Gather the PlanStages when a backup plan is used
```
cswanson310: This comment is confusing. What do we do when a backup plan is not used? 

I'd actually recommend removing this method, and instead adding a 'bool backupPlanUsed' parameter to 'getRejectedPlansTrialStats'.

cswanson310: If you do that, you should then extend the comment on that method to include something like "If a backup plan was used, include the stats of the new winning plan instead of the original winning plan. The original winning plan is expected to already be known by the caller, since it needs to be snapshotted and stored before execution completes."

src/mongo/db/query/explain.cpp
```diff
@@ -289,6 +314,18 @@ unique_ptr<PlanStageStats> getWinningPlanStatsTree(const PlanExecutor* exec) {
                : std::move(exec->getRootStage()->getStats());
 }
 
+
+/**
+ * Returns the root of the roginal winning plan used by 'exec'.
```
cswanson310: original, not 'roginal'

src/mongo/db/query/explain.cpp
```diff
@@ -289,6 +314,18 @@ unique_ptr<PlanStageStats> getWinningPlanStatsTree(const PlanExecutor* exec) {
...
+ * This might be different than the final winning plan in the case that the MultiPlanStage selected a 
+ * blocking plan which failed, and fell back to a non-blocking plan instead.
+ * If there is no MultiPlanStage in the tree, returns the root stage of 'exec'
```
cswanson310: Finish with a period.

src/mongo/db/query/explain.cpp
```diff
@@ -855,8 +913,9 @@ void Explain::explainStages(PlanExecutor* exec,
                             const Collection* collection,
                             ExplainOptions::Verbosity verbosity,
                             BSONObjBuilder* out) {
+    
     auto winningPlanTrialStats = Explain::getWinningPlanTrialStats(exec);
-
+
```
cswanson310: Looks like a stray whitespace change (and also on line 916), can you please undo this?

cswanson310 (commented):
A couple comments, mostly about the wording. I do think there is one serious problem with the global variable.

kelly-cs:
I'm leaving this open since SERVER-32721 is still unresolved - could you double check this PR is up to date with master?

Addtionally, if you're still looking to merge this, can you:

I checked in on this issue to see if it's something we're going to look at.

PR #1232, opened 2018-04-02
SERVER-34250 Update Tools builders in Evergreen

xdg:
This commit changes how tools are built to match the upstream
mongo-tools repo.  With the exception of the ARM architecture, all tools
projects now build with Go 1.8.

### DIFF
--- a/etc/evergreen.yml
+++ b/etc/evergreen.yml
@@ -2489,7 +2489,7 @@ tasks:
     - func: "update bypass expansions"
     - func: "get buildnumber"
     - func: "set up credentials"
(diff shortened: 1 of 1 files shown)


vielmetti:
@xdg do you refer to 32-bit ARM, or 64-bit arm64, in your description of build problems with Arm? Is there a log of failures to build on Arm somewhere?

xdg:
@vielmetti ARM64.  I was able to manually test with Go 1.9.something on ARM64, so once our build infrastructure is updated to offer Go 1.9.x or Go 1.10.x, everything should work once more.

PR #1233, opened 2018-04-03
if we are forcing fork, then systemd unit needs to enforce it

gbolo:
commit 3607059 introduced a default fork option, but it has not
been enforced in the command line. Since fork is enforced in the
systemd unit and ignores the configuration file, if the config
file has fork disabled (default value) then the unit will not
start properly. This is problematic for situations where a config
file is being managed by an external system like ansible


kelly-cs:
mongod.service has been updated since then - this has changed to my knowledge. Closing this for now

PR #1235, opened 2018-04-17
set umask=0027 in systemd service unit

Nirkus:
Set UMask=0027 in debian/mongod.service so that database files are created with no permissions for "other" users.


kelly-cs:
mongod.service has changed since, closing this unless there are new concerns

PR #1236, opened 2018-04-26
SERVER-25957 Optimize $filter + $arrayElemAt to avoid scanning entire array

KevinCybura:


src/mongo/db/pipeline/expression.cpp
```diff
@@ -3900,6 +3935,65 @@ Value ExpressionSlice::evaluate(const Document& root) const {
...
+                            thirdVal.integral());
+
+                    uassert(50801,
+                            str::stream() << "Third argument to $slice must be positive: "
+                                          << thirdVal.coerceToInt(),
+                            thirdVal.coerceToInt() > 0);
+
+                    int arg3 = thirdVal.coerceToInt();
+                    if (arg2 >= 0) {
+                        // If ExpressionSlice is given three arguments set limit to 'firstArg' +
```
cswanson310: Can you elaborate a little bit more here? Maybe something like `If we have three arguments, then we can stop the $filter once we have seen 'firstArg' + 'secondArg' + 1. We need to skip 'firstArg' elements, then include 'secondArg' things after that. For example, if we want to slice starting from 2 with length 3, we skip the first 2, then include 3 after that.`

Wait, I think we don't need the `+1` here? Is that right?

src/mongo/db/pipeline/expression.h
```diff
@@ -1174,6 +1176,9 @@ class ExpressionFilter final : public Expression {
     boost::intrusive_ptr<Expression> _input;
     // The expression determining whether each element should be present in the result array.
     boost::intrusive_ptr<Expression> _filter;
+    // For expression ArrayElemAt and Slice we can optimize filter by setting a limit to end the
```
cswanson310: I'd make it clearer that this optimization applies when $filter is being used _as an argument_ to either $slice or $arrayElemAt. Also, feel free to use the $slice name instead of ExpressionSlice here.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2894,6 +2894,92 @@ TEST(ExpressionObjectOptimizations,
 
 }  // namespace Object
 
+TEST(ExpressionFilter, ExpressionFilterWithASetLimitShouldReturnAArrayNoGreaterThanTheLimit) {
```
cswanson310: Use `AnArray` instead of `AArray`.

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2894,6 +2894,92 @@ TEST(ExpressionObjectOptimizations,
...
+    // Returns an array with all values greater than 3.
+    auto filterSpec = BSON(
+        "$filter" << BSON("input" << BSON_ARRAY(1 << 2 << 3 << 4 << 5 << 6 << 7 << 8 << 9) << "as"
+                                  << "arr"
+                                  << "cond" << BSON("$gt" << BSON_ARRAY("$$arr" << 3))));
+
+    auto arrayElemAtSpec = BSON("$arrayElemAt" << BSON_ARRAY(filterSpec << 2));
+
+    auto expArrayElemAt = ExpressionArrayElemAt::parse(expCtx, arrayElemAtSpec.firstElement(), vps);
+    auto xpArrayElemAt = dynamic_cast<ExpressionArrayElemAt*>(expArrayElemAt.get())->optimize();
```
cswanson310: Use a better variable name, maybe 'optimized'?

src/mongo/db/pipeline/expression_test.cpp
```diff
@@ -2894,6 +2894,92 @@ TEST(ExpressionObjectOptimizations,
...
+    auto val = xpArrayElemAt->evaluate(Document());
+    ASSERT_VALUE_EQ(val, Value(6));
+
+    auto eA = ExpressionArrayElemAt::parse(expCtx, BSON("$arrayElemAt" << BSON_ARRAY(BSON_ARRAY(1 << 2 <<3 << 4 <<5) << 1)).firstElement(), vps);
```
cswanson310: You can re-use the variable names from before, I think that'd be clearer than two-letter variable names.

cswanson310: Same with below. Alternatively, you could separate each one into it's own test.

kelly-cs:
Hey @KevinCybura!

I was checking through old PR's and saw that SERVER-25957 is actually still open, as this wasn't merged. I am poking that ticket to see where it's at.

If you are still looking to get these changes in, can you please resolve the merge conflicts and get this caught up with master? I saw you already signed our contributor's agreement. I know it's been a few years, so it's okay if that's not on the table. (I'm trying to clean up our pull request queue right now)

PR #1240, opened 2018-05-12
fixbug SERVER-31687: Missing minvalid in sync source cause oplog full table scan all the time

zydcom:
See details in [SERVER-31687](https://jira.mongodb.org/browse/SERVER-31687)

`$lte` condition may cause the oplog full scan, so we should just use `$gte`, and compare in the client side.


kelly-cs:
Fixed by [SERVER-29843](https://jira.mongodb.org/browse/SERVER-29843), closing

PR #1245, opened 2018-05-19
SERVER-34975 Include test for date greater than 2038-01-19

asya999:
Added a test for date beyond signed 4-bit representation limit


ramonfm:
@jseyster , @asya999 , looks like there's some progress in SERVER-34975 related to this so I'm going to close this pull request. Please feel free to reopen if this is the wrong thing to do.

PR #1248, opened 2018-06-28
Stop using SIGKILL when running mongod via systemd

dresnick-sf:
It may be that additional changes are needed for stopping the process via `systemd`.

But the docs are clear about not sending `SIGKILL`, which this `systemd.service` file currently does.

### DIFF
--- a/debian/mongod.service
+++ b/debian/mongod.service
@@ -25,6 +25,10 @@ LimitMEMLOCK=infinity
 TasksMax=infinity
 TasksAccounting=false
 
+# Never use kill -9 (i.e. SIGKILL) to terminate a mongod instance.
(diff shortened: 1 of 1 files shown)


ramonfm:
Thanks for working on a pull request @dresnick-sf . Please take a look at the [Contributor's guide](https://github.com/mongodb/mongo/wiki) since there are some steps you have to take before we can consider your pull request, like signing the contributor's agreement and opening a SERVER ticket.

Regards,
Ramón.

ramonfm:
@dresnick-sf,  I don't see your name in the list of people that have signed the contributor's agreement as requested above. Did you get a chance to do this?  Unfortunately we're not able to look at your pull request before you do so.

Thanks,
Ramón.

PR #1255, opened 2018-07-26
SERVER-36399 Cloud E2E testing

gmishkin:
Fills in the project file for Cloud testing

### DIFF
--- a/etc/cloud_nightly.yml
+++ b/etc/cloud_nightly.yml
@@ -32,7 +32,7 @@ tasks:
       local_file: cloud_nightly.json
       remote_file: cloud-nightly/${task_id}
       content_type: application/json
-      display_name: generated_tasks
+      display_name: "Generated tasks (cloud_nightly.json)"
       permissions: private
   - command: generate.tasks
     params:
--- a/etc/cloud_nightly.yml
+++ b/etc/cloud_nightly.yml
@@ -1,1 +1,48 @@
-# Evergreen configuration for MongoDB Cloud E2E tests with MongoDB
+stepback: true
+
+# Needed til EVG-3714 is deployed
+functions:
+  something:
+  - command: shell.exec
+    params:
+      shell: bash
+      script: |
+        exit
+
+tasks:
+- name: GENERATE_MMS_TESTS
+  commands:
+  - command: shell.exec
+    params:
+      shell: bash
+      script: |
+        set -ex
+        git clone --branch ${cloud_branch} --depth 1 git@github.com:${cloud_repo} mms
+        cd mms/scripts
+        ./init_virtualenv.sh ${workdir} 'no-om'
+        VIRTUAL_ENV_DISABLE_PROMPT=true source venv-activate
+        cd evergreen
+        sleep ${compile_wait_duration_seconds}
+        ./mongo_cloud_nightly/generate_cloud_nightly.py > ${workdir}/cloud_nightly.json
+  - command: s3.put
+    params:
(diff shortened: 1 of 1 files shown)


etc/cloud_nightly.yml
```diff
@@ -1 +1,48 @@
...
+    params:
+      shell: bash
+      script: |
+        set -ex
+        git clone --branch ${cloud_branch} --depth 1 git@github.com:${cloud_repo} mms
+        cd mms/scripts
+        ./init_virtualenv.sh ${workdir} 'no-om'
+        VIRTUAL_ENV_DISABLE_PROMPT=true source venv-activate
+        cd evergreen
+        sleep ${compile_wait_duration_seconds}
```
visemet: I was under the impression that you were going to poll to wait for the compile task to run if it hadn't already. Is that going to happen as a follow-up pull request?

It seems like it'd be beneficial to have a separate script in the `mongo_cloud_nightly/` directory that is responsible for waiting for the compile task to have run so the implementation can be changed without needing to make further changes to this embedded bash script.

etc/cloud_nightly.yml
```diff
@@ -1 +1,48 @@
...
+        ./mongo_cloud_nightly/generate_cloud_nightly.py > ${workdir}/cloud_nightly.json
```
visemet: [question] Is the Evergreen task meant to fail if the compile task hadn't run? I wasn't sure what definition for "red", "green", "purple", and "lavender" the Cloud team had wanted to use and what return code the `generate_cloud_nightly.py` script would exit with if the compile task hadn't run.

etc/cloud_nightly.yml
```diff
@@ -1 +1,48 @@
...
+  - command: s3.put
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      bucket: mciuploads
+      local_file: cloud_nightly.json
+      remote_file: cloud-nightly/${task_id}
+      content_type: application/json
+      display_name: generated_tasks
```
visemet: [optional nit] "Generated tasks (cloud_nightly.json)" might be a more slightly more informative name for the link in the "Files" section.

etc/cloud_nightly.yml
gmishkin: Yeah the polling happens externally because I set it up so that tasks can be generated for multiple server buildvariants if desired, and each one would have a different compile task that would need to be waited on. If you have access to the cloud repo I can show you.

The interval in this repo is more about a common sense minimum duration to apply to all cases, and to avoid any race condition where it will try to generate tasks that wait on tasks that have not been created yet because Evergreen dispatched tasks for this project before the server project itself got a new version created in Evergreen, and to reduce the time there would need to be potentially multiple Evergreen hosts waiting on tasks, if we ever have a case where we are using multiple buildvariants of the server.

etc/cloud_nightly.yml
gmishkin: This task should always succeed because it will always generate tasks just based on manipulating fragments of YAML we already have in the cloud repo. It would only fail if someone made incompatible changes to the YAML, or if there were a bug in the script that generates task definitions.

In general, the generated tasks will not fail if the compile fails since there's nothing we can do about that on our end. I have a separate status code for this in the script that waits for tasks to finish.

etc/cloud_nightly.yml
visemet: Got it, thanks for the detailed explanation.

PR #1258, opened 2018-08-30
about V4.0 gridfs C++ API

cxxcunow:
there is gridfs API in old release, but where can I get the gridFs C++ API for version4.0? thanks.


Schubes:
Hi @cxxcunow,

For MongoDB-related support, [mongodb-user group](http://groups.google.com/group/mongodb-user) or [Stack Overflow with the mongodb tag](http://stackoverflow.com/questions/tagged/mongodb). A question like this involving more discussion would be best posted on the mongodb-users group.

Thanks,
Kelsey

PR #1259, opened 2018-09-19
SERVER-37198 Add test for $arrayToObject behaviour with duplicate keys

jaksz:
According to the [documentation](https://docs.mongodb.com/manual/reference/operator/aggregation/arrayToObject/#behavior), when $arrayToObject encounters duplicate keys, it takes the value from the first key. This commit adds a test for this behaviour.


puppyofkosh:
Thanks for the PR! Fortunately the proposed test was covered by other work. See discussion [here](https://jira.mongodb.org/browse/SERVER-37198).

PR #1261, opened 2018-10-13
Add markdown to README

gaiscioch:
Added Markdown text to README, increasing readability.


Schubes:
Thanks for the pull request @gaiscioch. Please note that the main goal of the README is not to look pretty on github, but to be readable with system tools like cat, less, etc., so we're not accepting pull requests for format changes to the README at this time.

Kind regards,
Kelsey

PR #1277, opened 2018-11-25
static_cast -> double will be better data type here. It's better than long long for this case.

shoaibrayeen:


ramonfm:
Thanks for your pull request. Please take a look at the [Contributing to MongoDB](https://github.com/mongodb/mongo/wiki)
section of the wiki for more information on how to get started.

Thanks,
Ramón.

acmorrow:
@shoaibrayeen - Why do you think `double` is a better data type here?

PR #1278, opened 2018-11-25
Removing unnecessary Variable

shoaibrayeen:
first will be updated only once when i == 0          
So no need to take another variable. We can simply do same thing using "if" when i != 0 as first won't be updated after first iteration. So removing that variable.


ramonfm:
Thanks for your pull request. Please take a look at the [Contributing to MongoDB](https://github.com/mongodb/mongo/wiki)
section of the wiki for more information on how to get started.

Thanks,
Ramón.

PR #1279, opened 2018-11-30
SERVER-38338 add helpful error ouput to shell

terakilobyte:
This adds some helpful output to the shell if a new user enters `mongo` or `mongo someUri`, rather than `[js] SyntaxError: missing ; before statement @(shell):1:6`

### DIFF
--- a/src/mongo/shell/utils.js
+++ b/src/mongo/shell/utils.js
@@ -849,6 +849,12 @@ shellHelper.it = function() {
     shellPrintHelper(___it___);
 };
 
+shellHelper.mongo = function() {
+    print("You've typed the mongo command within a mongo shell.");
+    print("This is an error.");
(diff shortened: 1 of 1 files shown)


src/mongo/shell/utils.js
```diff
@@ -849,6 +849,12 @@ shellHelper.it = function() {
     shellPrintHelper(___it___);
 };
 
+shellHelper.mongo = function() {
+    print("You've typed the mongo command within a mongo shell.");
+    print("This is an error.");
+    print("If you meant to run a new mongo shell, please exit the current shell.");
```
rkumar-mongo: @terakilobyte thought: We could add a dochub link here to point users to https://docs.mongodb.com/manual/crud/ ,so they immediately have a point of reference to start using the shell.

PR #1288, opened 2018-12-29
socket connection when ssl is opened

leoloel:
Solve the problem that mongod can not be connected by socket connection when ssl is opened.

### DIFF
--- a/src/mongo/transport/transport_layer_asio.cpp
+++ b/src/mongo/transport/transport_layer_asio.cpp
@@ -433,11 +433,7 @@ StatusWith<SessionHandle> TransportLayerASIO::connect(HostAndPort peer,
     auto session = std::move(sws.getValue());
     session->ensureSync();
 
-#ifndef _WIN32
-    if (endpoints.front().family() == AF_UNIX) {
-        return static_cast<SessionHandle>(std::move(session));
-    }
-#endif
+
 
 #ifndef MONGO_CONFIG_SSL
     if (sslMode == kEnableSSL) {
@@ -452,6 +448,12 @@ StatusWith<SessionHandle> TransportLayerASIO::connect(HostAndPort peer,
         if (!sslStatus.isOK()) {
             return sslStatus;
         }
+    }
+#endif
+
+#ifndef _WIN32
+    if (endpoints.front().family() == AF_UNIX) {
+        return static_cast<SessionHandle>(std::move(session));
     }
 #endif
 


acmorrow:
@jbreams - Could you please take a look? Thanks.

jbreams:
@leoloel I'm trying to understand the issue this pull request is meant to solve. It looks like it will let you use SSL over UNIX domain sockets, which isn't something we've supported in the past. Can you explain the motivation for wanting to do that?

Also, before we can consider merging this, we will need to have a corresponding ticket in our [Jira bug tracker](https://jira.mongodb.org), the commit message should reference that ticket number, and we'll need you to sign our [contributor agreement](https://www.mongodb.com/legal/contributor-agreement).

leoloel:
> @leoloel I'm trying to understand the issue this pull request is meant to solve. It looks like it will let you use SSL over UNIX domain sockets, which isn't something we've supported in the past. Can you explain the motivation for wanting to do that?
> 
> Also, before we can consider merging this, we will need to have a corresponding ticket in our [Jira bug tracker](https://jira.mongodb.org), the commit message should reference that ticket number, and we'll need you to sign our [contributor agreement](https://www.mongodb.com/legal/contributor-agreement).

@jbreams Thanks for your reply. Yes I want to use SSL over UNIX domain sockets, which is supported in v3.4 and v3.6. Also, I have a question that I think no certificates are required when connecting to mongod with UNIX domain sockets whether or not SSL is turned on.

jbreams:
SSL over domain sockets was possible before 4.0, but it was not intended as a supported feature, and its removal in 4.0 was intentional. Since UNIX sockets only connect through the local system's memory, and the kernel can see the plaintext bytes anyways, using SSL over UNIX sockets did not seem like a feature that actually adds any security. I also think certificate verification is going to be broken unless your certificates have the UNIX path as their CN or alternate subject name. So you don't really get useful encryption or identity verification of the server. This feature was something we considered somewhat broken when it was removed, so I think we'll need some justification for bringing it back. Is there a security reason or use case this feature would enable?

Also, as I said in my reply, before considering merging this we'll need a ticket opened in our [Jira bug tracker](https://jira.mongodb.org/) and you will need to sign our [contributor agreement](https://www.mongodb.com/legal/contributor-agreement).

kelly-cs:
Closing this - please consider a [feature request here](https://feedback.mongodb.com/)

PR #1290, opened 2019-01-02
Fix gotools path problem.

utkonos:
These two commits changed the path of gotools by accident it appears. Looks like a copy paste error.

https://github.com/mongodb/mongo/commit/1c7e9f40da962b38151e6bdbc1443d7b55366718

https://github.com/mongodb/mongo/commit/647739b585c3d6dff7a883b8299b36412e71bb0c

My pull request changes that one path back to correct.


ramonfm:
@utkonos, this change is intentional as we're making some improvements in how we build and ship the tools. If you're having issues building after this change, will you please open a TOOLS ticket in https://jira.mongodb.org and provide some details so we can look into it?

Thanks,
Ramón.

Cc @xdg

PR #1291, opened 2019-01-16
SERVER-39057 Add distance expressions for image feature comparison 

marcvivet-we:
We added these expressions:
 - '$cossim', '$chi2', '$euclidean', '$squared_euclidean', '$manhattan'

Which allow us to compare long vectors (image features) stored as arrays or BSON.
It is useful to find the most similar images in a dataset. The usage is the following:

```python
db.test_speed.aggregate([
    {   
        '$project':
        {
            'id': '$id',
            "other_id": '$other_id',
            'distance': {'$cossim': [vector, '$vector']},
        },
    },
    {"$sort": {"distance": -1}},
    {"$limit": 20}
])
```

In addition implementations using avx2 and avx512 are included in this pull request.

### DIFF
--- a/src/mongo/db/db.cpp
+++ b/src/mongo/db/db.cpp
@@ -309,6 +309,27 @@ ExitCode _initAndListen(int listenPort) {
         l << (is32bit ? " 32" : " 64") << "-bit host=" << getHostNameCached() << endl;
     }
 
+    {
+        LogstreamBuilder l = log(LogComponent::kControl);
+
+        l << "Distance Expression Info: ";
+
+#ifdef DISTANCE_EXPRESSION_NOT_BSON 
+        l << "Not using BSON";
+#else 
+        l << "Using BSON";
+#ifdef USE_AVX512
+        l << " - With AVX512";
+#else
+#if USE_AVX2
+        l << " - With AVX2";
+#endif
+#endif
+#endif
+
+        l << std::endl;
+    }
+
     DEV log(LogComponent::kControl) << "DEBUG build (which is slower)" << endl;
 
 #if defined(_WIN32)
--- a/SConstruct
+++ b/SConstruct
@@ -433,6 +433,21 @@ add_option('disable-warnings-as-errors',
     nargs=0,
 )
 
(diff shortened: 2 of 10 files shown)


johnlpage:
I've seen a number of customer use cases where this would have been really useful

OscarRL:
I think that our implementation still needs some improvements.. 

1. I suggest to create a special data type "Vector" to  handle float32, float16.. etc vectors in binary format to avoid confusions with the BinData type.

2. Memory aligment and batch processing.. instead of compute vectors distances one by one we can improve the speed alot by creating batches of thousends of vectors to compute at once.

3. More testing and unit cases.

Any help is appreciated

marcvivet-we:
What it would be nice to speed up even more these operations would be to have a new function on the Value object that returns the memory aligned:

so instead of -> value1.getBinData(); -> value1.getAlignedBinData();

This can be done by changing the memory allocation to __mm_malloc and __mm_free (alignment of 32bits). This would allow us to use _mmXXX_load_ps instead of _mm256_loadu_ps.

Schubes:
Hi @marcvivet-we,

Thanks so much for opening this pull request! It looks like a really interesting and worthwhile extension. I see you've already signed the contributor's agreement, so I've gone ahead a created [SERVER-39057](https://jira.mongodb.org/browse/SERVER-39057) to track this issue on your behalf. The query team will look over the pull request and provide more substantive comments.

Thanks again,
Kelsey

src/mongo/db/db.cpp
```diff
@@ -309,6 +309,27 @@ ExitCode _initAndListen(int listenPort) {
         l << (is32bit ? " 32" : " 64") << "-bit host=" << getHostNameCached() << endl;
```
anhems: Những

anhems (changes requested):

src/mongo/db/db.cpp
MiguelPF: Can you explain what is your intention @anhems with this line?
Seems completely unrelated to this PR,

Best,
Miguel

PR #1294, opened 2019-01-24
SERVER-39195 Fix getUserDir to not solely rely on env var HOME

Yuxael:
Shell should not rely solely on env var HOME. It's very often not set at all when launching scripts e.g from cron or other apps. This causes shell to output a warning that it could not save .dbshell and thus making it very hard to parse output from shell as it would result in malformed JSON when retrieving data.
This change won't fix that kind of behaviour on MacOS as users settings are not kept in /etc/passwd so getpwuid() can't be used for this purpose.

### DIFF
--- a/src/mongo/shell/shell_utils.cpp
+++ b/src/mongo/shell/shell_utils.cpp
@@ -49,6 +49,7 @@
 #include "mongo/util/quick_exit.h"
 #include "mongo/util/text.h"
 #include "mongo/util/version.h"
+#include <pwd.h>
 
 namespace mongo {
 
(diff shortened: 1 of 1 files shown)


Schubes:
Hi @Yuxael,

Thank you for the pull request. So we can review these changes, would you please be sure to sign the [contributor agreement](https://www.mongodb.com/legal/contributor-agreement) and open a [SERVER ticket](https://jira.mongodb.org/browse/SERVER/) so we can track the fix?

Thanks again,
Kelsey

Yuxael:
SERVER-39195

PR #1315, opened 2019-07-03
SERVER-27976 Do not accept unknown argument for the explain command.

krk:
Example error message added with this PR:
```js
> db.runCommand({explain: {find: "c"}, unknownArg: "unknown"})
{
	"ok" : 0,
	"errmsg" : "explain command requires a single argument with an optional verbosity",
	"code" : 72,
	"codeName" : "InvalidOptions"
}
```

### DIFF
--- a/jstests/core/explain_invalid_arguments.js
+++ b/jstests/core/explain_invalid_arguments.js
@@ -16,7 +16,7 @@ explain = db.runCommand({
   verosity: "executionStats"
 });
 assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
-assert.eq("explain does not support 'verosity' argument(s).", explain.errmsg);
+assert.eq("explain does not support 'verosity' argument.", explain.errmsg);
 
 explain = db.runCommand({
   explain: { find: collName },
@@ -24,7 +24,15 @@ explain = db.runCommand({
   foo: 1
 });
 assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
-assert.eq("explain does not support 'foo' argument(s).", explain.errmsg);
+assert.eq("explain does not support 'foo' argument.", explain.errmsg);
+
+explain = db.runCommand({
+  explain: { find: collName },
+  foo: 1,
+  bar: 1
+});
+assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
+assert.eq("explain does not support 'foo' argument.", explain.errmsg);
 
 explain = db.runCommand({
   explain: { find: collName },
@@ -33,7 +41,7 @@ explain = db.runCommand({
   bar: 1
 });
 assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
-assert.eq("explain does not support 'foo', 'bar' argument(s).", explain.errmsg);
+assert.eq("explain does not support 'foo' argument.", explain.errmsg);
 
 explain = db.runCommand({
   explain: { find: collName },
@@ -41,4 +49,4 @@ explain = db.runCommand({
   bar: 1
 });
 assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
-assert.eq("explain does not support 'bar' argument(s).", explain.errmsg);
+assert.eq("explain does not support 'bar' argument.", explain.errmsg);
--- a/jstests/core/explain_invalid_arguments.js
+++ b/jstests/core/explain_invalid_arguments.js
@@ -1,0 +1,44 @@
+// Tests for validating arguments of the explain command.
+
+var collName = "jstests_explain_invalid_arguments";
+var t = db[collName];
+t.drop();
+t.insert({});
+
+var explain = db.runCommand({
+  explain: { find: collName },
+  verbosity: "executionStats"
+});
+assert.commandWorked(explain);
+
+explain = db.runCommand({
+  explain: { find: collName },
+  verosity: "executionStats"
+});
+assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
+assert.eq("explain does not support 'verosity' argument(s).", explain.errmsg);
+
+explain = db.runCommand({
+  explain: { find: collName },
+  verbosity: "executionStats",
+  foo: 1
+});
+assert.commandFailedWithCode(explain, ErrorCodes.InvalidOptions);
+assert.eq("explain does not support 'foo' argument(s).", explain.errmsg);
+
+explain = db.runCommand({
+  explain: { find: collName },
+  foo: 1,
+  verbosity: "executionStats",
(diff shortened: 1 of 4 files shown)


dhatcher42:
Thanks for the Pull Request! I've forwarded this on to the Query team to take a look.

src/mongo/db/query/explain_options.cpp
```diff
@@ -80,8 +81,23 @@ StatusWith<ExplainOptions::Verbosity> ExplainOptions::parseCmdBSON(const BSONObj
...
+        verbositySpecified = true;
     }
 
+    int numFields = 0;
+    BSONObjIterator i(cmdObj);
+    while (i.moreWithEOO()) {
+        BSONElement e = i.next();
+        if (e.eoo())
+            break;
+        if (e.fieldNameStringData() == "$db")
```
TedTuckman: The "lsid" field would also have to be allowed. There are some tests in jstests/core that are titled explain that may be helpful to run to show what needs to work.

src/mongo/db/query/explain_options.cpp
```diff
@@ -80,8 +81,23 @@ StatusWith<ExplainOptions::Verbosity> ExplainOptions::parseCmdBSON(const BSONObj
...
+            continue;
+        numFields++;
+    }
+    if (numFields != (verbositySpecified ? 2 : 1)) {
```
TedTuckman: I think it makes more sense here to catch the invalid argument in some kind of default case while you're looping over the fields in the document. That way you're able to tell the user which argument is invalid with a more specific error message. If you do it that way you also may not need the verbositySpecified variable.

src/mongo/db/query/explain_options_test.cpp
```diff
@@ -60,6 +60,18 @@ TEST(ExplainOptionsTest, ExplainOptionsSerializeToBSONCorrectly) {
...
+              ErrorCodes::InvalidOptions);
+    ASSERT_EQ(
+        ExplainOptions::parseCmdBSON(fromjson("{explain: {}, verbosity: 'queryPlanner', foo: 1}"))
+            .getStatus(),
+        ErrorCodes::InvalidOptions);
+    ASSERT_EQ(
+        ExplainOptions::parseCmdBSON(fromjson("{explain: {}, verosity: {foo: 'bar'}}")).getStatus(),
+        ErrorCodes::InvalidOptions);
+}
+
```
TedTuckman: For completeness can you add a test that should work and not assert?

TedTuckman:
I was shown that you already signed the agreement, so never mind that part. Sorry about that

src/mongo/db/query/explain_options.cpp
krk: Thank you for the pointer, I have removed the `verbositySpecified` variable.

src/mongo/db/query/explain_options_test.cpp
krk: I have added a non-asserting test at https://github.com/mongodb/mongo/pull/1315/commits/ce74977927ddf3f24bfd2eb17ce6e97a6b9d121c#diff-a4bfb6c740b4df06e6e94c7d256a3bbbR65, not sure if it is meaningful.

krk:
I have added javascript tests to the new file `explain_invalid_arguments.js`.

jstests/core/explain_invalid_arguments.js
```diff
@@ -0,0 +1,44 @@
+// Tests for validating arguments of the explain command.
```
TedTuckman: So I maybe should have been a little more specific, I apologize. I pointed you to an explain jstest example, but we prefer a different format for jstests. We prefer immediately invoked functions for jstests. See [this test](https://github.com/mongodb/mongo/blob/master/jstests/core/capped_update.js) for an example.

jstests/core/explain_invalid_arguments.js
```diff
@@ -0,0 +1,44 @@
...
+
+var collName = "jstests_explain_invalid_arguments";
+var t = db[collName];
```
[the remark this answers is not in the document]
krk: I have copied most of this test case from `explain_find.js`, will change the variable names.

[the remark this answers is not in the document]
krk: Deleted it instead.

TedTuckman:
This is looking good. I ran our javascript tests against your patch and a large number failed. In order to merge this the tests have to be fixed in the patch with the changes. 

In order to run the javascript tests, see [this page](https://github.com/mongodb/mongo/wiki/Test-The-Mongodb-Server). I would suggest testing using [resmoke.py](https://github.com/mongodb/mongo/wiki/Test-The-Mongodb-Server#test-using-resmokepy) to test suites to figure out which tests in those suites failed. The suites I would look at first are: jsCore, aggregation, auth, noPassthrough, and sharding. There are at least a few tests in each one of those that fail because of this change. Feel free to reach out if you run into any problems with the tests. 

Also, could you run the formatter again?

krk:
Thank you for the explanations. It seems that tests are using the explain command in an undocumented way, at least I could not find any documentation for this behavior:

```js
var cName = "explain_find_and_modify";
var t = db.getCollection(cName);
// Insert a document to make sure the database exists.
t.insert({ will: "be dropped" });
// Make sure the collection doesn't exist.
t.drop();

var explainUpsert = {
  explain: {
    findAndModify: cName,
    update: { $inc: { i: 1 } },
    query: { _id: 0 },
    upsert: true
  }
};

var correct = {
  explain: Object.merge(explainUpsert.explain, { fields: { x: 1 } })
};
db.runCommand(correct);

// explain_find_and_modify.js:54 `assert.commandWorked(db.runCommand(Object.merge(explainUpsert, {fields: {x: 1}})));`
// is equivalent to:
var merged = Object.merge(explainUpsert, { fields: { x: 1 } });
db.runCommand(merged);
```

1. `correct` command succeeds on both vanilla master and this PR.
2. `merged` command succeeds on vanilla and fails with this PR.

Could you point to the documentation or code that handles the `fields` in this:

```js
db.runCommand({
  explain: {
    findAndModify: "explain_find_and_modify",
    update: { $inc: { i: 1 } },
    query: { _id: 0 },
    upsert: true
  },
  fields: { x: 1 }
});
```

krk:
Rebased and formatted with `clang-format-7.0.1`.

TedTuckman:
In this situation my guess is that the fields argument is part of the [findAndModify](https://docs.mongodb.com/manual/reference/method/db.collection.findAndModify/) command, and can probably be moved inside explain. From when I was looking through the tests it looked like there were a number of cases where there were arguments outside the command object instead of inside -- you may have to find and move those.

kelly-cs:
Fixed in 4.7.0, closing

PR #1318, opened 2019-08-15
Update and rename README to README.md

ahtee:


Schubes:
Hi @ahtee,

Thanks for taking the time to do a pull request. Unfortunately we're not able to accept it since the README file also has to be readable in plain text, so we're not changing it to any other format at this time.

Kind regards,
Kelsey

PR #1319, opened 2019-08-18
Add mips64le support

lina-not-linus:


acmorrow:
Interesting work, thanks for sharing it. I have several comments.

- Overall, this looks like the right changes technically, but there are a lot of process aspects of this PR that aren't correct.
- Even without those process issues, there is a serious obstacle to our merging this, which is that we have no access to the necessary hardware to add mips64le to our continuous integration environment. Having such hardware is basically a prerequisite to our adding a new platform.
- There would be a significant cost to our obtaining and operating such hardware, which needs to be justified by a real world business case for the platform, and we are currently unaware of any such demand for mips64le support for MongoDB.

Regarding the process issues:

- The direct changes to the top level SConstruct and the files under src/mongo look reasonably correct, though we lack a way to test them.
- The changes to gperftools and boost should start with PRs to the upstream maintainers of those packages. Once those are committed, we can look into cherry-picking those fixes out of those repositories and into ours, each as separate commits.
- We manually vendor wiredtiger from their upstream, so any changes to support mips64le should start with a PR issued to https://github.com/wiredtiger/wiredtiger

lina-not-linus:
Thank you for your comment on this PR. I didn't expect you didn't have a mips64le machine. Sorry.

I'm really sorry that my rash PR has increased your heavy review work. I will split my patch and PR to these upstream maintainers. I will start PR from https://github.com/wiredtiger/wiredtiger.

Good luck

acmorrow:
@lina-not-linus - I think starting with those upstream PRs will be a good idea. In the future, if you have ideas for changes you would like to see in MongoDB like this one, I recommend reaching out on either our bug tracker at jira.mongodb.org and opening a ticket, or email the [mongodb-dev](https://groups.google.com/forum/#!forum/mongodb-dev) mailing list to discuss what you would like to change. That will give as an opportunity to provide feedback on the concept and suggest a development approach. That way, you don't end up doing a lot of work only to end up not being able to deploy it.

PR #1320, opened 2019-08-22
SERVER-42964 Wrong ulimit recommended settings link in init / systemd files

karras:
This PR updates the referenced ulimit recommended settings documentation link in all init and systemd service files.

### DIFF
--- a/debian/init.d
+++ b/debian/init.d
@@ -126,7 +126,7 @@ running() {
 
 start_server() {
             # Recommended ulimit values for mongod or mongos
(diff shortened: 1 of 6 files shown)


debian/init.d
```diff
@@ -126,7 +126,7 @@ running() {
 
 start_server() {
             # Recommended ulimit values for mongod or mongos
-            # See http://docs.mongodb.org/manual/reference/ulimit/#recommended-settings
+            # See https://docs.mongodb.com/manual/reference/ulimit/#recommended-ulimit-settings
```
mivkov: Why the changes to https and .com here and in all the other files?

PR #1321, opened 2019-08-22
SConstruct: Remove invalid 'f' char

yangx-jy:
Fixes: 3abf0cce9db6("SERVER-33259 add libunwind to third_party")
Signed-off-by: Xiao Yang <xiaox.yang@intel.com>

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -359,8 +359,8 @@ for pack in [
     pretty = name
     if len(pack) == 2:
         pretty = pack[1]
-    add_option(f'use-system-{name}',
-               help=f'use system version of {pretty} library',
+    add_option('use-system-{name}',
+               help='use system version of {pretty} library',
                nargs=0)
 
 add_option('system-boost-lib-search-suffixes',


milkie:
The 'f' isn't invalid.  Isn't it just this? https://realpython.com/python-f-strings/

yangx-jy:
Hi @milkie 

Sorry, I didn't know f-strings syntax before, thanks for your explanation.

It seems that the new string formatting syntax is availabe since Python3.6, so I get the "invalid syntax" when using Python3.5 to run buildscripts/scons.py.

Do you want user to use Python3.6 or newer forcely?
Or should we use old string formatting syntax to be compatible with Python 3.5?

Best Regards,
Xiao Yang

milkie:
Hi Xiao Yang,
Yes, indeed in docs/building.md, it says Python 3.7.x is currently required.
-Eric

yangx-jy:
Hi @milkie 

Could we correct required Python version in SConstruct? It is inconsistent with definition in doc.
---------------------------------------------------
grep -r EnsurePythonVersio SConstruct 
EnsurePythonVersion(3, 5)
--------------------------------------------------

Best Regards,
Xiao Yang

acmorrow:
@chasinglogic - Should we bump the required python version to 3.6 or 3.7? Or remove the f-string to stay 3.5 compatible?

BillyDonahue:
oops. I didn't know f-strings were so new, but nothing in our testing caught my use of it.
We don't test with 3.5 so it's going to be hard to support it, as we'd have to keep eyeballing changes to SCons files for this sort of thing. I don't think we can do that reliably.

PR #1322, opened 2019-09-10
SERVER-43244 delete limited expired docs in every TTL period

pengzhenyi2015:
A server parameter 'ttlDeleteBatch' is declared to control the maximum deleted docs number every TTL period.
The default value is std::numeric_limits<int>::max(), which means no limitation


pengzhenyi2015:
JIRA: https://jira.mongodb.org/browse/SERVER-43244

yzhang1991:
Hi @pengzhenyi2015 

Thank you for submitting this pull request. We had a discussion internally about this proposed change and decided not to accept this change because we think adding `ttlDeleteBatch` can introduce the risk of never being able to catch up with TTL deletes.

As a side note, we recommend you to keep an eye on https://jira.mongodb.org/browse/SERVER-6712. By making the TTL period configurable it might mitigate this issue by setting a small TTL period.

PR #1328, opened 2019-10-06
SERVER-41365 Change ns not found to Namespace not found

wreiske:
This change is to help alleviate some confusion around "ns". 

There may be more here to change, I just blindly edited this untested on my phone. There may also be more to add here... the errors should maybe include more detailed information or a link to documentation about why the error occurred. 

https://stackoverflow.com/a/44488227 for example shows some reasons for why this error may be triggered. When dropping an index on a collection that doesn't exist, for example.

### DIFF
--- a/src/mongo/db/catalog/drop_collection.cpp
+++ b/src/mongo/db/catalog/drop_collection.cpp
@@ -57,12 +57,12 @@ Status _dropView(OperationContext* opCtx,
                  const NamespaceString& collectionName,
                  BSONObjBuilder& result) {
     if (!db) {
-        return Status(ErrorCodes::NamespaceNotFound, "ns not found");
+        return Status(ErrorCodes::NamespaceNotFound, "Namespace not found");
     }
     auto view =
         ViewCatalog::get(db)->lookupWithoutValidatingDurableViews(opCtx, collectionName.ns());
     if (!view) {
-        return Status(ErrorCodes::NamespaceNotFound, "ns not found");
+        return Status(ErrorCodes::NamespaceNotFound, "Namespace not found");
     }
 
     // Validates the view or throws an "invalid view" error.
@@ -110,7 +110,7 @@ Status _dropCollection(OperationContext* opCtx,
     Lock::CollectionLock collLock(opCtx, collectionName, MODE_X);
     Collection* coll = CollectionCatalog::get(opCtx).lookupCollectionByNamespace(collectionName);
     if (!coll) {
-        return Status(ErrorCodes::NamespaceNotFound, "ns not found");
+        return Status(ErrorCodes::NamespaceNotFound, "Namespace not found");
     }
 
     if (MONGO_unlikely(hangDuringDropCollection.shouldFail())) {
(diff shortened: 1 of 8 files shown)


dhatcher42:
Thank you for submitting this Pull Request. I have linked this to [SERVER-41365](https://jira.mongodb.org/browse/SERVER-41365) which is currently in our backlog. As you mentioned, this does not appear to be a complete change. In order for us to consider this PR, could you go through the rest of the codebase and ensure that what you are submitting is the full extent of what you are trying to resolve?

Additionally, please make sure to sign our [Contributor Agreement](https://www.mongodb.com/legal/contributor-agreement) as its a requirement for any PRs we accept.

wreiske:
I'm going to start changing these each little by little as I find them.

Before:
```
> db.test.dropIndex({});
{
        "ok" : 0,
        "errmsg" : "ns not found",
        "code" : 26,
        "codeName" : "NamespaceNotFound"
}
```
After:
```
> db.test.dropIndex({});
{
        "ok" : 0,
        "errmsg" : "Namespace 'test.test' not found when dropping index",
        "code" : 26,
        "codeName" : "NamespaceNotFound"
}
```

p-mongo:
As I just written in https://jira.mongodb.org/browse/SERVER-41365?focusedCommentId=2472165&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2472165, it will be helpful for backwards compatibility to preserve existing "ns not found" phrase because it is handled specially by some of the drivers.

wreiske:
> As I have just written in https://jira.mongodb.org/browse/SERVER-41365?focusedCommentId=2472165&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2472165, it will be helpful for backwards compatibility to preserve existing "ns not found" phrase because it is handled specially by some of the drivers.

So instead of "Namespace 'test.test' not found when dropping index" we should keep "ns not found"....

For example, "ns not found: 'test.test' not found when dropping index"?

p-mongo:
> So instead of "Namespace 'test.test' not found when dropping index" we should keep "ns not found"....

If the type of object is known, it should be included. For example:

<pre>
Collection db.foo does not exist (ns not found)
</pre>

>   if (ret.errmsg == "ns not found")

Any client code performing an exact match comparison will be affected, I don't think there is anything that can be done about that.

> What do you think about adding a "errdesc" that can contain descriptive information about the error?

I am a driver engineer and not on the server team, therefore my opinion just has to do with drivers. I think if it's possible to include additional information into the existing field structure this would be preferred.

p-mongo:
Looking at the diff so far, I think it is reasonable to change the messages already changed to specify the type of object that is not found (i.e. "Collection" in all of the cases so far).

DiannaHohensee:
SERVER-41365 has been successfully committed and closed.
https://github.com/mongodb/mongo/commit/485fdf001d39c7f809ead645d5d892b5094e9431

wreiske:
Oh wow, amazing! Thank you for taking my contribution :) <3 MongoDB!

PR #1331, opened 2019-10-14
SERVER-44035 Remove duplicate item from allowedFieldNames

tfogo:
`IndexDescriptor::k2dIndexBitsFieldName` is listed in the set twice.


dhatcher42:
This was pulled in as a result of SERVER-44035.

PR #1334, opened 2019-10-28
readme updated

vbc221:
Changed readme to an actual readme.md and formated it.


dhatcher42:
Thanks for taking the time to do a pull request. Unfortunately we're not able to accept it since the README file also has to be readable in plain text, so we're not changing it to any other format at this time.

PR #1336, opened 2019-11-12
Make compilation_db decider compatible with scons 2.

DaanDeMeyer:
This allows the compilation_db tool to be used in projects that still use scons 2.

### DIFF
--- a/site_scons/site_tools/compilation_db.py
+++ b/site_scons/site_tools/compilation_db.py
@@ -49,7 +49,7 @@ class __CompilationDbNode(SCons.Node.Python.Value):
         self.Decider(changed_since_last_build_node)
 
 
-def changed_since_last_build_node(child, target, prev_ni, node):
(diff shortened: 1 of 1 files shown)


acmorrow:
Thank you for the PR.

Currently, the master branch of MongoDB assumes python3 and SCons 3. I'm worried that we are more likely to keep breaking it for you as we iterate it. As an alternative, could you just use the compilation_db from an older stable branch? The implementation has been pretty stable for some time so I don't think you will lose any important features or fixes.

- The most recent version that supports Python2 and SCons 2 is on the v4.0 branch, here: https://github.com/mongodb/mongo/blob/v4.0/site_scons/site_tools/compilation_db.py

Could you see if that works for you in your environment?

DaanDeMeyer:
If the implementation is stable I'll stick to the scons 2 version on the v4.0 branch. Thanks for the quick response!

PR #1337, opened 2019-11-25
add mips64el support, test on loongson 3A3000

loongson-gz:
add mips64el support, test on loongson 3A3000


acmorrow:
@loongson-gz - We have had a prior PR along these lines. Please see the following comment for background on why it is unlikely that we would be able to merge this, especially to an older stable branch: https://github.com/mongodb/mongo/pull/1319#issuecomment-522341924

PR #1340, opened 2019-12-11
SERVER-45063 fix: fix the bug in mongo::uriDecode

levy5307:
In mongo, we use mongo::uriDecode to decode the encoded uri. But there is a bug in corner case. For example: 
If the toDecode string is ```%2```(the %2 must be a wrong uri, because it is partial escape sequence at end of string), the size of toDecode is 2. when i is 0, i + 2 = 2,  so i + 2 > toDecode.size() is not satisfied, so the process does not enter the branch of ```if (i + 2 > toDecode.size()) ```,  which would return a right status like this:
```return Status(ErrorCodes::FailedToParse, "Encountered partial escape sequence at end of string")```

### DIFF
--- a/src/mongo/client/mongo_uri.cpp
+++ b/src/mongo/client/mongo_uri.cpp
@@ -86,7 +86,7 @@ mongo::StatusWith<std::string> mongo::uriDecode(StringData toDecode) {
     for (size_t i = 0; i < toDecode.size(); ++i) {
         const char c = toDecode[i];
         if (c == '%') {
-            if (i + 2 > toDecode.size()) {
+            if (i + 2 >= toDecode.size()) {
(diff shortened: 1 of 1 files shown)


dhatcher42:
Thanks for the pull request! For all code changes, no matter how small, we require submitters to sign our [Contributor Agreement](https://www.mongodb.com/legal/contributor-agreement). Could you please do so?

levy5307:
> Thanks for the pull request! For all code changes, no matter how small, we require submitters to sign our [Contributor Agreement](https://www.mongodb.com/legal/contributor-agreement). Could you please do so?

yes, of course

henrikedin:
Thanks for making this pull request! Merged in b1edffd.

PR #1345, opened 2020-01-14
SERVER-45568 Expand OCSP+CRL features in test cert generator

vincentkam:
https://jira.mongodb.org/browse/SERVER-45568


shreyaskalyan:
Also, along with this can you add a certificate in certs.yml with the mustStaple extension? Make it a server cert (server_ocsp_must_staple.crt or something)

PR #1348, opened 2020-02-06
SERVER-46007 ignore SIGINT when input password in stdin

lifubang:
### Problem:
mongo causes os shell crash if we type ctrl + c when entering password in stdin.

In linux platform:

When we login to mongo server in shell with `-p` args but don't pass the password in the argument list, then mongo shell will prompt us `Enter password:` to enter the password from stdin. At that time, if we send `ctrl +c`(SIGINT) to stdin, the mongo shell will be closed, but at the same time, user's os shell will be crashed, and we can't input anything in the shell. We must close the whole shell session and login again to do other maintaining job.

In windows platform, there is no such problem.

I think we should ignore `ctrl c` when we type password in stdin, just like other database client shell, such as mysql.

### Step to reproduce:
login a new linux shell, and use mongo shell:
```
root@DESKTOP-UVUP1SF:/opt/mongo# build/opt/mongo/mongo localhost:14821/admin -u node -p
```

Press enter key, then it will prompt us to enter password without echo in the screen:
```
root@DESKTOP-UVUP1SF:/opt/mongo# build/opt/mongo/mongo localhost:14821/admin -u node -p
Enter password:
```

At that time, we input some password, but don't press enter key. Instead of it, we type `ctrl c`:
```
root@DESKTOP-UVUP1SF:/opt/mongo# build/opt/mongo/mongo localhost:14821/admin -u node -p
Enter password: 2020-02-06T12:14:21.173+0800 I CONTROL [main] shutting down with code:0
root@DESKTOP-UVUP1SF:/opt/mongo#
```

Because of SIGINT, the mongo shell closed normally, and returned to user's os main shell.

But in user's os main shell, we can't input anything more.
```
root@DESKTOP-UVUP1SF:/opt/mongo# build/opt/mongo/mongo localhost:14821/admin -u node -p
Enter password: 2020-02-06T12:14:21.173+0800 I CONTROL [main] shutting down with code:0
root@DESKTOP-UVUP1SF:/opt/mongo# root@DESKTOP-UVUP1SF:/opt/mongo# root@DESKTOP-UVUP1SF:
```

User's os main shell is crash now.

Signed-off-by: lifubang <lifubang@acmcoder.com>

### DIFF
--- a/src/mongo/util/password.cpp
+++ b/src/mongo/util/password.cpp
@@ -44,6 +44,28 @@
 
 namespace mongo {
 
+#ifndef _WIN32
+char getachar(void)  
+{  
+    struct termios tm, tm_temp;
+    char chr;
+    if (isatty(STDIN_FILENO)) {
+        if(tcgetattr(STDIN_FILENO, &tm) < 0)
+            return -1;
+        tm_temp = tm;
+        cfmakeraw(&tm);
+        if(tcsetattr(STDIN_FILENO, TCSANOW, &tm) < 0)
+            return -1;
+    }
+    chr = getchar();
+    if (isatty(STDIN_FILENO)) {
+        if(tcsetattr(STDIN_FILENO, TCSANOW, &tm_temp) < 0)
+            return -1;
+    }
+    return chr;
+}
+#endif
+
 std::string askPassword() {
     std::string password;
     std::cerr << "Enter password: ";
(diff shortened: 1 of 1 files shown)


kelly-cs:
We have decided not to pursue [SERVER-46007](https://jira.mongodb.org/browse/SERVER-46007). Closing this PR.

PR #1349, opened 2020-02-07
Don't use libdl with OpenBSD

ledeuns:


kelly-cs:
I believe this has changed - closing for now. Please reopen if that is in error.

PR #1351, opened 2020-02-20
SERVER-46275 Add ECDSA w/OCSP support to cert gen tool

vincentkam:
https://jira.mongodb.org/browse/SERVER-46275


jstests/ssl/x509/mkcert.py
```diff
@@ -479,6 +482,35 @@ def process_client_multivalue_rdn(cert):
     os.remove(rsa)
     os.remove(pem)
 
+def convert_ecdsa_key(ec_key_file, pkcs8_key_file):
```
shreyaskalyan: Can we title this method letting people know that this is converting the key to a pkcs8 key?

jstests/ssl/x509/mkcert.py
```diff
@@ -520,17 +571,42 @@ def process_ecdsa_leaf(cert):
...
     x509args = ['openssl', 'x509', '-in', csr, '-out', pem, '-req', '-CA', ca, '-CAkey', ca, '-days', '7300', '-sha256', '-set_serial', serial]
     if mode == 'server':
-        reqargs.append('-reqexts=v3_req')
+        reqargs = reqargs + ['-reqexts', 'v3_req']
         extfile = tempfile.mkstemp()[1]
-        open(extfile, 'wt').write('subjectAltName=DNS:localhost,DNS:127.0.0.1')
+        open(extfile, 'wt').write('basicConstraints=CA:FALSE\n')
+        open(extfile, 'at').write('subjectAltName=DNS:localhost,IP:127.0.0.1\n')
+        open(extfile, 'at').write('subjectKeyIdentifier=hash\n')
+        key_usage = ('keyUsage=nonRepudiation,digitalSignature,keyEncipherment\n' if "responder" in cert['name']
```
shreyaskalyan: Should these extensions be listed in `certs.yml` ?

jstests/ssl/x509/mkcert.py
```diff
@@ -545,11 +621,12 @@ def process_cert(cert):
         process_client_multivalue_rdn(cert)
         return
 
-    if cert['name'] == 'ecdsa-ca.pem':
+    if cert['name'] in ['ecdsa-ca.pem', 'ecdsa-ca-ocsp.pem']:
         process_ecdsa_ca(cert)
         return
 
-    if cert['name'] in ['ecdsa-client.pem', 'ecdsa-server.pem']:
+    if cert['name'] in ['ecdsa-client.pem', 'ecdsa-server.pem', 'ecdsa-server-ocsp.pem',
```
shreyaskalyan: I think we can generalize this to search and see if the cert['name'] is prefixed with ecdsa, something like
`cert['name'].startsWith('ecdsa'):` but @sgolemon may have a stronger opinion on this.

jstests/ssl/x509/mkcert.py
```diff
@@ -26,8 +27,10 @@
 
 CONFIG = Dict[str, Any]
 
-MUST_STAPLE_KEY = b'1.3.6.1.5.5.7.1.24'
-MUST_STAPLE_VALUE = str('DER:30:03:02:01:05').encode('utf-8')
+MUST_STAPLE_KEY_STR = '1.3.6.1.5.5.7.1.24'
+MUST_STAPLE_KEY = bytes(MUST_STAPLE_KEY_STR, "utf-8")
+MUST_STAPLE_VALUE_STR ='DER:30:03:02:01:05'
```
sgolemon: I know this commit isn't adding it, but could either of you document what SEQ: [ 1, NULL ] actually signifies? (Also that I'm decoding that correctly?)

jstests/ssl/x509/mkcert.py
sgolemon: I'd lean to keeping it explicit, but how about we add `tags: [ ecdsa, ocsp, responder ]` type field to the YAML definition?  Then we can fix the slightly-too-magic `if 'responder' in cert['name']` and `if 'ocsp' in cert['name']:` checks above as well.

jstests/ssl/x509/mkcert.py
```diff
@@ -520,17 +571,42 @@ def process_ecdsa_leaf(cert):
     reqargs = ['openssl', 'req', '-new', '-key', key, '-out', csr, '-subj', subject]
     x509args = ['openssl', 'x509', '-in', csr, '-out', pem, '-req', '-CA', ca, '-CAkey', ca, '-days', '7300', '-sha256', '-set_serial', serial]
     if mode == 'server':
-        reqargs.append('-reqexts=v3_req')
+        reqargs = reqargs + ['-reqexts', 'v3_req']
         extfile = tempfile.mkstemp()[1]
-        open(extfile, 'wt').write('subjectAltName=DNS:localhost,DNS:127.0.0.1')
+        open(extfile, 'wt').write('basicConstraints=CA:FALSE\n')
```
sgolemon: This template is getting big enough we might *consider* putting it in an external file which we load from here.

jstests/ssl/x509/mkcert.py
```diff
@@ -479,6 +482,35 @@ def process_client_multivalue_rdn(cert):
...
+    # wrap the PEM-encoded EC key with explicit text header into a PEM-encoded PKCS#8 object
+    pkcs8args = ['openssl', 'pkcs8', '-topk8', '-nocrypt', '-in', ec_key_file, '-out', pkcs8_key_file]
+    subprocess.check_call(pkcs8args)
+
+def process_ecdsa_cert(cert, pem, key, dest, filename, split_pem=True):
+    """Convert the ECDSA key and write the public/private key pair key into .pem file, optionally writing the
+     public key to a .crt file and private e key to a .key file"""
+
+    # copy the public key to a .crt file
+    cert_filename = make_filename({'name': f"{filename}.crt"})
```
sgolemon: I'm slightly uncomfortable with creating (possibly) temporary files in the working tree.  Could we make these `tempfile`s then conditionally move them into place if it turns out we do want to keep them?

jstests/ssl/x509/mkcert.py
vincentkam: I've kept it explicit and also added a "tags" field and am now using that instead of the slightly-too-magic names :)

jstests/ssl/x509/mkcert.py
vincentkam: Noted! I've left it the way it is for now since other drivers work is more pressing at the moment, but if you feel strongly about making this optimization now as part of this PR, I can look into it

jstests/ssl/x509/certs.yml
```diff
@@ -558,7 +558,7 @@ certs:
   - 'trusted-ca.pem'
```
shreyaskalyan: I think the other parts of this file need to be tagged as well, like the other ECDSA certs and the other OCSP certs.

jstests/ssl/x509/mkcert.py
```diff
@@ -520,17 +595,43 @@ def process_ecdsa_leaf(cert):
...
+                        open(extfile, 'at').write(f'{MUST_STAPLE_KEY_STR}={MUST_STAPLE_VALUE_STR}\n')
         x509args.append('-extfile')
         x509args.append(extfile)
 
     subprocess.check_call(['openssl', 'ecparam', '-name', 'prime256v1', '-genkey', '-out', key])
     subprocess.check_call(reqargs)
     subprocess.check_call(x509args)
 
-    open(dest, 'wt').write(get_header_comment(cert) + "\n" + open(pem, 'rt').read() + open(key, 'rt').read())
+    if "responder" in cert['name']:
```
shreyaskalyan: I'm confused as to where you're using the tags field here? Shouldn't it be something like - 
`if 'responder' in cert['tags']:` (with cases for each tag) to reroute the cert for processing from different functions?

jstests/ssl/x509/mkcert.py
```diff
@@ -26,8 +27,14 @@
...
 
-MUST_STAPLE_KEY = b'1.3.6.1.5.5.7.1.24'
-MUST_STAPLE_VALUE = str('DER:30:03:02:01:05').encode('utf-8')
+# tlsfeature = status_request isn't supported by older versions of OpenSSL so we manually define this below
+# 1.3.6.1.5.5.7.1.24: "tls_feature" extension as defined in https://tools.ietf.org/html/rfc7633#section-6
+MUST_STAPLE_KEY_STR = '1.3.6.1.5.5.7.1.24'
+MUST_STAPLE_KEY = bytes(MUST_STAPLE_KEY_STR, "utf-8")
+
+# status_request extension as defined in https://tools.ietf.org/html/rfc4366#section-2.3
+MUST_STAPLE_VALUE_STR ='DER:30:03:02:01:05' # ASN.1 value: SEQUENCE { INTEGER 0x05 (5 decimal) }
```
p-mongo: Missing space after `=`?

jstests/ssl/x509/mkcert.py
```diff
@@ -479,31 +503,88 @@ def process_client_multivalue_rdn(cert):
...
+    convert_ecdsa_key_to_pkcs8(ec_key_file=key, pkcs8_key_file=temp_key_filename)
+
+    # combine public and private key into a .pem file with no comments
+    open(dest, 'wt').write(open(pem, 'rt').read() + open(temp_key_filename, 'rt').read())
+
+    if split_pem: # copy the temp files into .crt + .key files
+        cert_filename = make_filename({'name': f"{filename}.crt"})
+        open(cert_filename, 'wt').write(open(temp_cert_filename, 'rt').read())
+        key_filename = make_filename({'name': f"{filename}.key"})
+        open(key_filename, 'wt').write(open(temp_key_filename, 'rt').read())
```
p-mongo: This line does not close either of the two files that it opens. Consider using shutil.copyfile instead (https://docs.python.org/2/library/shutil.html#shutil.copyfile).

jstests/ssl/x509/mkcert.py
```diff
@@ -520,17 +601,43 @@ def process_ecdsa_leaf(cert):
...
+        open(extfile, 'at').write('subjectAltName=DNS:localhost,IP:127.0.0.1\n')
+        open(extfile, 'at').write('subjectKeyIdentifier=hash\n')
+        key_usage = ('keyUsage=nonRepudiation,digitalSignature,keyEncipherment\n'
+                     if (cert.get('tags') and "responder" in cert['tags'])
+                     else 'keyUsage=digitalSignature,keyEncipherment\n')
+        open(extfile, 'at').write(key_usage)
+        extended_key_usage = ('extendedKeyUsage=serverAuth,clientAuth,OCSPSigning\n'
+                              if (cert.get('tags') and "responder" in cert['tags'])
+                              else 'extendedKeyUsage=serverAuth,clientAuth\n')
+        open(extfile, 'at').write(extended_key_usage)
```
p-mongo: Instead of the individual `open` calls that do not close the file (and do not guarantee that the data has been written at all or before the next `open` call happens), the writes will be better expressed with a scope:

<pre>
with open(extfile, 'wt') as f:
  f.write(...)
  #  ...
  f.write(...)
</pre>

vincentkam: Ah thanks! I've taken your suggestion :)

jstests/ssl/x509/mkcert.py
```diff
@@ -520,17 +602,44 @@ def process_ecdsa_leaf(cert):
...
+                            f.write(f'{MUST_STAPLE_KEY_STR}={MUST_STAPLE_VALUE_STR}\n')
         x509args.append('-extfile')
         x509args.append(extfile)
 
     subprocess.check_call(['openssl', 'ecparam', '-name', 'prime256v1', '-genkey', '-out', key])
     subprocess.check_call(reqargs)
     subprocess.check_call(x509args)
 
-    open(dest, 'wt').write(get_header_comment(cert) + "\n" + open(pem, 'rt').read() + open(key, 'rt').read())
+    if cert.get('tags') and "responder" in cert['tags']:
```
sgolemon: nit; You can write this as:   `if 'responder' in cert.get('tags', []):` to avoid some copypasta.

vincentkam:
Rebased and squashed! I don't have merge permissions, so I leave the rest to @shreyaskalyan 😸

vincentkam:
@shreyaskalyan I've added one more commit to strip the trailing slashes from the ECDSA endpoints (and regenerate the certs), per https://jira.mongodb.org/browse/SERVER-46413.

jstests/ssl/x509/mkcert.py
vincentkam: Thanks for the tip! Nit addressed :)

vincentkam:
Merged here: https://github.com/mongodb/mongo/commit/8289739969ad56e78a21803aebb489f032dc18b9
so I'm closing this PR :) Thanks everyone!

PR #1355, opened 2020-03-21
MSI installer produces incorrect config file

tomasherceg:
If the user selects a shorter path of the log file or data directory, the written buffer is shorter than the original content, and the remaining part of the file is not truncate automatically. This leads to an incorrect config file, and thus the mongod service cannot start. 

(I hope that I haven't done some silly mistake - C++ is really not my language)

### DIFF
--- a/src/mongo/installer/msi/ca/CustomAction.cpp
+++ b/src/mongo/installer/msi/ca/CustomAction.cpp
@@ -331,6 +331,9 @@ extern "C" UINT __stdcall UpdateMongoYAML(MSIHANDLE hInstall) {
         DWORD written;
         if (!WriteFile(hFile, str.c_str(), str.length(), &written, NULL)) {
             CHECKGLE_AND_LOG("Failed to write yaml file");
+        }        
+        if (!SetEndOfFile(hFile) {
+            CHECKGLE_AND_LOG("Failed to truncate yaml file");
         }
     } catch (const std::exception& e) {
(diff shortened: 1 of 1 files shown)


carlchampain:
Thanks for making this Pull Request! We just need you to create a new SERVER ticket on [JIRA](https://jira.mongodb.org/) before we can accept the PR.

tomasherceg:
@carlchampain Thank you, just done that: https://jira.mongodb.org/browse/SERVER-47138

cr-nirbar-git:
@tomasherceg I think the root cause is in the WriteFile call that writes str.length() bytes rather than strlen(str.c_str())
When the replacement string is smaller than the original then str holds more bytes than actually needed

nirbar:
@tomasherceg I think the root cause is in the WriteFile call that writes str.length() bytes rather than strlen(str.c_str())
When the replacement string is smaller than the original then str holds more bytes than actually needed

acmorrow:
@tomasherceg - I ended up merging an equivalent change to yours as part of my work with @nirbar on this and several other MSI bugs. It was just a few steps shorter for me to merge his changes than yours given that I already had his work pulled locally, but I would like to thank you for taking the time and effort to investigate this issue and put together the correct fix. I'm closing this PR now.

tomasherceg:
@acmorrow Happy to help. Thank you all for making such a great product!

PR #1361, opened 2020-05-14
SERVER-48226 Improve TLA+ CI

lemmy:
@ajdavis @will62794 @visualzhou 

* "-workers auto" introduced in https://github.com/tlaplus/tlaplus/commit/4111431a169b96d0b2eb850fd6fb4416cf5af4ff

* OffHeapDiskFPSet: https://lemmster.de/talks/MSc_MarkusAKuppe_1497363471.pdf

* Data that is shared as part of exec stats:
https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/util/ExecutionStatisticsCollector.md

### DIFF
--- a/src/mongo/db/repl/tla_plus/model-check.sh
+++ b/src/mongo/db/repl/tla_plus/model-check.sh
@@ -37,4 +37,4 @@ fi
 
 cd "$1"
 # Defer liveness checks to the end with -lncheck, for speed.
-"$JAVA_BINARY" -XX:+UseParallelGC -Dtlc2.tool.fp.FPSet.impl=tlc2.tool.fp.OffHeapDiskFPSet -cp ../tla2tools.jar tlc2.TLC -lncheck final -workers auto "$TLA_FILE"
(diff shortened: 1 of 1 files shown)


ajdavis:
Thanks! Performance seems about the same with this patch (13m26s in the last successful run, 13m16s when I tried this patch). But I suppose it could be a significant improvement for models with more states.

Have you signed the CLA?: https://www.mongodb.com/legal/contributor-agreement

ajdavis:
Question: if the usage stats service is down, will TLC still run to completion and return exit code 0? We want to avoid spurious test failures.

lemmy:
> Question: if the usage stats service is down, will TLC still run to completion and return exit code 0? We want to avoid spurious test failures.

* Stats do not block TLC and fail silently if e.g. the remote host is down
* OffHeap's properties become noticeable with larger specs and higher core counts
* CLA has been taken care of

ajdavis:
Thanks @lemmy!

PR #1362, opened 2020-05-14
Fix ubuntu 20.04 repos

AlexKovynev:


zakhark:
@AlexKovynev, thanks for making this Pull Request! Even for small items like this we require signing of our [Contributor's Agreement](https://www.mongodb.com/legal/contributor-agreement) before we can accept the PR. Could you please do so?

PR #1363, opened 2020-05-14
Update README

Linag94:


Schubes:
Thanks for taking the time to do a pull request. Unfortunately we're not able to accept it since the README file also has to be readable in plain text, so we're not changing it to any other format at this time. It also appears that a few important sections were removed.

Kind regards,
Kelsey

PR #1366, opened 2020-05-28
Fix mongod.cfg file size when replacing tokens in the custom action. …

nirbar:
…Fix https://jira.mongodb.org/browse/SERVER-47138

When the length of the new replacement texts is lesser than the length of the tokens, then the string object is larger than the text it holds
The fix writes the length of the text rather than of the string object

### DIFF
--- a/src/mongo/installer/msi/ca/CustomAction.cpp
+++ b/src/mongo/installer/msi/ca/CustomAction.cpp
@@ -332,6 +332,9 @@ extern "C" UINT __stdcall UpdateMongoYAML(MSIHANDLE hInstall) {
         if (!WriteFile(hFile, str.c_str(), str.length(), &written, NULL)) {
             CHECKGLE_AND_LOG("Failed to write yaml file");
         }
+        if (!SetEndOfFile(hFile)) {
+            CHECKGLE_AND_LOG("Failed to truncate yaml file");
+        }
     } catch (const std::exception& e) {
         CHECKHR_AND_LOG(E_FAIL, "Caught C++ exception %s", e.what());
     } catch (...) {


src/mongo/installer/msi/ca/CustomAction.cpp
```diff
@@ -329,7 +329,7 @@ extern "C" UINT __stdcall UpdateMongoYAML(MSIHANDLE hInstall) {
         SetFilePointer(hFile, 0, 0, SEEK_SET);
 
         DWORD written;
-        if (!WriteFile(hFile, str.c_str(), str.length(), &written, NULL)) {
+        if (!WriteFile(hFile, str.c_str(), ::strlen(str.c_str()), &written, NULL)) {
```
acmorrow, line 332: @nirbar 

- Did you see the other proposed solution to this issue: https://github.com/mongodb/mongo/pull/1355/files?
- I'd expect `str.length()` to give the same result as `::strlen(str.c_str())`, except in cases where `str` has, say, embedded `\0` bytes. So I don't see how this change addresses the underlying issue.

Also, the ticket has this comment:

> BTW, AFAICT there's another bug resulting in corrupt configuration: when updating MongoDb with MSI installer. IMO the Custom Action isn't run at all in this case. Such the configuration file contains template values like %MONGODB_PATH%.

Do you believe this to be true based on your current reading of the MSI setup? Should a separate ticket be filed for that issue?

nirbar: @acmorrow I saw the proposed change but didn't understand why it should fix the issue. The contributor did comment that he could not build the code so he didn't test it.

str.length() may indeed contain embedded NULL characters when the original string is larger than the string after replacements.
Since std::string.replace(..) replaces text in place, that explains the different lengths comparing str.length() to strlen(str.c_str())

I tested it and after the fix the yaml file looked good

About the other comment, I don't think this should happen since we don't upgrade older versions but install side by side

acmorrow:
@nirbar - I tested this out, and I don't think it addresses the issue, which I can reproduce after applying your change. I suspect that the older PR is actually the right fix. What I observe is that if you shorten the values for those two fields, then the file has junk at the end, which suggests that the missing file truncation may in fact be the issue. I don't think the C++ `strlen(str.c_str())` change has the effect you intend.

nirbar:
@acmorrow I just tested it with the same fix and I think it looks good
Am I getting something wrong? See the attached mongod.cfg (had to change the extension because github doesn't accept .cfg)
[mongod.txt](https://github.com/mongodb/mongo/files/4721441/mongod.txt)

nirbar:
@acmorrow Got it, I applied both fixes to be on the safe side

src/mongo/installer/msi/ca/CustomAction.cpp
```diff
@@ -329,8 +329,11 @@ extern "C" UINT __stdcall UpdateMongoYAML(MSIHANDLE hInstall) {
...
             CHECKGLE_AND_LOG("Failed to write yaml file");
+        }
+		if (!SetEndOfFile(hFile)) {
```
acmorrow, line 335: Indentation here looks incorrect.

src/mongo/installer/msi/ca/CustomAction.cpp
acmorrow: @nirbar - No, `std::string::length` adjusts correctly to reflect the valid span of characters post `std::string::replace`. But invoking `::strlen(str.c_str())` will give an inaccurate length if the post-replacement string does in fact contain null bytes. So I think using `std::string::length` here is and always has been correct. Could you revert that part of this review? I think that really the original PR, #1355, was entirely correct and sufficient (thanks @tomasherceg). We should consider just merging that PR.

nirbar: @acmorrow std::string::length returns the capacity of the buffer managed by this instance, which may be different than string length.
Whether or not std::string::do_replace reallocates the buffer with the actual string length or keeps the existing buffer if it larger than required, is implementation specific and out of our control.
The yaml file is textual so it definitely not supposed to hold null characters
Overall I think this is more fail proof than just setting the end of file

acmorrow: @nirbar - No, that isn't correct, as the following simple example will demonstrate:

```
#include <cstdlib>
#include <iostream>
#include <string>

int main() {
    std::string lengthy{"I'm a looong string"};
    std::cout << lengthy << "\n";
    std::cout << "Lengthy before replace: " << lengthy.size() << " " << lengthy.length() << " " << lengthy.capacity() << "\n";
    lengthy.replace(6, 6, "trim");
    std::cout << lengthy << "\n";
    std::cout << "Lengthy after replace: " << lengthy.size() << " " << lengthy.length() << " " << lengthy.capacity() << "\n";
    return EXIT_SUCCESS;
}
```

This produces the following output:

 ```
I'm a looong string
Lengthy before replace: 19 19 22
I'm a trim string
Lengthy after replace: 17 17 22
```

It is `std::string::capacity` that returns the capacity. The `std::string::length` and `std::string::size` methods are equivalent and return number of valid characters in the current contents of the string.

Comment (nirbar) [reply src/mongo/installer/msi/ca/CustomAction.cpp:332]: @acmorrow What happens if you set a character in the middle to null?
What are the string length, size, and capacity then?

Comment (acmorrow) [reply src/mongo/installer/msi/ca/CustomAction.cpp:332]: @nirbar - Lets try it out and see! Here I've also added the result of `::strlen(lengthy.c_str())` so we can observe how it changes:

```
$ cat foo.cpp
#include <cstdlib>
#include <iostream>
#include <string>

int main() {
    std::string lengthy{"I'm a looong string"};
    std::cout << lengthy << "\n";
    std::cout << "Lengthy before replace: " << lengthy.size() << " " << lengthy.length() << " " << lengthy.capacity() << " " << ::strlen(lengthy.c_str()) << "\n";

    lengthy.replace(6, 6, "trim");
    std::cout << lengthy << "\n";
    std::cout << "Lengthy after replace: " << lengthy.size() << " " << lengthy.length() << " " << lengthy.capacity() << " " << ::strlen(lengthy.c_str()) << "\n";


    lengthy[4] = '\0';
    std::cout << lengthy << "\n";
    std::cout << "Lengthy after null byte: " << lengthy.size() << " " << lengthy.length() << " " << lengthy.capacity() << " " << ::strlen(lengthy.c_str()) << "\n";

    return EXIT_SUCCESS;
}
```

```
$ g++ foo.cpp -std=c++17
```

```
$  ./a.out
I'm a looong string
Lengthy before replace: 19 19 22 19
I'm a trim string
Lengthy after replace: 17 17 22 17
I'm  trim string
Lengthy after null byte: 17 17 22 4
```

nirbar: @acmorrow So I think that strlen(str.c_str()) does the job considering we're dealing with yaml which is textual
However given the specific scenario str.length() should return the same result
Up to you to merge this or PR by @tomasherceg

acmorrow:
@nirbar - Could you squash this down to one commit and rebase it and I'll get it merged?

nirbar:
@acmorrow Done

acmorrow:
> @acmorrow Done

It looks like the strlen thing is still there though. I'd prefer we drop that as I really don't think it is part of this fix.

nirbar:
@acmorrow Done

acmorrow:
Thanks @nirbar, I'll get this merged tonight.

PR #1367, opened 2020-05-28
Set valid result for service account check when using the builtin account

nirbar:
Set valid result for service account check when using the builtin account. Fix https://jira.mongodb.org/browse/SERVER-46445

### DIFF
--- a/src/mongo/installer/msi/wxs/UIFragment.wxs
+++ b/src/mongo/installer/msi/wxs/UIFragment.wxs
@@ -71,9 +71,10 @@
                     <Text>&amp;Next &gt;</Text>
                     <Publish Event="AddLocal" Value="ServerNoService">NOT MONGO_SERVICE_INSTALL</Publish>
                     <Publish Event="AddLocal" Value="ServerService">MONGO_SERVICE_INSTALL</Publish>
-                    <Publish Event="DoAction" Value="ValidateServiceLogon" Order="1"><![CDATA[MONGO_SERVICE_INSTALL AND MONGO_SERVICE_ACCOUNT_TYPE <> "ServiceLocalNetwork"]]></Publish>
-                    <Publish Event="SpawnDialog" Value="InvalidServiceAccount" Order="2">MONGO_SERVICE_ACCOUNT_VALID = "0"</Publish>
-                    <Publish Event="NewDialog" Value="CompassDlg" Order="2">NOT MONGO_SERVICE_INSTALL OR (MONGO_SERVICE_ACCOUNT_TYPE = "ServiceLocalNetwork" OR MONGO_SERVICE_ACCOUNT_VALID)</Publish>
+                    <Publish Property="MONGO_SERVICE_ACCOUNT_VALID" Value="1" Order="1"><![CDATA[MONGO_SERVICE_ACCOUNT_TYPE ~= "ServiceLocalNetwork"]]></Publish>
+                    <Publish Event="DoAction" Value="ValidateServiceLogon" Order="2"><![CDATA[MONGO_SERVICE_INSTALL AND MONGO_SERVICE_ACCOUNT_TYPE <> "ServiceLocalNetwork"]]></Publish>
+                    <Publish Event="SpawnDialog" Value="InvalidServiceAccount" Order="3">MONGO_SERVICE_ACCOUNT_VALID = "0"</Publish>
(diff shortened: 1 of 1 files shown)


src/mongo/installer/msi/wxs/UIFragment.wxs
```diff
@@ -71,9 +71,10 @@
...
-                    <Publish Event="SpawnDialog" Value="InvalidServiceAccount" Order="2">MONGO_SERVICE_ACCOUNT_VALID = "0"</Publish>
-                    <Publish Event="NewDialog" Value="CompassDlg" Order="2">NOT MONGO_SERVICE_INSTALL OR (MONGO_SERVICE_ACCOUNT_TYPE = "ServiceLocalNetwork" OR MONGO_SERVICE_ACCOUNT_VALID)</Publish>
+                    <Publish Property="MONGO_SERVICE_ACCOUNT_VALID" Value="1" Order="1"><![CDATA[MONGO_SERVICE_ACCOUNT_TYPE ~= "ServiceLocalNetwork"]]></Publish>
+                    <Publish Event="DoAction" Value="ValidateServiceLogon" Order="2"><![CDATA[MONGO_SERVICE_INSTALL AND MONGO_SERVICE_ACCOUNT_TYPE <> "ServiceLocalNetwork"]]></Publish>
+                    <Publish Event="SpawnDialog" Value="InvalidServiceAccount" Order="3">MONGO_SERVICE_ACCOUNT_VALID = "0"</Publish>
+                    <Publish Event="NewDialog" Value="CompassDlg" Order="3">NOT MONGO_SERVICE_INSTALL OR (MONGO_SERVICE_ACCOUNT_TYPE = "ServiceLocalNetwork" OR MONGO_SERVICE_ACCOUNT_VALID)</Publish>
```
acmorrow, line 77: It looks like there are now two `order=3` entries. Is that intented?

src/mongo/installer/msi/wxs/UIFragment.wxs
```diff
@@ -71,9 +71,10 @@
                     <Text>&amp;Next &gt;</Text>
                     <Publish Event="AddLocal" Value="ServerNoService">NOT MONGO_SERVICE_INSTALL</Publish>
                     <Publish Event="AddLocal" Value="ServerService">MONGO_SERVICE_INSTALL</Publish>
-                    <Publish Event="DoAction" Value="ValidateServiceLogon" Order="1"><![CDATA[MONGO_SERVICE_INSTALL AND MONGO_SERVICE_ACCOUNT_TYPE <> "ServiceLocalNetwork"]]></Publish>
-                    <Publish Event="SpawnDialog" Value="InvalidServiceAccount" Order="2">MONGO_SERVICE_ACCOUNT_VALID = "0"</Publish>
-                    <Publish Event="NewDialog" Value="CompassDlg" Order="2">NOT MONGO_SERVICE_INSTALL OR (MONGO_SERVICE_ACCOUNT_TYPE = "ServiceLocalNetwork" OR MONGO_SERVICE_ACCOUNT_VALID)</Publish>
+                    <Publish Property="MONGO_SERVICE_ACCOUNT_VALID" Value="1" Order="1"><![CDATA[MONGO_SERVICE_ACCOUNT_TYPE ~= "ServiceLocalNetwork"]]></Publish>
```
acmorrow, line 74: If I understand correctly, the issue was that because the `MONGO_SERVICE_ACCOUNT_VALID` property was unpublished, the conditionals in the following `Publish` sections did not work as expected, so the result is simply to do the publish. Which required introducing a new `order=1` step and renumbering the others. (GitHub seems to have done a poor job creating the diff).

src/mongo/installer/msi/wxs/UIFragment.wxs
acmorrow: Actually, reading the old code again, I see that there were previously two `order=2` items. So I guess it makes sense that there are now two `order=3`. These are disjoint outcomes, presumably.

nirbar:
@acmorrow You got it right again.
The order doesn't matter here since only one of the options can evaluate to true

acmorrow:
@nirbar - This LGTM. Please drop the `:` out of the commit message and I will merge this.

nirbar:
@acmorrow Done

acmorrow:
@nirbar - I hand tested MSIs built from both the base commit and this change. I was able to reproduce this issue on the base commit but not with your change.

acmorrow:
@nirbar - Could you please rebase this to latest master and push to your branch?

PR #1369, opened 2020-06-24
SERVER-39191 Optimize count queries on sharded collections

Sebmaster:
When counts were made orphan-aware the addition of the SHARD_FILTER
stage started preventing the COUNT_SCAN optimization on all sharded
collections, leading to ~5x performance degradation due to the necessity
of IXSCAN (and possibly FETCH).

However, there can not be any orphans if the shard key is fully
specified: A mongodb instance will never have orphans for chunk ranges
it owns and mongos will send a query with a full shard key to one
replicaset only.

This allows stripping the SHARDING_FILTER stage for cases where the
query specifies the full shard key, regaining the lost performance.

### DIFF
--- a/src/mongo/db/query/get_executor.cpp
+++ b/src/mongo/db/query/get_executor.cpp
@@ -305,7 +305,21 @@ void fillOutPlannerParams(OperationContext* opCtx,
         auto collDesc = CollectionShardingState::get(opCtx, canonicalQuery->nss())
                             ->getCollectionDescription(opCtx);
         if (collDesc.isSharded()) {
-            plannerParams->shardKey = collDesc.getKeyPattern();
+            const auto& keyPattern = collDesc.getKeyPattern();
+            ShardKeyPattern shardKeyPattern(keyPattern);
+
+            // If the shard key is specified exactly, the query is guaranteed to only target one
+            // shard. Shards cannot own orphans for the key ranges they own, so there is no need
+            // to include a shard filtering stage. By omitting the shard filter, it may be possible
+            // to get a more efficient plan (for example, a COUNT_SCAN may be used if the query is
+            // eligible).
+            const BSONObj extractedKey = shardKeyPattern.extractShardKeyFromQuery(*canonicalQuery);
+
+            if (extractedKey.isEmpty()) {
+                plannerParams->shardKey = keyPattern;
+            } else {
+                plannerParams->options &= ~QueryPlannerParams::INCLUDE_SHARD_FILTER;
+            }
         } else {
             // If there's no metadata don't bother w/the shard filter since we won't know what
             // the key pattern is anyway...
--- a/jstests/sharding/queries_elide_shard_filter.js
+++ b/jstests/sharding/queries_elide_shard_filter.js
@@ -1,0 +1,107 @@
+//
+// Tests that queries in sharded collections will be properly optimized
+//
+// @tags: [requires_fcv_48]
+
+(function() {
+"use strict";
+
+load("jstests/libs/analyze_plan.js");
+
+function assertShardFilter(explain) {
+    const filterStage = getPlanStage(explain.queryPlanner.winningPlan, "SHARDING_FILTER");
+    assert.eq(filterStage.stage, "SHARDING_FILTER");
+    const scanStage = filterStage.inputStage;
+    assert.contains(scanStage.stage, ["IXSCAN", "FETCH"]);
+}
+
+function assertNoShardFilter(explain) {
+    const filterStage = getPlanStage(explain.queryPlanner.winningPlan, "SHARDING_FILTER");
+    assert.eq(filterStage, null, explain);
+}
+
+function assertCountScan(explain) {
+    const countStage = getPlanStage(explain.queryPlanner.winningPlan, "COUNT_SCAN");
+    assert.eq(countStage.stage, "COUNT_SCAN");
+}
+
+const st = new ShardingTest({shards: 1});
+const coll = st.s0.getCollection("foo.bar");
+
+function createCollection(coll, shardKey) {
+    coll.drop();
+    assert.commandWorked(st.s0.adminCommand({shardCollection: coll.getFullName(), key: shardKey}));
+
+    assert.commandWorked(coll.insert({_id: true, a: true, b: true, c: true, d: true}));
+    assert.commandWorked(coll.createIndex({a: 1, b: 1}));
+    assert.commandWorked(coll.createIndex({b: 1, a: 1}));
+}
+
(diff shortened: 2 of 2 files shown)


puppyofkosh:
I will take some time to review this when my current project wraps up in a few weeks.

jstests/sharding/count_queryplan.js
```diff
@@ -0,0 +1,77 @@
...
+createCollection(coll, {a: 1, b: 1});
+
+explain = assert.commandWorked(coll.explain('executionStats').count({a: true}));
+assertShardFilter(explain);
+explain = assert.commandWorked(coll.explain('executionStats').count({a: true, b: true}));
+assertCountScan(explain);
+explain = assert.commandWorked(coll.explain('executionStats').count({a: true, b: {$in: [true, false]}}));
+assertShardFilter(explain);
+
+
```
puppyofkosh, line 64: We avoid multiple lines of vertical whitespace, so delete one of these blank lines.

src/mongo/db/query/get_executor.cpp
```diff
@@ -306,7 +306,20 @@ void fillOutPlannerParams(OperationContext* opCtx,
         auto collDesc = CollectionShardingState::get(opCtx, canonicalQuery->nss())
                             ->getCollectionDescription(opCtx);
         if (collDesc.isSharded()) {
-            plannerParams->shardKey = collDesc.getKeyPattern();
+            const auto keyPattern = collDesc.getKeyPattern();
```
puppyofkosh, line 309: Suggest using `const auto&` here to avoid copying the key pattern object unnecessarily.

src/mongo/db/query/get_executor.cpp
```diff
@@ -306,7 +306,20 @@ void fillOutPlannerParams(OperationContext* opCtx,
...
+            ShardKeyPattern shardKeyPattern(keyPattern);
+
+            // If we're specifying one shard exactly, we're always going to land on exactly one
```
puppyofkosh, line 312: I find the first sentence of this comment a little awkward. I think the intention was to say "if we're specifying an exact value for the shard key, we're always going ...". I would suggest rewording the comment to something like this:

```
// If the shard key is specified exactly, the query is guaranteed to only target one
// shard. Shards cannot own orphans for the key ranges they own, so there is no need
// to include a shard filtering stage. By omitting the shard filter, it may be possible
// to get a more efficient plan (for example, a COUNT_SCAN may be used if the query is
// eligible).
```

jstests/sharding/count_queryplan.js
```diff
@@ -0,0 +1,77 @@
...
+jsTest.log('Tests with compound shard key');
+createCollection(coll, {a: 1, b: 1});
+
+explain = assert.commandWorked(coll.explain('executionStats').count({a: true}));
+assertShardFilter(explain);
+explain = assert.commandWorked(coll.explain('executionStats').count({a: true, b: true}));
+assertCountScan(explain);
+explain = assert.commandWorked(coll.explain('executionStats').count({a: true, b: {$in: [true, false]}}));
```
puppyofkosh, line 61: This line is just over our 100-char limit, so suggest changing to:
```
explain =
    assert.commandWorked(coll.explain('executionStats').count({a: true, b: {$in: [true, false]}}));
```

puppyofkosh (changes requested):
Hello Sebastian, thank you for your pull request. I reviewed the changes and they look great! All of my comments are relatively minor nitpicks. Most of them are just around our coding style.

If you like, I can deal with the comments and then merge the patch under your name. If you prefer to make the changes yourself that's okay too.

Also, you will have to sign the [Contributor's Agreement](https://www.mongodb.com/legal/contributor-agreement) as part of our considering the PR. 

IB

jstests/sharding/count_queryplan.js
```diff
@@ -0,0 +1,77 @@
...
+assertCountScan(explain);
+
+// We're not checking shard key for equality, therefore need a sharding filter.
+explain = assert.commandWorked(coll.explain('executionStats').count({a: {$in: [true, false]}}));
+assertShardFilter(explain);
+
+// We're requesting a disjoint key from shardkey, therefore need a sharding filter.
+explain = assert.commandWorked(coll.explain('executionStats').count({b: true}));
+assertShardFilter(explain);
+
```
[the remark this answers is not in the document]
Sebmaster: Shall we use `assertNoShardFilter` for the count queries as well then, or keep checking for COUNT_SCAN exactly?

Sebmaster:
> Also, you will have to sign the Contributor's Agreement as part of our considering the PR.

I signed the CLA already, at least I have the confirmation email around. Is it tied to the commit email or GH username, because only the username matches.

I'm working on the style changes, I think I may have used an older file as a template. Might be good to add an example for a file that matches the latest expectations to the wiki/contributor guidelines 😅

Sebmaster:
Turns out the finds are producing query plans with FETCH, so I made the SHARDING_FILTER check accept both IXSCAN and FETCH.

Also took me a while to get the mongo test setup back, but these changes are passing for me again.

jstests/sharding/count_queryplan.js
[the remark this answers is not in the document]
puppyofkosh: What you did looks good.

puppyofkosh:
> I signed the CLA already, at least I have the confirmation email around. Is it tied to the commit email or GH username? , Only the username matches in this case.

Ok, got it. I think just the github username is what matters.

> I'm working on the style changes, I think I may have used an older file as a template. Might be good to add an example for a file that matches the latest expectations to the wiki/contributor guidelines
Yeah, I'll look into this. The JS tests are a real mixed bag.

Anyway, with your latest changes this LGTM! Once I get confirmation you signed the contributor agreement (waiting to hear back from someone) and once https://jira.mongodb.org/browse/SERVER-49802 is done, I'll merge this and close the PR.

Sebmaster:
Nice! Thanks for the good review. I squashed the commits together to make it slightly prettier/easier to merge.

puppyofkosh:
Hi Sebastian, it seems like there's some trouble finding your name in the list of contributors. Do you know which email you used to sign originally? Alternatively, could you re-sign the agreement, probably with the email tied to the commit (smayr [at] atlassian).

Sebmaster:
Original email should've been `mongodb [at] smayr.name`.

puppyofkosh:
Thanks, they were able to find your name. Now we just have to wait for SERVER-50849 to go in.

puppyofkosh:
Alright, this made it through the commit queue. Congrats @Sebmaster and thank you for the contribution.

PR #1370, opened 2020-06-28
fix: aquire -> acquire typo

niftylettuce:


acmorrow:
Hi @niftylettuce -  Thank you for the PR. The PR changes a file that is internal to our vendored copy of MozJS, which is the Mozilla Javascript engine. The correct place to fix this typo is upstream with the MozJS maintainers. We try not to make direct edits to vendored third party code as it creates a maintenance burden for us to carry forward our local modifications when we need to upgrade the vendored copy. I'm closing this PR as I don't think we should merge it.

PR #1375, opened 2020-09-08
SERVER-48076: fixing the clang 10 compile error in bsonelement.h

androxylo:
This fixes the compile error: https://jira.mongodb.org/browse/SERVER-48076

The idea is to make the most non-intrusive change - the code works properly when the branch

`if (type() == NumberDouble) {`

is actually visited. The compile errors happen only when template T is specialized on something else than 'double', e.g. 'long long', but the code is not executed when T is 'long log'. So my choice here is to suppress the warning and leave the code untouched, with comments.

This is reproduced with clang 10 and ninja build.


androxylo:
I am no longer confident that this fix is appropriate based on the discussion in the bug and I'm closing it for now. More work is needed.

PR #1378, opened 2020-09-23
SERVER-47616: improve error message plus tests

shuvalov-mdb:
Trivial fix to improve error message according to the bug. It took me time though to find a test reproducing the problem, perhaps there is an easy way to run all tests but I had to cycle them from command line loop.


shuvalov-mdb:
This was merged using normal evergreen workflow.

PR #1386, opened 2020-12-29
FreeBSD update.

devnexen:
thread debug capabilities enabled and
native thread identifier fix.


kelly-cs:
Thanks for the PR, @devnexen. MongoDB isn't actively supporting FreeBSD right now. I am going to close this since it's a few years old now.

PR #1390, opened 2021-01-10
Fix build on FreeBSD/powerpc64le.

pkubaj:
Changes done:
1. Add registers for R01_sig and R32_sig on FreeBSD/powerpc64le.
2. Add mozjs config and source files.
3. Rename src/checksum/power8/crc32.sx to src/checksum/power8/crc32.S.
The reason is that Clang refuses to build .sx files and needs .S for assembly.
This should also help building with Clang on other operating systems.
4. Clang doesn't have ppc-asm.h header. When building with Clang, use the
bundled ppc-asm.h file from GCC (this is what upstream at https://github.com/antonblanchard/crc32-vpmsum does).
This should also help building with Clang on other operating systems.


edzh:

pkubaj:
https://jira.mongodb.org/browse/SERVER-53744

pkubaj:
I also signed CLA.

acmorrow:
Hi @pkubaj -

Thanks for the pull request. There are a few obstacles to getting this merged:

1) The source under src/third_party/wiredtiger is not edited directly in the server repository. Instead, such changes should be made to the upstream wiredtiger repository, which will then be synchronized into the server repository. So I suggest taking the wiredtiger parts of this there. I also seem to remember something subtle about all this suffix stuff with the assembler and clang. We've done stuff here before, so I'd like to move cautiously.

2) The direct changes to source files under MozJS have a similar issue. We don't directly modify them without creating an associated patch file under src/third_party/mozjs-60/patches. And, hopefully the change you are propsing here exists in upstream MozJS so that we can drop the patch when we do our next MozJS to a newer ESR (probably the first one that supports Apple Silicon).

3) If we were to merge the new configuration files for FreeBSD/powerpc64le, we would have no way to test them. This isn't necessarily a problem right now, but would become so if we ever needed to upgrade. As we don't see a great deal of commercial demand for that platform, it is difficult to justify investing resources in supporting it. It may be best if that part of the change remains something that lives in the FreeBSD ports tree.

Happy to discuss any of the above points further.

pkubaj:
1. https://github.com/wiredtiger/wiredtiger/pull/6220
2. I didn't modify MozJS files at all. I just generated it using `src/third_party/mozjs-60/gen-config.sh`, so I don't think there's anything special about maintaining those. `src/third_party/mozjs-60/platform/ppc64le/freebsd/build/js-confdefs.h` is actually the only file that's different:
```
--- ./linux/build/js-confdefs.h 2020-11-10 00:33:26.000000000 +0100
+++ ./freebsd/build/js-confdefs.h       2021-01-13 22:07:44.264215000 +0100
@@ -11,8 +11,6 @@
 #define CROSS_COMPILE
 #define EDITLINE 1
 #define HAVE_64BIT_BUILD 1
-#define HAVE_ALLOCA_H 1
-#define HAVE_BYTESWAP_H 1
 #define HAVE_CLOCK_MONOTONIC 1
 #define HAVE_CPP_DYNAMIC_CAST_TO_VOID_PTR 1
 #define HAVE_DIRENT_H 1
@@ -23,21 +21,14 @@
 #define HAVE_INTTYPES_H 1
 #define HAVE_LANGINFO_CODESET 1
 #define HAVE_LIBM 1
-#define HAVE_LINUX_IF_ADDR_H 1
-#define HAVE_LINUX_PERF_EVENT_H 1
-#define HAVE_LINUX_QUOTA_H 1
-#define HAVE_LINUX_RTNETLINK_H 1
 #define HAVE_LOCALECONV 1
 #define HAVE_LOCALTIME_R 1
-#define HAVE_MALLOC_H 1
 #define HAVE_MALLOC_USABLE_SIZE 1
-#define HAVE_MEMALIGN 1
 #define HAVE_NETINET_IN_H 1
 #define HAVE_NL_TYPES_H 1
 #define HAVE_POSIX_FADVISE 1
 #define HAVE_POSIX_FALLOCATE 1
 #define HAVE_POSIX_MEMALIGN 1
-#define HAVE_PTHREAD_GETNAME_NP 1
 #define HAVE_RES_NINIT 1
 #define HAVE_SINCOS 1
 #define HAVE_SSIZE_T 1
@@ -45,11 +36,8 @@
 #define HAVE_STRNDUP 1
 #define HAVE_SYS_MOUNT_H 1
 #define HAVE_SYS_QUEUE_H 1
-#define HAVE_SYS_QUOTA_H 1
-#define HAVE_SYS_STATFS_H 1
 #define HAVE_SYS_STATVFS_H 1
 #define HAVE_SYS_TYPES_H 1
-#define HAVE_SYS_VFS_H 1
 #define HAVE_THREAD_TLS_KEYWORD 1
 #define HAVE_TM_ZONE_TM_GMTOFF 1
 #define HAVE_UNISTD_H 1
@@ -64,8 +52,8 @@
 #define JS_POSIX_NSPR 1
 #define JS_PUNBOX64 1
 #define JS_STANDALONE 1
-#define MALLOC_H <malloc.h>
-#define MALLOC_USABLE_SIZE_CONST_PTR
+#define MALLOC_H <malloc_np.h>
+#define MALLOC_USABLE_SIZE_CONST_PTR const
 #define MOZILLA_UAVERSION "60.0"
 #define MOZILLA_VERSION "60.3.0"
 #define MOZILLA_VERSION_U 60.3.0
@@ -73,16 +61,14 @@
 #define MOZJS_MINOR_VERSION 3
 #define MOZ_BUILD_APP js
 #define MOZ_DLL_SUFFIX ".so"
-#define MOZ_GLUE_IN_PROGRAM 1
-#define MOZ_MEMORY 1
 #define MOZ_UPDATE_CHANNEL default
 #define NO_NSPR_10_SUPPORT 1
 #define RELEASE_OR_BETA 1
 #define STDC_HEADERS 1
 #define VA_COPY va_copy
-#define XP_LINUX 1
 #define XP_UNIX 1
 #define _REENTRANT 1
+#define _THREAD_SAFE 1

 #include "js/RequiredDefines.h"
```
3. I just prefer to keep the patches upstreamed, if possible to not carry this huge patchset (since 95% of that are just the MozJS files, same as for other platforms). Currently, the FreeBSD/powerpc64le port doesn't require any big change in mongodb. If this port will be something like Tier 2 or whatever (not tested upstream), that's fine with me, it will be tested by our package builders anyway.

acmorrow:
> 1. [wiredtiger/wiredtiger#6220](https://github.com/wiredtiger/wiredtiger/pull/6220)

Great. I dropped a comment in there, because as I mentioned before, something about this proposed change is making me itchy and I want to know why. We've been here before.


> 2. I didn't modify MozJS files at all. I just generated it using `src/third_party/mozjs-60/gen-config.sh`, so I don't think there's anything special about maintaining those. `src/third_party/mozjs-60/platform/ppc64le/freebsd/build/js-confdefs.h` is actually the only file that's different:

There appears to be a relevant change to `WasmSignalHandlers.cpp`, unless I've misunderstood.


> 
> 3. I just prefer to keep the patches upstreamed, if possible to not carry this huge patchset (since 95% of that are just the MozJS files, same as for other platforms). Currently, the FreeBSD/powerpc64le port doesn't require any big change in mongodb. If this port will be something like Tier 2 or whatever (not tested upstream), that's fine with me, it will be tested by our package builders anyway.

Sure. Let's focus on getting the other changes through in one form or another, and then we can think about what the best way to handle the generated configuration might be.

pkubaj:
On 21-01-14 06:41:42, Andrew C. Morrow wrote:
> > 1. [wiredtiger/wiredtiger#6220](https://github.com/wiredtiger/wiredtiger/pull/6220)
> 
> Great. I dropped a comment in there, because as I mentioned before, something about this proposed change is making me itchy and I want to know why. We've been here before.
> 
> 
> > 2. I didn't modify MozJS files at all. I just generated it using `src/third_party/mozjs-60/gen-config.sh`, so I don't think there's anything special about maintaining those. `src/third_party/mozjs-60/platform/ppc64le/freebsd/build/js-confdefs.h` is actually the only file that's different:
> 
> There appears to be a relevant change to `WasmSignalHandlers.cpp`, unless I've misunderstood.
Yes, there is. This is why I wrote 95% is autogenerated, not 100% :) My above comment was only about MozJS differences.

But it's a pretty simple change. Registers don't change every other week, so this change doesn't add a maintainance overhead. I guess this is the only change that is related to MongoDB itself (not MozJS or Wiretiger).

acmorrow (changes requested):
@pkubaj - Thanks for separating the WT stuff into another PR. Could you please undo the WT relevant changes in this review, and migrate the direct MozJS edit into a patch file? Then I think we could merge with the patch file and `gen_config.sh` changes, and you could keep the generated configurations in your tree so that you can update them when necessary. Does that seem like a reasonable compromise?

pkubaj:
I think I changed everything that was requested.

src/third_party/mozjs-60/platform/ppc64le/freebsd/build/Unified_cpp_js_src0.cpp
```diff
@@ -0,0 +1,55 @@
+#define MOZ_UNIFIED_BUILD
```
acmorrow, line 1: Let's remove the all generated sources from this review. I think it is better if you hold them on your side, since we have no way to regenerate this when we need to.

src/third_party/wiredtiger/src/checksum/power8/crc32.sx
```diff
@@ -26,7 +26,11 @@
  * as published by the Free Software Foundation; either version
```
acmorrow, line 26: This needs to get reverted too, should go in via a WT PR.

acmorrow (changes requested):
Basically ready to go once you take out the generated files and the direct WT modification.

acmorrow (approved):

acmorrow:
Thanks @pkubaj, this looks good, I'll get it merged.

src/third_party/mozjs-60/patches/freebsd-powerpc64le-fix.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/src/third_party/mozjs-60/extract/js/src/wasm/WasmSignalHandlers.cpp b/src/third_party/mozjs-60/extract/js/src/wasm/WasmSignalHandlers.cpp
+index b35461579c..0cd89e301d 100644
+--- a/src/third_party/mozjs-60/extract/js/src/wasm/WasmSignalHandlers.cpp
++++ b/src/third_party/mozjs-60/extract/js/src/wasm/WasmSignalHandlers.cpp
+@@ -263,6 +263,10 @@ struct AutoSignalHandler
+ #  define EPC_sig(p) ((p)->uc_mcontext.mc_pc)
+ #  define RFP_sig(p) ((p)->uc_mcontext.mc_regs[30])
+ # endif
++# if defined(__FreeBSD__) && defined(__powerpc64__)
++#  define R01_sig(p) ((p)->uc_mcontext.mc_frame[1])
```
acmorrow, line 10: @pkubaj - Actually, taking a closer look at this, and two small questions:

Upstream has this for  `__FreeBSD__` on `ppc64le`: https://searchfox.org/mozilla-central/rev/40205f1b55f1ba8a26fcbd362c20d843e8718a87/js/src/wasm/WasmSignalHandlers.cpp#212-215

Things that look a little different to me are:
 - They look for different defines for the architecture than you do.
 - Instead of accessing `mc_frame[1]` for `R01` they do `mc_gpr[1]`. This file makes me think they are synonyms though: http://web.mit.edu/freebsd/head/sys/powerpc/include/ucontext.h.

Do you think it would be better to stick a little closer to exactly what upstream has here?

acmorrow (changes requested):

acmorrow:
@pkubaj  - Please let me know on those questions before making any changes. I'm pretty happy to merge this as if the answers don't suggest the need for further changes.

src/third_party/mozjs-60/patches/freebsd-powerpc64le-fix.patch
pkubaj: From `/usr/include/machine/ucontext.h`:
```
/* GPRs and supervisor-level regs */
#define mc_gpr          mc_frame
#define mc_lr           mc_frame[32]
#define mc_cr           mc_frame[33]
#define mc_xer          mc_frame[34]
#define mc_ctr          mc_frame[35]
#define mc_srr0         mc_frame[36]
#define mc_srr1         mc_frame[37]
#define mc_exc          mc_frame[38]
#define mc_dar          mc_frame[39]
#define mc_dsisr        mc_frame[40]
```

So `mc_frame[1]` and `mc_gpr[1]` are actually exactly the same. If you feel like `mc_gpr[1]` is better, I will obviously change it, but it won't change the behaviour of mozjs.

acmorrow: Yeah, I agree they are the same. My question was more about whether you wanted to stick closer to what upstream MozJS has, which is different than yours. But it probably doesn't matter. When we get around to ugprading MozJS, your patch will drop out anyway since it won't be needed. So it probably doesn't matter. I'm OK to merge this with you confirming that you agree the code means the same thing.

acmorrow (approved):

PR #1400, opened 2021-03-18
SERVER-53566: Fix "opCtx != nullptr && _opCtx == nullptr" invariant

bozaro:
Check and skip already used `opId` to avoid `_nextOpId` wraparound issue.


bozaro:
Fixed by https://github.com/mongodb/mongo/commit/fbf13d24435cc05a04cbc56e560397e8b52d4ac9

PR #1404, opened 2021-05-26
SERVER-57206 Reslove Compiler warnings in JSStringWrapper::JSStringWr…

YuanYingdong:
Resolve compiler warnings 
```
In file included from /usr/include/string.h:494,
                 from /usr/include/c++/9/cstring:42,
                 from src/third_party/mozjs-60/include/mozilla/Span.h:33,
                 from src/third_party/mozjs-60/include/mozilla/Range.h:12,
                 from src/third_party/mozjs-60/include/jsapi.h:15,
                 from src/mongo/scripting/mozjs/jsstringwrapper.h:33,
                 from src/mongo/scripting/mozjs/jsstringwrapper.cpp:32:
In function 'char* strncpy(char*, const char*, size_t)',
    inlined from 'mongo::mozjs::JSStringWrapper::JSStringWrapper(int32_t)' at src/mongo/scripting/mozjs/jsstringwrapper.cpp:48:12:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:106:34: error: 'char* __builtin_strncpy(char*, const char*, long unsigned int)' specified bound 64 equals destination size [-Werror=stringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

### DIFF
--- a/src/mongo/scripting/mozjs/jsstringwrapper.cpp
+++ b/src/mongo/scripting/mozjs/jsstringwrapper.cpp
@@ -45,7 +45,8 @@ namespace mozjs {
 JSStringWrapper::JSStringWrapper(std::int32_t value) : _isSet(true) {
     auto formatted = fmt::format_int(value);
     _length = formatted.size();
-    strncpy(_buf, formatted.c_str(), sizeof(_buf));
+    strncpy(_buf, formatted.c_str(), sizeof(_buf) - 1);
+    _buf[sizeof(_buf) - 1] = '\0';
 }
 
 JSStringWrapper::JSStringWrapper(JSContext* cx, JSString* str) : _isSet(true) {


src/mongo/scripting/mozjs/jsstringwrapper.cpp
```diff
@@ -45,7 +45,8 @@ namespace mozjs {
 JSStringWrapper::JSStringWrapper(std::int32_t value) : _isSet(true) {
     auto formatted = fmt::format_int(value);
     _length = formatted.size();
-    strncpy(_buf, formatted.c_str(), sizeof(_buf));
+    memset(_buf, 0, sizeof(_buf));
```
acmorrow, line 48: Instead of doing a `memset` here, why not just do `_buf[sizeof(buf)-1] = '\0'` after calling `strncpy`?

YuanYingdong: Updated

acmorrow:
@YuanYingdong - Can you confirm that with the suggested edit that the warning/error no longer appears? If so, I'll work to get this merged.

acmorrow:
As a note, it might be good to include a static assertion that the size of `_buf` is sufficient to avoid truncation via the `strncpy`. The current code doesn't do that, but I think it would be as simple as writing:

```
static_assert(sizeof(_buf) > (std::numeric_limits<decltype(value)>::digits10 + 2));
```

You need the +2 to account for the sign character and the round-down behavior of `digits10`. The inequality ensures a byte remains for the null terminator.

YuanYingdong:
> @YuanYingdong - Can you confirm that with the suggested edit that the warning/error no longer appears? If so, I'll work to get this merged.

Yes, the warning no longer appears after the modification. The os I use is ubuntu 18.04LTS and GCC version `g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0`

acmorrow:
@YuanYingdong - Thanks for confirming, I'll get this merged today.

PR #1405, opened 2021-06-01
Use explicit typecast to size_t

kraj:
maxMemoryUsageBytes is size_t type which may not match long long value
internalDocumentSourceGroupMaxMemoryBytes.load() returns, so typecast it
to avoid narrowing warning from clang

document_source_group.cpp:378:22: error: non-constant-expression cannot be narrowed from type 'long long' to 'size_t' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]
                     maxMemoryUsageBytes ? *maxMemoryUsageBytes
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Upstream-Status: Pending
Signed-off-by: Khem Raj <raj.khem@gmail.com>


BillyDonahue:
Thanks for the heads-up. I've taken a slightly larger fix for this issue because the narrowing was actually happening in 2 places. Only one of these could emit a warning, but they were both suspicious.

It's being tracked by our Jira database, so I'll close this PR. Thanks again.
https://jira.mongodb.org/browse/SERVER-57329

PR #1410, opened 2021-08-03
Blank Field issue

NiteshKoushik:
Hello Team,

Not sure this is a right category to raise a bug request - apologies if it not proper category.

Documents in MongoDB is allowing to store blank fields and how this makes sense having blank fields.
Examples : 
db.collection.insert({'Name' : 'Nitesh', '' : '' })

Concern on the 2nd field which is inserted empty string at both place. I feel MongoDb should not allow blank field.

Regards,
Nitesh


edzh:
Hi Nitesh,

Can you please create a new SERVER ticket on [JIRA](https://jira.mongodb.org/) with the description of this behavior and steps to reproduce this issue?

Best,
Edwin

PR #1422, opened 2021-09-29
SERVER-60097 Add method to wait for topology state

mbroadst:
The TenantSplitDonorService needs to wait for three non-voting
replica set nodes to complete an initial sync before the reconfig
which will establish them as the recipient slice. This patch
introduces a new method `awaitTopologyState` which allows us to
wait for the local replica set to meet some provided condition.
Additionally, a new "waiter list" was introduced since the
existing `WaiterList` is primarily focused on waiting for opTimes
which is unrelated to topology state changes.

### DIFF
--- a/src/mongo/db/repl/replication_coordinator_impl.cpp
+++ b/src/mongo/db/repl/replication_coordinator_impl.cpp
@@ -107,7 +107,9 @@
 #include "mongo/rpc/metadata/repl_set_metadata.h"
 #include "mongo/transport/hello_metrics.h"
 #include "mongo/util/assert_util.h"
+#include "mongo/util/cancellation.h"
 #include "mongo/util/fail_point.h"
+#include "mongo/util/future_util.h"
 #include "mongo/util/scopeguard.h"
 #include "mongo/util/stacktrace.h"
 #include "mongo/util/testing_proctor.h"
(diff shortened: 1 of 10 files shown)


src/mongo/db/repl/replication_coordinator_impl.cpp
```diff
@@ -4172,6 +4198,29 @@ void ReplicationCoordinatorImpl::_updateWriteAbilityFromTopologyCoordinator(
     _readWriteAbility->setCanAcceptNonLocalWrites(lk, opCtx, canAcceptWrites);
 }
 
+std::vector<ReplicationCoordinator::NodeInfo> ReplicationCoordinatorImpl::getNodeInfoList() {
```
mbroadst, line 4201: Neither the `ReplicationCoordinator` nor the `TopologyCoordinator` have a data structure which tracks the state + configuration of every node in the replica set, which is what the `TenantSplitDonorService` needs to wait for 3 non-voting (config) nodes to finish initial sync (state). While this solution is somewhat inefficient in that a new list of `NodeInfo` is created on each call to this method, I tried to make the API opaque to a future improvement which returns cached information.

mbroadst:
wrong repo

PR #1427, opened 2021-11-02
SERVER-61202 remove timing dependency in cluster_ip_allowlist_replset

brushless-glitch:
Getting a failure on MacOS DEBUG. Unclear why, but it seems to be a timing issue. In original test a replication is initiated, while another thread waits 15 sec, during which access denied log entries are supposed to appear, and it does not happen on https://jira.mongodb.org/browse/BF-23201.

This patch refactors test to wait as long as necessary, but make sure that access denied happens.

EVG: https://spruce.mongodb.com/version/6181ad703e8e866d1df9feb1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/jstests/auth/cluster_ip_allowlist_replset.js
+++ b/jstests/auth/cluster_ip_allowlist_replset.js
@@ -25,25 +25,22 @@ admin.auth('admin', 'admin');
 assert.commandWorked(primary.adminCommand(
     {setDefaultRWConcern: 1, defaultWriteConcern: {w: "majority"}, writeConcern: {w: "majority"}}));
 
-jsTest.log("ReplSet started, will block __system connections and expect error");
-
-assert.commandWorked(
-    primary.adminCommand({setParameter: 1, "clusterIpSourceAllowlist": ["192.0.2.1"]}));
-
 function resetClusterIpSourceAllowlist(host) {
     jsTest.log("resetClusterIpSourceAllowlist: started");
-    sleep(15000);
     const mongo = new Mongo(host);
     const admin = mongo.getDB("admin");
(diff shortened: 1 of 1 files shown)


jstests/auth/cluster_ip_allowlist_replset.js
```diff
@@ -25,38 +25,38 @@ admin.auth('admin', 'admin');
...
 // successfully resets IP for connection
+jsTest.log("ReplSet started, will block __system connections and expect error");
+assert.commandWorked(
+    primary.adminCommand({setParameter: 1, "clusterIpSourceAllowlist": ["192.0.2.1"]}));
 assert.commandWorked(admin.runCommand({replSetReconfig: conf}));
 thread.join();
 
 jsTest.log("Verify that connections were denied");
-checkLog.containsJson(admin, 20240, {}, 15000);
+checkLog.containsJson(admin, 20240, {}, 1);
```
shreyaskalyan, line 59: Curious if this is even necessary in the end, since the threaded function also has this call. Can an assert failure from a thread not trigger a test failure?

shreyaskalyan (approved):
Just a question, otherwise looks good.

jstests/auth/cluster_ip_allowlist_replset.js
brushless-glitch: I verified that an error does propagate, but it looks like threading code in tests is not very mature and the way it works could [inadvertently] change in future. This line is to ensure test fails even if an assert does not propagate from `thread.join()`

shreyaskalyan (approved):
Approval stands.

PR #1428, opened 2021-11-05
SERVER-61494 Fix "mongod --shutdown" to check size of "mongod.lock" instead of existence

tianon:
In a07dc78e49c53de539dff92748334cd7fbe2e2a9, the code for `mongod --shutdown` was refactored, and appears to have inadvertently changed from `while (boost::filesystem::exists(procPath))` to `while (boost::filesystem::exists(pidfile))`, which never becomes false as the `mongod.lock` file this new version is waiting for is only ever emptied, not deleted.

This fixes that to wait for the file to be either deleted *or* emptied.  I debated switching back to `procPath` like the previous code, but I figured maybe it was changed on purpose?  I'm not sure how that previous change was tested, though.

The difference between the behavior currently in 5.0 and this new implementation is that the old `--shutdown` would exit even if the server exited uncleanly (because it was waiting for the PID to no longer exist) while the new implementation would instead hang forever (which is what happens on the 5.1 RCs when you run `mongod --shutdown` whether the server shuts down or not, and how I discovered this issue).

I'll add my standard disclaimer that I'm happy to adjust, rebase, amend, etc this change in any way. :smile:

(I've already signed the CLA, so hopefully that carries forward here and I don't get slapped by a bot :crossed_fingers:)

### DIFF
--- a/src/mongo/db/mongod_main.cpp
+++ b/src/mongo/db/mongod_main.cpp
@@ -927,7 +927,8 @@ Status shutdownProcessByDBPathPidFile(const std::string& dbpath) {
                 str::stream() << "Failed to kill process: " << errnoWithDescription(e)};
     }
 
-    while (boost::filesystem::exists(pidfile)) {
+    uintmax_t size;
+    while ((size = boost::filesystem::file_size(pidfile)) != 0 && size != static_cast<uintmax_t>(-1)) {
         sleepsecs(1);
     }
 


src/mongo/db/mongod_main.cpp
```diff
@@ -909,7 +909,8 @@ Status shutdownProcessByDBPathPidFile(const std::string& dbpath) {
                 str::stream() << "Failed to kill process: " << errnoWithDescription(e)};
     }
 
-    while (boost::filesystem::exists(pidfile)) {
+    uintmax_t size;
+    while ((size = boost::filesystem::file_size(pidfile)) != 0 && size != static_cast<uintmax_t>(-1)) {
```
tianon, line 913: I wanted to just use `boost::filesystem::is_empty` here, but it's implemented as `file_size(...) == 0`, and https://www.boost.org/doc/libs/1_68_0/libs/filesystem/doc/reference.html#file_size says that `file_size` will return `-1` casted to be an unsigned integer when the file doesn't exist or isn't a regular file, so I ended up with this blurb (to avoid the code calling `stat` more than once per loop unnecessarily).

tianon:
Friendly ping? :pray: :sweat_smile:

(cc @sgolemon ? :see_no_evil:)

edzh:
Hi @tianon,
Thanks for making this pull request! We just need you to create a new SERVER ticket on [JIRA](https://jira.mongodb.org/) to help us track the PR.

tianon:
Sure: https://jira.mongodb.org/browse/SERVER-61494

sgolemon:
@tianon Sorry for the delay. Holidays and all. This should be merged shortly (tests and such)

PR #1435, opened 2021-12-18
SERVER-61959 Add MSVS 2022 support to vendored SCons

bdbaddog:
Backported MSVS2022 support from SCons 4.3.0


acmorrow:
I'm testing this locally, and so far it looks good. Is there a way to get the merge commit out of the PR though?

acmorrow:
@bdbaddog - Confirmed, this works correctly:
- If I don't have it, I can't build with `MSVC_VERSION=14.3`
- If I do have it, I can build with `MSVC_VERSION=14.3`
- If I add enforcement to the server build for requiring VS 2022, building with `MSVC_VERSION=14.2` fails, but building with this patch and `MSVC_VERSION=14.3` it passes.

acmorrow (approved):
LGTM, mod removing the merge commit somehow. Or, if that proves difficult, I can probably fix it up on my side.

PR #1438, opened 2022-01-03
[DRAFT] make WriteConcernOptions immutable

mbroadst:
This is a large change set over many files, so I'm putting it up as a draft first to see if we agree on the approach. Broadly, the PR is split up into two commits: one to deduplicate the parsing logic in `WriteConcernOptions`, and a second which makes `WriteConcernOptions` an immutable class (as much as possible). I've left comments on the PR on all of the interesting changes, but most of the noise here are simple text changes converting direct member access into calls to an accessor.

patch build: https://spruce.mongodb.com/version/61d33b4d2fbabe54a46ae96c

### DIFF
--- a/src/mongo/db/repl/replication_coordinator_impl_test.cpp
+++ b/src/mongo/db/repl/replication_coordinator_impl_test.cpp
@@ -622,9 +622,8 @@ TEST_F(ReplCoordTest, NodeReturnsImmediatelyWhenAwaitReplicationIsRanAgainstASta
 
     OpTimeWithTermOne time(100, 1);
 
-    WriteConcernOptions writeConcern;
-    writeConcern.wTimeout = WriteConcernOptions::kNoWaiting;
-    writeConcern.wNumNodes = 2;
+    WriteConcernOptions writeConcern(
+        2, WriteConcernOptions::SyncMode::UNSET, WriteConcernOptions::kNoWaiting);
 
     // Because we didn't set ReplSettings.replSet, it will think we're a standalone so
     // awaitReplication will always work.
@@ -652,10 +651,8 @@ TEST_F(ReplCoordTest, NodeReturnsNotPrimaryWhenRunningAwaitReplicationAgainstASe
 
     OpTimeWithTermOne time(100, 1);
 
-    WriteConcernOptions writeConcern;
-    writeConcern.wTimeout = WriteConcernOptions::kNoWaiting;
-    writeConcern.wNumNodes = 0;  // Waiting for 0 nodes always works
-    writeConcern.wMode = "";
+    WriteConcernOptions writeConcern(
+        0, WriteConcernOptions::SyncMode::UNSET, WriteConcernOptions::kNoWaiting);
 
     // Node should fail to awaitReplication when not primary.
     ReplicationCoordinator::StatusAndDuration statusAndDur =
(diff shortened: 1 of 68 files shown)


src/mongo/db/write_concern_options.h
```diff
@@ -127,52 +121,92 @@ struct WriteConcernOptions {
...
+        return _wTimeout;
+    }
+
+    Date_t wDeadline() const {
+        return _wDeadline;
+    }
+
+    // TODO: this is only used in one test
+    // `NodeReturnsWriteConcernFailedWhenAWriteConcernTimesOutBeforeBeingSatisified`
+    void setDeadline_forTest(const Date_t& deadline) {
```
mbroadst, line 184: This is only used in [one test](https://github.com/mongodb/mongo/blob/b7181e6813fbc771f4225d8161b4bd7302630c5c/src/mongo/db/repl/replication_coordinator_impl_test.cpp#L1288-L1331). I think we should probably outright remove the concept of `wDeadline` since its only used in that test, but also think that should be follow-on work.

src/mongo/db/write_concern_options.h
```diff
@@ -127,52 +121,92 @@ struct WriteConcernOptions {
...
+    }
 
-    // Timeout in milliseconds.
-    int wTimeout;
+    CheckCondition checkCondition() const {
+        return _checkCondition;
+    }
+
+private:
+    friend class repl::ReplicationCoordinatorImpl;
```
mbroadst, line 193: The `ReplicationCoordinator` does serious surgery to `WriteConcernOptions`
  - [populating unset syncModes](https://github.com/mongodb/mongo/blob/b7181e6813fbc771f4225d8161b4bd7302630c5c/src/mongo/db/repl/replication_coordinator_impl.cpp#L5892-L5904)
  - [setting a checkCondition](https://github.com/mongodb/mongo/blob/b7181e6813fbc771f4225d8161b4bd7302630c5c/src/mongo/db/repl/replication_coordinator_impl.cpp#L3234-L3240)

These aren't changes that are easy to support through API (should we add `checkCondition` to  a constructor, or support it in IDL?), so I made the decision to only allow the `ReplicationCoordinator` to mess with the internals of a write concern. I think we should also reconsider this, but in some future ticket.

src/mongo/db/repl/replication_coordinator_impl_test.cpp
```diff
@@ -5511,21 +5494,14 @@ TEST_F(
...
-    writeConcern.wNumNodes = 3;
+    WriteConcernOptions writeConcern(
+        3, WriteConcernOptions::SyncMode::UNSET, WriteConcernOptions::kNoTimeout);
 
     ReplicationAwaiter awaiter(getReplCoord(), getServiceContext());
     awaiter.setOpTime(time);
     awaiter.setWriteConcern(writeConcern);
     awaiter.start();
 
-    ReplicationAwaiter awaiterJournaled(getReplCoord(), getServiceContext());
```
mbroadst, line 5523: This awaiter originally returned an error because the `WriteConcernOptions` above was reused, which is a bug. In `ReplSetConfig::checkIfWriteConcernCanBeSatisfied` if a `wMode` of "majority" is used then we check the `wNumNodes` of the write concern to determine if there are enough nodes to satisfy the write concern. A properly constructed majority `WriteConcernOptions` has `wNumNodes: 0` so this code would now just block indefinitely waiting for an actual majority. I chose to remove the test since I didn't think it was actually testing the correct functionality

src/mongo/db/write_concern_options_test.cpp
```diff
@@ -123,76 +120,62 @@ TEST(WriteConcernOptionsTest, ParseSetsWNumNodesIfWIsANumber) {
     auto sw = WriteConcernOptions::parse(BSON("w" << 3));
     ASSERT_OK(sw.getStatus());
     WriteConcernOptions options = sw.getValue();
-    ASSERT_TRUE(WriteConcernOptions::SyncMode::UNSET == options.syncMode);
-    ASSERT_EQUALS(3, options.wNumNodes);
-    ASSERT_EQUALS("", options.wMode);
-    ASSERT_EQUALS(0, options.wTimeout);
-}
-
-TEST(WriteConcernOptionsTest, ParseSetsWTimeoutToZeroIfWTimeoutIsNotANumber) {
```
mbroadst, line 132: This represents a behavioral change: we used to coerce a string value for `wTimeout` into 0, and now we will report an error. It's not very difficult to support the old behavior, but upon reflection it seems like this would always have been user error in practice so I wanted to leave the door open for fixing that if possible.

src/mongo/db/write_concern_options.h
mbroadst: Billy thinks that we should have deprecated mutators that the ReplicationCoordinator* classes can use (and document them as deprecated in comments) instead of giving them free reign with a friend declaration.

PR #1439, opened 2022-01-03
SERVER-62333 use std::ptrdiff_t to avoid UB in MurmurHash3

mbroadst:
Ensure we are using `std::ptrdiff_t` to prevent undefined behavior in the `getblock`/`putblock` implementations,.


BillyDonahue (approved):

mbroadst:
evergreen merge

PR #1448, opened 2022-02-17
add compatible GPL-V3 license

libererer:


Schubes:
Hi @libererer,

Our licensing is an intentional choice. You can read more about it here: https://www.mongodb.com/licensing/server-side-public-license

PR #1449, opened 2022-02-21
WIP Change the network chunk version format to the new format

m4nti5:
This is a WIP test that changes the chunk version format from an array format:

`[<major|minor>, <epoch>, <timestamp>]`

To an object like format:

`{v: <major|minor>, e: <epoch>, t: <timestamp>}`

There are several tests that need to be fixed, and some extra code necessary to fully complete the transition, one of the major problems seems to be split chunk, the full evergreen patch can be found [here](https://spruce.mongodb.com/version/6213afd9c9ec4477ac9e8f53/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).


afuschetto:
Leaving this as a reviewer since no code review is no longer necessary.

PR #1450, opened 2022-02-22
SERVER-61165 update logkeeper fcv to 5.3

jonstreets:


vrachev (approved):
LGTM on the change but can you run a patch as well with the initial sync tasks?

PR #1451, opened 2022-02-24
change IDL gen, remove security token from opCtx

jannaerin:
In this patch:
- Added a concatenate_with_tenant_id_and_db namespace type for IDL defined commands
- Removed securityToken from opCtx and put on the request object

Not in this patch:
- Changes to logging
- Changes to audit

Example of command parse method (I realized it might be nice to see the generated code):
```
CreateCommand CreateCommand::parse(const IDLParserErrorContext& ctxt, const OpMsgRequest& request) {
    NamespaceString localNS;
    CreateCommand object(localNS);
    object.parseProtected(ctxt, request);
    return object;
}
void CreateCommand::parseProtected(const IDLParserErrorContext& ctxt, const OpMsgRequest& request) {
    std::bitset<23> usedFields;
    const size_t kCappedBit = 0;
    const size_t kAutoIndexIdBit = 1;
    const size_t kIdIndexBit = 2;
    const size_t kSizeBit = 3;
    const size_t kMaxBit = 4;
    const size_t kStorageEngineBit = 5;
    const size_t kValidatorBit = 6;
    const size_t kValidationLevelBit = 7;
    const size_t kValidationActionBit = 8;
    const size_t kIndexOptionDefaultsBit = 9;
    const size_t kViewOnBit = 10;
    const size_t kPipelineBit = 11;
    const size_t kCollationBit = 12;
    const size_t kRecordPreImagesBit = 13;
    const size_t kChangeStreamPreAndPostImagesBit = 14;
    const size_t kTimeseriesBit = 15;
    const size_t kClusteredIndexBit = 16;
    const size_t kExpireAfterSecondsBit = 17;
    const size_t kEncryptedFieldsBit = 18;
    const size_t kTempBit = 19;
    const size_t kFlagsBit = 20;
    const size_t kDbNameBit = 21;
    const size_t kDollarTenantIdBit = 22;
    BSONElement commandElement;
    bool firstFieldFound = false;

    for (const auto& element :request.body) {
        const auto fieldName = element.fieldNameStringData();

        if (firstFieldFound == false) {
            commandElement = element;
            firstFieldFound = true;
            continue;
        }

        if (fieldName == kCappedFieldName) {
            if (MONGO_likely(ctxt.checkAndAssertTypes(element, {Bool, NumberLong, NumberInt, NumberDecimal, NumberDouble}))) {
                if (MONGO_unlikely(usedFields[kCappedBit])) {
                    ctxt.throwDuplicateField(element);
                }

                usedFields.set(kCappedBit);

                _capped = element.trueValue();
            }
        }
       ******
       // continue setting all other fields, removing to be more concise
      *******
        else if (fieldName == kDbNameFieldName) {
            if (MONGO_likely(ctxt.checkAndAssertType(element, String))) {
                if (MONGO_unlikely(usedFields[kDbNameBit])) {
                    ctxt.throwDuplicateField(element);
                }

                usedFields.set(kDbNameBit);

                _hasDbName = true;
                _dbName = element.str();
            }
        }
        else if (fieldName == kDollarTenantIdFieldName) {
            if (MONGO_unlikely(usedFields[kDollarTenantIdBit])) {
                ctxt.throwDuplicateField(element);
            }

            usedFields.set(kDollarTenantIdBit);

            _dollarTenantId = TenantId::parseFromBSON(element);
        }
        else {
            if (!mongo::isGenericArgument(fieldName)) {
                ctxt.throwUnknownField(fieldName);
            }
        }
    }


    if (MONGO_unlikely(!usedFields.all())) {
        if (!usedFields[kDbNameBit]) {
            ctxt.throwMissingField(kDbNameFieldName);
        }
    }

    boost::optional<TenantId> tenantId;

    if (request.securityToken.nFields() > 0) {
        _securityToken.emplace(auth::SecurityToken::parse({"Security Token"}, request.securityToken));
        uassert(ErrorCodes::BadValue, "Security token authenticated user requires a valid Tenant ID", _securityToken->getAuthenticatedUser().getTenant());
        tenantId.emplace(*_securityToken->getAuthenticatedUser().getTenant());
    }

    if (!tenantId && _dollarTenantId) {
        tenantId.emplace(*_dollarTenantId);
    }
    invariant(_nss.isEmpty());

    _nss = ctxt.parseNSCollectionRequired(_dbName, commandElement, false, tenantId);
}
```

### DIFF
--- a/buildscripts/idl/idl/binder.py
+++ b/buildscripts/idl/idl/binder.py
@@ -403,11 +403,23 @@ def _inject_hidden_command_fields(command):
     db_field.cpp_name = "dbName"
     db_field.serialize_op_msg_request_only = True
 
+    # Inject a "$tenant" which we can decode during command parsing
+    dollar_tenantid_field = syntax.Field(command.file_name, command.line, command.column)
+    dollar_tenantid_field.name = "$tenant"
+    dollar_tenantid_field.type = syntax.FieldTypeSingle(command.file_name, command.line, command.column)
+    dollar_tenantid_field.type.type_name = "tenant_id"  # This comes from basic_types.idl
+    dollar_tenantid_field.cpp_name = "dollarTenantId"
+    dollar_tenantid_field.optional = True
+
     # Commands that require namespaces do not need to have db defaulted in the constructor
-    if command.namespace == common.COMMAND_NAMESPACE_CONCATENATE_WITH_DB:
+    if command.namespace == common.COMMAND_NAMESPACE_CONCATENATE_WITH_DB or command.namespace == common.COMMAND_NAMESPACE_CONCATENATE_WITH_TENANTID_AND_DB:
         db_field.constructed = True
+    
+    if command.namespace == common.COMMAND_NAMESPACE_CONCATENATE_WITH_TENANTID_AND_DB:
+        dollar_tenantid_field.constructed = True
 
     command.fields.append(db_field)
+    command.fields.append(dollar_tenantid_field)
 
 
 def _bind_struct_type(struct):
--- a/src/mongo/db/auth/authorization_session.h
+++ b/src/mongo/db/auth/authorization_session.h
@@ -145,7 +145,7 @@ public:
      * Adds the User identified by "UserName" to the authorization session, acquiring privileges
      * for it in the process.
      */
-    virtual Status addAndAuthorizeUser(OperationContext* opCtx, const UserName& userName) = 0;
+    virtual Status addAndAuthorizeUser(OperationContext* opCtx, const UserName& userName, bool fromSecurityToken = false) = 0;
 
     // Returns the authenticated user with the given name.  Returns NULL
     // if no such user is found.
--- a/src/mongo/db/namespace_string.h
+++ b/src/mongo/db/namespace_string.h
@@ -39,6 +39,7 @@
 #include "mongo/bson/util/builder.h"
 #include "mongo/db/repl/optime.h"
 #include "mongo/db/server_options.h"
+#include "mongo/idl/tenant_id.h"
 #include "mongo/logv2/log_attr.h"
 #include "mongo/util/assert_util.h"
 #include "mongo/util/uuid.h"
@@ -212,6 +213,7 @@ public:
     explicit NamespaceString(StringData ns) {
         _ns = ns.toString();  // copy to our buffer
         _dotIndex = _ns.find('.');
+        _tenantIndex = std::string::npos;
         uassert(ErrorCodes::InvalidNamespace,
                 "namespaces cannot have embedded null characters",
                 _ns.find('\0') == std::string::npos);
@@ -235,9 +237,39 @@ public:
         ++it;
         it = std::copy(collectionName.begin(), collectionName.end(), it);
         _dotIndex = dbName.size();
+        _tenantIndex = std::string::npos;
 
         dassert(it == _ns.end());
         dassert(_ns[_dotIndex] == '.');
+
+        uassert(ErrorCodes::InvalidNamespace,
+                "namespaces cannot have embedded null characters",
+                _ns.find('\0') == std::string::npos);
+    }
+
+    NamespaceString(TenantId tenantId, StringData dbName, StringData collectionName)
+           : _ns(tenantId.toString().size() + dbName.size() + collectionName.size() + 2, '\0') {
+        uassert(ErrorCodes::InvalidNamespace,
+                "'.' is an invalid character in the database name: " + dbName,
+                dbName.find('.') == std::string::npos);
+        uassert(ErrorCodes::InvalidNamespace,
+                "Collection names cannot start with '.': " + collectionName,
+                collectionName.empty() || collectionName[0] != '.');
+
+        std::string::iterator it = std::copy(tenantId.toString().begin(), tenantId.toString().end(), _ns.begin());
+        *it = '_';
+        ++it;
+        it = std::copy(dbName.begin(), dbName.end(), it);
+        _tenantIndex = tenantId.toString().size();
+
+        *it = '.';
+        ++it;
+        it = std::copy(collectionName.begin(), collectionName.end(), it);
+        _dotIndex = tenantId.toString().size() + dbName.size() + 1;
+
+        dassert(it == _ns.end());
+        dassert(_ns[_dotIndex] == '.');
+        dassert(_ns[_tenantIndex] == '_');
 
         uassert(ErrorCodes::InvalidNamespace,
                 "namespaces cannot have embedded null characters",
@@ -275,6 +307,10 @@ public:
         return _dotIndex == std::string::npos
             ? StringData()
             : StringData(_ns.c_str() + _dotIndex + 1, _ns.size() - 1 - _dotIndex);
+    }
+
+    boost::optional<TenantId> tenantId() {
+        return _tenantIndex == std::string::npos ? boost::none : boost::make_optional<TenantId>(TenantId(OID::createFromString(StringData(_ns.data(), _tenantIndex))));
     }
 
     const std::string& ns() const {
@@ -577,6 +613,7 @@ public:
 private:
     std::string _ns;
     size_t _dotIndex = 0;
+    size_t _tenantIndex = 0;
 };
 
 /**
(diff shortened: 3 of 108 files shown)


src/mongo/db/service_entry_point_common.cpp
```diff
@@ -1311,11 +1312,16 @@ void ExecCommandDatabase::_initiateCommand() {
     });
 
     rpc::readRequestMetadata(opCtx, request, command->requiresAuth());
-    uassert(ErrorCodes::Unauthorized,
+    if (auto token = _invocation->securityToken()) {
+        verifySecurityToken(*token, request.securityToken);
+        _tokenAuthorizationSessionGuard.emplace(opCtx, token);
+    } else if (auto dollarTenant = _invocation->ns().tenantId()) {
+        verifyDollarTenantField(opCtx);
+    }
```
jannaerin, line 1320: I'm grabbing the security token from the command invocation itself, and then manually setting it on the scope guard rather than putting it on the opCtx. Any other ideas here are most welcome!

Also, from an auth/authz POV I think we need a way to identify a tenant id as having come from $tenant as opposed to having come from a db prefix (if we do still support tenant ids coming in as prefixes until we swap out Atlas Proxy for mongoq).

src/mongo/db/auth/authorization_session.h
```diff
@@ -145,7 +145,7 @@ class AuthorizationSession {
      * Adds the User identified by "UserName" to the authorization session, acquiring privileges
      * for it in the process.
      */
-    virtual Status addAndAuthorizeUser(OperationContext* opCtx, const UserName& userName) = 0;
+    virtual Status addAndAuthorizeUser(OperationContext* opCtx, const UserName& userName, bool fromSecurityToken = false) = 0;
```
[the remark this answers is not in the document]
sgolemon: Actually, just thought about this some more in context of my concerns about the signature validation not being performed reliably.

How about if we have two interface methods:

```
/**
 * Connection based authentication using `saslStart/saslContinue` or `authenticate`
 */
 virtual Status addAndAuthorizeUser(OperationContext*, const UserName&) = 0;
 
 /**
 * Security Token based authentication used with mutltitenancy support.
 */
virtual Status addAndAuthorizeUser(OperationContext*, const SecurityToken&) = 0;
```

Then the signature validation happens at the same time as authorization grants, which will dramatically reduce the chances that we fail at this.

sgolemon (changes requested):
Here's my initial thoughts, still working through finer details.

I'm also wondering if we can avoid putting the *entire* SecurityToken on the Command objects and keep it to ONLY the TenantID.  The authenticatedUser portion really only belongs to authorization processing.

Also, can you save me some effort and help me understand where logging and auditing are getting tenant ID from now? (sorry, I'm headed straight to another meeting and it's not leaping out at me)

jannaerin:
Thanks for the quick initial feedback! I actually didn't make changes to
either logging or audit in this patch - I'm curious:
1. Is it necessary/useful for the tenant id show up in every log line
and/or audit event? If we do take this approach then any log line/audit
event that contains a namespace will contain the tenant id as a part of the
namespace (or db name).
2. How does the tenantId show up in the audit log today? For process
logging, it will only show up in the namespace, if a namespace is logged
(as far as I'm aware).

I'll respond to your other comments later today as well.

sgolemon:
> 1. Is it necessary/useful for the tenant id show up in every log line
and/or audit event? If we do take this approach then any log line/audit
event that contains a namespace will contain the tenant id as a part of the
namespace (or db name).

I was very specifically asked to make this part of my work, yes.  Not all log messages contain namespaces, many don't.

> 2. How does the tenantId show up in the audit log today? For process
logging, it will only show up in the namespace, if a namespace is logged
(as far as I'm aware).

If an TenantID is active, then a new low-level field is included at the same nesting as "id", "severity", or "category".  Similar with auditing (where it's easier to comprehend due to the structure of both systems). https://github.com/10gen/mongo-enterprise-modules/blob/536a2e918d6fae3334b0ca6cb3d789914db73164/src/audit/audit_event.cpp#L72-L76

jannaerin:
1. Got it, I'll continue to think about it - maybe we do leave it on the
opCtx just for logging, I'm not sure the best way to handle this yet.
2. Right, we added that in the last number of months but prior to that
change did the tenant ID show up in audit logs (I guess by "today" I meant
today in Atlas Serverless)? Also, do serverless customers have access to
the audit logs?

EshaMaharishi:
2. Just a note that audit logs aren't supported in Serverless today.

sgolemon:
> Just a note that audit logs aren't supported in Serverless today.

True. During the scope, audit log support was a "nice to have", but it would be very good to continue supporting it.  I also feel like it's the easier support since we (almost) always have a valid `Client` passed along, and can pull out authorized security token user (including tenant) from the auth session.  System log is more complex because we *only* have global scope to work with there.

buildscripts/idl/idl/binder.py
```diff
@@ -403,11 +403,23 @@ def _inject_hidden_command_fields(command):
     db_field.cpp_name = "dbName"
     db_field.serialize_op_msg_request_only = True
 
+    # Inject a "$tenant" which we can decode during command parsing
```
markbenvenuto, line 406: So like `$db`, this is a field we will not check for on ALL commands?

src/mongo/idl/idl_parser.cpp
```diff
@@ -241,9 +241,16 @@ void IDLParserErrorContext::throwAPIStrictErrorIfApplicable(StringData fieldName
...
+void IDLParserErrorContext::throwDuplicateTenantIdErrorIfApplicable(boost::optional<TenantId> dollarTenant) const {
+    uassert(6223901,
+            str::stream() << "Cannot pass $tenant id if also passing securityToken  $tenant: " << dollarTenant,
+            !dollarTenant);
+}
+
 NamespaceString IDLParserErrorContext::parseNSCollectionRequired(StringData dbName,
                                                                  const BSONElement& element,
-                                                                 bool allowGlobalCollectionName) {
+                                                                 bool allowGlobalCollectionName,
```
markbenvenuto, line 252: As Sarah suggested, use an enum instead of a bool

src/mongo/db/namespace_string.h
```diff
@@ -277,6 +309,10 @@ class NamespaceString {
             : StringData(_ns.c_str() + _dotIndex + 1, _ns.size() - 1 - _dotIndex);
     }
 
+    boost::optional<TenantId> tenantId() {
```
markbenvenuto, line 312: This allocate heap memory on every call if tenantId exists. I wonder if you need to add a `hasTenantId` that returns bool that avoids the memory allocation if callers simply need to test for tenantId.

markbenvenuto (changes requested):
From the IDL perspective, the changes look good. For a real commit, you will need to add tests to idl_test.cpp/unittest.idl for your new command type.

src/mongo/db/auth/authorization_session.h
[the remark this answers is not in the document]
jannaerin: I'm tying to synthesize your comments in order to implement the changes, but I don't think I'm following exactly what you're suggesting - my current understanding is that we could make changes that look something like:
-  Change `verifySecurityToken(tokenObj)` to `parseSecurityToken(tokenObj)`. This calls SecurityToken::parse(), then validates the signature.
- Create some type of wrapper for SecurityToken and put that on the request object that looks something like:
```
class SecurityTokenWrapper {
public:
    SecurityTokenWrapper(const BSONObj& tokenObj) : _tokenObj(tokenObj) {}
 
    TenantId getTenantId() {
        if (!token) {
            _token = parseSecurityToken(_tokenObj);
        }
       return _token->getAuthenticatedUser().getTenant();
    }
   
    getAuthenticatedUser() { ... looks similar to getTenantId}
private:
    BSONObj _tokenObj;
    SecurityToken* _token;
}
```
- In command request parsers, the call to parseNSCollectionRequired would now be `ctxt.parseNSCollectionRequired(_dbName, commandElement, _securityTokenWrapper.getTenantId())`
- Implement the changes you mention in this comment to create two `addAndAuthorizeUser` methods.

The main thing I'm confused about is where we'd want to construct the `SecurityTokenAuthenticationGuard` and call `addAndAuthorizeUser` - for signature validation to happen at the same time as authz grants as you mentioned, we'd need to call `addAndAuthorizeUser` when we call` parseSecurityToken`. We'll already have constructed and parsed the actual SecurityToken object by the time we construct the SecurityTokenAuthenticationGuard and do authz grants though, because we access the tenantId when parsing the command, and I don't think we can move handling the scope guard/authz grants into command parsing.

PR #1456, opened 2022-03-10
SERVER-64287 Create sharding variant of fle2 test suite

markbenvenuto:


markbenvenuto:
This code review is spread across multiple repositories. Here are the other Pull Requests associated with the code review:
* [enterprise](https://github.com/10gen/mongo-enterprise-modules/pull/395)

jstests/fle2/libs/encrypted_client_util.js
```diff
@@ -0,0 +1,190 @@
...
+    assertOneEncryptedDocumentFields(coll, query, fields) {
+        let rawDocs = this._db.getCollection(coll).find(query).toArray();
+        assert.eq(rawDocs.length, 1);
+        let encryptedDocs = this._edb.getCollection(coll).find(query).toArray();
+        assert.eq(encryptedDocs.length, 1);
+
+        let rawDoc = rawDocs[0];
+        let encryptedDoc = encryptedDocs[0];
+
+        assert(rawDoc["__safeContent__"] !== undefined);
```
erwee, line 137: So `encryptedDoc` is the document after client-side decryption? and `rawDoc` contains the fields in their encrypted form? If so, the naming confused me at first.

jstests/fle2/libs/encrypted_client_util.js
```diff
@@ -0,0 +1,190 @@
...
+     * @param {string} name Name of EDC
+     * @param {number} edc Number of documents in EDC
+     * @param {number} esc Number of documents in ESC
+     * @param {number} ecc Number of documents in ECC
+     * @param {number} ecoc Number of documents in ECOC
+     */
+    assertEncryptedCollectionCounts(name, expectedEdc, expectedEsc, expectedEcc, expectedEcoc) {
+        const ci = this._edb.getCollectionInfos({"name": name})[0];
+
+        const ef = ci.options.encryptedFields;
```
erwee, line 97: should `ef` (and subfields) be checked if undefined before proceeding?

src/mongo/shell/db.js
```diff
@@ -1808,6 +1808,39 @@ DB.prototype.getSession = function() {
     return this._session;
 };
 })(Object.prototype.hasOwnProperty);
+
+DB.prototype.createEncryptedCollection = function(name, opts) {
+    const res = assert.commandWorked(this.createCollection(name, opts));
+
+    const ci = this.getCollectionInfos({"name": name})[0];
+
+    const ef = ci.options.encryptedFields;
```
erwee, line 1817: same comment as above re: checking undefined `ci` and `ef`

erwee (approved):
Just a couple of nits, but overall lgtm. I appreciate the refactor of the key vault code, I need that for creating compaction tokens.

src/mongo/shell/encrypted_dbclient_base.h
```diff
@@ -78,7 +79,9 @@ constexpr std::array<StringData, 11> kEncryptedCommands = {"aggregate"_sd,
                                                            "insert"_sd,
                                                            "update"_sd};
 
-class EncryptedDBClientBase : public DBClientBase, public mozjs::EncryptionCallbacks {
+class EncryptedDBClientBase : public DBClientBase,
+                              public mozjs::EncryptionCallbacks,
+                              public FLEKeyVault {
```
brushless-glitch, line 84: are we allowing multiple inheritance from non-interface classes?

brushless-glitch (approved):

src/mongo/shell/encrypted_dbclient_base.cpp
```diff
@@ -677,10 +675,25 @@ std::shared_ptr<SymmetricKey> EncryptedDBClientBase::getDataKeyFromDisk(const UU
...
+    return decryptedKey;
+}
+
+std::shared_ptr<SymmetricKey> EncryptedDBClientBase::getDataKeyFromDisk(const UUID& uuid) {
+    auto decryptedKey = getKeyMaterialFromDisk(uuid);
     return std::make_shared<SymmetricKey>(
         std::move(decryptedKey), crypto::aesAlgorithm, "kms_encryption");
 }
 
+KeyMaterial EncryptedDBClientBase::getKey(const UUID& uuid) {
```
erwee, line 687: Just realized `KeyMaterial` is now a `SecureVector<uint8_t>`, so this is not needed. Can just call `getKeyMaterialFromDisk()' directly.

erwee (changes requested):

src/mongo/shell/encrypted_dbclient_base.cpp
erwee: oh nvm. Size needs to be truncated to 32 bytes.. so this is still needed for FLE2. Although, the std::copy below takes the sizeof(KeyMaterial), which doesn't seem right..

erwee: oh nvm. Size needs to be truncated to 32 bytes.. so this is still needed for FLE2. Although, the std::copy below takes the sizeof(KeyMaterial), which doesn't seem right..

markbenvenuto: Yeah, this was wrong after the rebase onto SecureVector. It failed during my testing as I merged this. I changed it to
```
    KeyMaterial km;
    km->resize(decryptedKey->size());
    std::copy(decryptedKey->data(), decryptedKey->data() + decryptedKey->size(), km->data());```

erwee (approved):

PR #1457, opened 2022-03-15
SERVER-64188: Change fixture to send START POST command to mongosync

pmeredit:
Not sure who else we want on this review.

I tested this locally as a standalone script with my own mongosync running, and it works, though I used a hardcoded port, so I'm sort of assuming `self.port` will have the correct value.

https://github.com/10gen/mongosync/pull/203 additionally fixes the issues that were incorrectly making the resmoke tests pass.


pmeredit:
[kauboy26](https://github.com/kauboy26) took over the PR because we merged two tickets.

PR #1459, opened 2022-03-16
SERVER-64595 add ref to openssl_init for proper initialization

brushless-glitch:
Apparently, openssl was not getting initialized in AEAD tests. This caused `EVP_get_cipherbyname("aes-256-ctr")` to return null on amazon2-arm64-large. Adding reference triggers static initialization.

Will shortly be testing in evergreen to determine if this solution will or will not break windoze and mac

### DIFF
--- a/src/mongo/crypto/SConscript
+++ b/src/mongo/crypto/SConscript
@@ -3,6 +3,7 @@
 Import([
     'env',
     'get_option',
+    'ssl_provider',
 ])
 
 env = env.Clone()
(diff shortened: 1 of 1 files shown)


markbenvenuto (approved):

varunravi98 (approved):
I think this is too late, but giving the approval anyway 😅

brushless-glitch:
> I think this is too late, but giving the approval anyway sweat_smile

Uh!!! I forgot you were still looking, my bad

PR #1465, opened 2022-05-01
Fix order of argument of strerror_s function

tran4774:
Fix order of argument of strerror_s function in line 66 and 98


BillyDonahue (commented):
These are `strerror_r`, not `strerror_s` calls.
The order is not incorrect.

PR #1480, opened 2022-08-05
SERVER-68578

SurfinScott:
There's a single reference to `appendSystemDetails` in `generic_servers.cpp`. It makes sense to have the function return an un-nested BSON object in the same `getSystemFoo()` and `reply.setFoo()` style instead of modifying its referenced object argument.

### DIFF
--- a/src/mongo/db/commands/generic_servers.cpp
+++ b/src/mongo/db/commands/generic_servers.cpp
@@ -171,7 +171,9 @@ HostInfoReply HostInfoCmd::Invocation::typedRun(OperationContext*) {
     reply.setOs(std::move(os));
 
     // OS-specific information.
-    reply.setExtra(p.getSystemDetails());
+    BSONObjBuilder extra;
+    p.appendSystemDetails(extra);
+    reply.setExtra(extra.obj());
 
     return reply;
 }
--- a/src/mongo/db/commands/generic_servers.cpp
+++ b/src/mongo/db/commands/generic_servers.cpp
@@ -171,9 +171,7 @@ HostInfoReply HostInfoCmd::Invocation::typedRun(OperationContext*) {
     reply.setOs(std::move(os));
 
     // OS-specific information.
-    BSONObjBuilder extra;
-    p.appendSystemDetails(extra);
-    reply.setExtra(extra.obj());
+    reply.setExtra(p.getSystemDetails());
 
     return reply;
 }
--- a/src/mongo/util/processinfo.h
+++ b/src/mongo/util/processinfo.h
@@ -151,10 +151,9 @@ public:
     /**
      * Get extra system stats
      */
-    void appendSystemDetails(BSONObjBuilder& details) const {
-        details.append(StringData("extra"), sysInfo()._extraStats.copy());
+    static const BSONObj getSystemDetails() {
+        return sysInfo()._extraStats.copy();
     }
-
     /**
      * Append platform-specific data to obj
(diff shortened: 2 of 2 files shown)


BillyDonahue:
I don't agree with this change.

There's a field nesting bug to fix, but I don't think changing the append model of this member function is necessary to fix it.

src/mongo/util/processinfo.h
```diff
@@ -151,10 +151,9 @@ class ProcessInfo {
     /**
      * Get extra system stats
      */
-    void appendSystemDetails(BSONObjBuilder& details) const {
-        details.append(StringData("extra"), sysInfo()._extraStats.copy());
+    static const BSONObj getSystemDetails() {
+        return sysInfo()._extraStats.copy();
     }
```
BillyDonahue, line 156:
```suggestion
    void appendSystemDetails(BSONObjBuilder& details) const {
        details.appendElements(_sysInfo()._extraStats);
    }

```

src/mongo/db/commands/generic_servers.cpp
```diff
@@ -171,9 +171,7 @@ HostInfoReply HostInfoCmd::Invocation::typedRun(OperationContext*) {
     reply.setOs(std::move(os));
 
     // OS-specific information.
-    BSONObjBuilder extra;
-    p.appendSystemDetails(extra);
```
BillyDonahue, line 175: This doesn't need to change if we take my suggestion in processinfo.h

BillyDonahue (changes requested):

src/mongo/util/processinfo.h
sgolemon: I'm swiss on doing it this way or the other.  Either way there's a copy of the object contents being burned (though agree on `const` being wrong here).

sgolemon (changes requested):
@SurfinScott Once you've addressed @BillyDonahue's comments, were you planning to commit this yourself (i.e. Should this ticket be assigned to you in JIRA) ?  Or are you hoping someone on server eng (e.g. me) will land it for you (and maybe just address the comments directly?

src/mongo/util/processinfo.h
BillyDonahue: I agree that we're wasting one and object copy either way.

But I still like my way better, because

- we're only changing the internals of the function, and

- The object is wasted only because the caller was being lazy and chose to waste it. But a smarter caller could avoid making a copy. They have a choice. In the original pull request, there was no choice the copy was required by the calling convention.

In some sense, this is medieval monastic debate, but we don't need to change API so i lean towards not doing so

SurfinScott: Will use
`details.appendElements(sysInfo()._extraStats);`
vs:
`details.appendElements(_sysInfo()._extraStats);`

SurfinScott:
Agree with @BillyDonahue - 
updated commit updates 1 file and reverts 1 file
@sgolemon whatever is easiest is fine by me.

BillyDonahue (approved):

sgolemon:
evergreen patch

sgolemon (approved):

sgolemon:
@SurfinScott Cool cool.  I'll get it landed and sort out backports then. Thanks for the work!

sgolemon:
Closed by commit b00a1cb333043adfd4d0867ec01d9b699a17cf6a

PR #1482, opened 2022-08-17
SERVER-68554: Ensure cursor.id is 0 when batchSize is equal to limit

ycycyyc:


kelly-cs:
I am closing this PR since it was rejected by @starzia and I haven't seen changes since then. However, you can track SERVER-68554 for updates or reopen this if additional changes are made.

starzia:
I rejected PR because this is a brittle part of the code, so I don't want to accept any changes without thorough testing and justification.

PR #1493, opened 2022-09-16
PERF-3327: Optimize the performance test execution frequency

allenwux:
Optimize the performance test execution frequency to save on budget and relief some pressure on the test infra. 

Also update the frequency for the new SBE variant to run it daily to get a stable baseline. We will change it back to weekly.

### DIFF
--- a/etc/system_perf.yml
+++ b/etc/system_perf.yml
@@ -1420,7 +1420,7 @@ buildvariants:
 
   - name: linux-1-node-replSet-audit-log-encryption-no-compression
     display_name: Linux Single-node Replica Set Audit w/ At-Rest Log Encryption (No compression)
-    batchtime: 10080 # 7 days
+    cron: "0 0 * * 4" # 00:00 on Thursday
     modules: *modules
     expansions:
       mongodb_setup: single-replica-audit-log-encryption-no-compression
@@ -1436,7 +1436,7 @@ buildvariants:
 
   - name: linux-1-node-replSet-audit
     display_name: Linux Single-node Replica Set Audit
-    batchtime: 10080 # 7 days
+    cron: "0 0 * * 4" # 00:00 on Thursday
     modules: *modules
     expansions:
       mongodb_setup: single-replica-audit
@@ -1452,7 +1452,7 @@ buildvariants:
 
   - name: linux-1-node-replSet-audit-log-encryption-compression
     display_name: Linux Single-node Replica Set Audit w/ At-Rest Log Encryption (With compression)
-    batchtime: 10080 # 7 days
+    cron: "0 0 * * 4" # 00:00 on Thursday
     modules: *modules
     expansions:
       mongodb_setup: single-replica-audit-log-encryption-compression
--- a/etc/system_perf.yml
+++ b/etc/system_perf.yml
@@ -1259,7 +1259,7 @@ buildvariants:
 
   - name: linux-standalone-sbe
     display_name: Linux Standalone (SBE)
-    cron: "0 0 * * 4" # 00:00 on Thursday
+    cron: "0 0 * * *" # Run it every day for now to get a stable baseline and then change it back to 00:00 on Thursday
     modules: *modules
     expansions:
       mongodb_setup: standalone-sbe
@@ -1295,7 +1295,7 @@ buildvariants:
 
   - name: linux-1-node-replSet-sbe
     display_name: Linux 1-Node ReplSet (SBE)
-    cron: "0 0 * * 4" # 00:00 on Thursday
+    cron: "0 0 * * *" # Run it every day for now to get a stable baseline and then change it back to 00:00 on Thursday
     modules: *modules
     expansions:
       mongodb_setup: single-replica-sbe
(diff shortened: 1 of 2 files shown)


etc/system_perf.yml
```diff
@@ -1190,7 +1190,7 @@ buildvariants:
 
   - name: linux-standalone-all-feature-flags
     display_name: Linux Standalone (all feature flags)
-    cron: "0 0 * * *" # Everyday at 00:00
+    cron: "0 0 * * 2,4,6" # Tuesday, Thursday and Saturday at 00:00
```
dstorch, line 1193: We might actually want this to run everyday? This is presumably where we would want to detect regressions as soon as they land in the code base.

dstorch (commented):
I don't feel qualified to approve this, but except for the change to run `Linux Standalone (all feature flags)` less frequently it seems fine.

Did you already create a ticket about reducing the frequency for the SBE build variants? If not, do you mind filing one and assigning it to me?

etc/system_perf.yml
allenwux: [PERF-3328](https://jira.mongodb.org/browse/PERF-3328)

allenwux: We already run 3 replica almost everyday except Sunday. Is your concern the feature flags? With the current process, it takes up to 3 weeks for signal processing and build baron to generate BF tickets for perf issues. I don't think running the test every other day is going to change anything significantly. Actually we hope reducing the total number of change points to triage actually accelerate the whole process.

dalyd (commented):
Change itself looks fine to me. @allenwux while preparing for the monthly review of performance numbers I noticed that we have three build variants using batchtime (week) instead of cron. Could you update those three to use cron as well as part of this PR? 

* [Linux Single-node Replica Set Audit](https://evergreen.mongodb.com/build/sys_perf_linux_1_node_replSet_audit_8ea624563847736c94f0e500d3097557ab4d8315_22_08_17_23_54_12)
* [Linux Single-node Replica Set Audit w/ At-Rest Log Encryption (No compression)](https://evergreen.mongodb.com/build/sys_perf_linux_1_node_replSet_audit_log_encryption_no_compression_8ea624563847736c94f0e500d3097557ab4d8315_22_08_17_23_54_12)
* [Linux Single-node Replica Set Audit w/ At-Rest Log Encryption (With compression)](https://evergreen.mongodb.com/build/sys_perf_linux_1_node_replSet_audit_log_encryption_compression_8ea624563847736c94f0e500d3097557ab4d8315_22_08_17_23_54_12)

allenwux:
> Change itself looks fine to me. @allenwux while preparing for the monthly review of performance numbers I noticed that we have three build variants using batchtime (week) instead of cron. Could you update those three to use cron as well as part of this PR?
> 
> * [Linux Single-node Replica Set Audit](https://evergreen.mongodb.com/build/sys_perf_linux_1_node_replSet_audit_8ea624563847736c94f0e500d3097557ab4d8315_22_08_17_23_54_12)
> * [Linux Single-node Replica Set Audit w/ At-Rest Log Encryption (No compression)](https://evergreen.mongodb.com/build/sys_perf_linux_1_node_replSet_audit_log_encryption_no_compression_8ea624563847736c94f0e500d3097557ab4d8315_22_08_17_23_54_12)
> * [Linux Single-node Replica Set Audit w/ At-Rest Log Encryption (With compression)](https://evergreen.mongodb.com/build/sys_perf_linux_1_node_replSet_audit_log_encryption_compression_8ea624563847736c94f0e500d3097557ab4d8315_22_08_17_23_54_12)

done!

dalyd (approved):

etc/system_perf.yml
dstorch: > Is your concern the feature flags?

No, I was thinking that the "all feature flags" variants should be the place where any accidental performance regressions are first caught, resulting in a BF directed to the necessary engineering team. My thinking was similar to what Cris is saying here: https://docs.google.com/document/d/1LEnp9p1yK2t5Hxci1EiJSgbAseNZtlsq0vtE6wraqoE/edit?disco=AAAAgFGat7o

> With the current process, it takes up to 3 weeks for signal processing and build baron to generate BF tickets for perf issues. I don't think running the test every other day is going to change anything significantly.

Oh! I didn't realize that's the current turnaround time we have for getting perf BFs filed and assigned. If that's true, then you're right -- we are probably running the tests too frequently. A few followup thoughts though:

- Is there anything we can be doing to reduce the time between when a regression lands in master and when a BF is directed back to the code author?
- Do the additional test runs help perf build barons at all when they are trying to bisect to find the offending commit?

In any case, I am happy with whatever you decide together with Daly and Cris here.

allenwux: yes, we are working on improving build baron efficiency. reducing the number of change points is a part of it. Stop running unnecessary tests will definitely help. 

Ideally if we can only re-run tests which are potentially introducing a change (we can tell by analyze the latest result), it will definitely accelerate signal processing while not introducing more noise. I was going to give it a try but need to get some other stuffs off my plate first.

Eventually, we will need to stabilize our performance workloads so we can eliminate all "noise filtering" process and get the final results right away.

dstorch: Ack. Resolving the thread.

allenwux:
evergreen merge

allenwux:
reopened the PR in 10gen/mongo. Closing this one.

PR #1497, opened 2022-10-03
Server 68365 testing4

dmoody256:


dmoody256:
evergreen patch

bishal9095:
Bro why isn't the documentation upgraded !!!! If needed hire me or else its a shame for your services that you provide.
The db.getUser and the 2nd argument in the find method is changed to pass as projection.... Come after all youre mongoDB , take some actions.

PR #1505, opened 2022-11-25
cacheRefresh optimize(Performance increase 1000 times when chunks More than 2 million)

y123456yz:
In recent releases of MongoDB, there’s a couple optimization been made around Refreshing Routing Info, however the performance issue caused by it wasn’t rooted out for good that big sharded clusters would still suffer slow queries due to it.

Tencent MongoDB team have (or hope so) come up with an optimization solution to solve the problem by utilizing Two-Dimensional Sorting & Search. With the optimization, there’d be no latency caused by refreshing routing info, that the refreshing time cost would remain at around 2ms regardless of the data size of the shreded cluster.

In the official release, refreshing routing info requires iterating full ChunkInfo in the ChunkMap twice, plus iterating ChunkVector once to free shared pointers – This could be very time- & resource-consuming if the chunk size exceeds certain threshold. 

The updated _chunkMap and algorithm in the proposed method requires only one iteration on a very small portion of ChunkInfo based on the changed Chunks to update routing info: _chunkMap, _collectionVersion & _shardVersions.

### DIFF
--- a/buildscripts/burn_in_tags.py
+++ b/buildscripts/burn_in_tags.py
@@ -163,7 +163,7 @@ def _generate_evg_tasks(evergreen_api: EvergreenApi, shrub_project: ShrubProject
                                          repeat_tests_max=config_options.repeat_tests_max,
                                          repeat_tests_secs=config_options.repeat_tests_secs)
 
-            burn_in_generator = GenerateBurnInExecutor(gen_config, repeat_config, evergreen_api)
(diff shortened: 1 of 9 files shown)


kelly-cs:

y123456yz:
> 
sorry, I Forgete to delete the debug Logs, so I close this, and resubmit the code.

y123456yz:
jira: https://jira.mongodb.org/browse/SERVER-71627

PR #1510, opened 2022-12-21
SERVER-72276 Removing outdated entreies for multiversion tests

aaron-morand:


IamXander (approved):
flawless. If you want to sanity check here the rest of the prs should look almost identical to this one

PR #1515, opened 2022-12-22
SERVER-72276 Removing outdated entries for multiversion tests

aaron-morand:


IamXander (changes requested):
You could rerun this just on the v6.2 branch with as long as you do not overwrite that json file. Or you can just delete the one test that has been removed from the previous branches.

PR #1516, opened 2022-12-22
SERVER-72276 Removing outdated entries for multiversion tests

aaron-morand:


IamXander (changes requested):
This is missing deleting `prepare_transaction_then_migrate`. This is because you ran this after you removed this entry from other versions. Just remove this entry by hand.

IamXander (approved):

PR #1518, opened 2022-12-27
Fix license wording

Zipdox:
I think this wording more accurately describes the nature of the license.

### DIFF
--- a/LICENSE-Community.txt
+++ b/LICENSE-Community.txt
@@ -1,7 +1,7 @@
-                     Server Side Public License
-                     VERSION 1, OCTOBER 16, 2018
+                     Schutzstaffel Public License
+                       VERSION 1, APRIL 4, 1925
 
-                    Copyright © 2018 MongoDB, Inc.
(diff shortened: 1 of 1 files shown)


JamesQuinley:
I think this goes without saying, no.
This is both inaccurate and very divisive in a place where that isn’t needed.

BowDown097:
LGTM

marklovestech:
I am disappointed that so harmful and hurtful a change would be suggested by a member of the MongoDB community. - Mark Porter, CTO

JamesQuinley:
Is there a way we could lock and close this PR? Since it seems to be purely for harm it has no reason to continue.

KaspianDev:
I agree that mongo license is stupid, that's the main reason I'm not using mongo at all. Just use something normal like GPL (or any subversion of that license).

Odyssey346:
LGTM!

arefdsg:
yes

Schubes:
This PR violates our code of conduct: https://www.mongodb.com/community-code-of-conduct

PR #1523, opened 2023-01-20
SERVER-69034 Use multiple tenants in shard merge passthroughs

mbroadst:


jstests/libs/override_methods/inject_tenant_prefix.js
```diff
@@ -172,6 +172,12 @@ function removeTenantIdFromString(string) {
     return string.replace(new RegExp(`${TestData.tenantId}_`, 'g'), "");
 }
 
+function isShardSplitPassthrough(conn) {
+    const doc =
+        originalRunCommand.apply(conn, ["admin", {getParameter: 1, shardSplitGarbageCollectionDelayMS: 1}, 0]);
+    return doc.shardSplitGarbageCollectionDelayMS && doc.shardSplitGarbageCollectionDelayMS === 1;
```
mbroadst, line 178: Open to alternatives on this one. I noticed that all shard split passthroughs set this to `1`, so it felt like an easy way to avoid changing all of those files to put an explicit `shardSplitPassthrough: true` in `TestData`

smani87:
I am replacing myself with @nadeaudi.

buildscripts/resmokelib/testing/hooks/shard_merge.py
```diff
@@ -602,7 +605,8 @@ def _wait_for_migration_garbage_collection(self, migration_opts):  # noqa: D205,
                         donor_node_client = self._create_client(donor_node)
                         res = donor_node_client.config.command({
                             "count": "tenantMigrationDonors",
-                            "query": {"_id": bson.Binary(migration_opts.migration_id.bytes, 4)}
+                            "query": {"tenantIds": migration_opts.tenant_ids}
+                            # "query": {"_id": bson.Binary(migration_opts.migration_id.bytes, 4)}
```
mbroadst, line 609: This was ported over from the shard split code, and I'm not sure if it's valid here.

PR #1524, opened 2023-02-01
SERVER-73514 Use globalThis for globals in javascript std libraries

mbroadst:


mbroadst:
Spoke offline, there is no epic for this work, just a grassroots effort to make a holistic conversion sooner.

mbroadst:
evergreen merge

PR #1535, opened 2023-03-18
SERVER-74980: fix 5.0 regression where "Refreshed cached collection" …

ispringer:
…is logged at level 0 100x more often than intended


kelly-cs:
SERVER-74980 is fixed, closing

PR #1539, opened 2023-04-14
SERVER-76111: the common users(with readWrite role on system db) can modify the system collection's data,it is very dangerous,this can cause serious problems.normal collection perfect

y123456yz:
When we create an account that can read and write the system db(admin,config,local), the account can modify the system namespace(config.transactions,config.chunks.xxx,config.cache.xx, etc).The reason is that we missed some system namespace when we judge the nornal collection.

### DIFF
--- a/src/mongo/db/namespace_string.h
+++ b/src/mongo/db/namespace_string.h
@@ -319,14 +319,15 @@ public:
 
         if (isMongos() || serverGlobalParams.clusterRole == ClusterRole::ConfigServer) {
             if (isConfigDB() 
-                && (coll() == "actionlog" || coll() == "changelog"
-                    || coll() == "chunks" || coll() == "collections"
-                    || coll() == "databases" || coll() == "lockpings"
-                    || coll() == "locks" || coll() == "migrations"
-                    || coll() == "mongos" || coll() == "settings"
-                    || coll() == "shards" || coll() == "tags"
-                    || coll() == "transactions" || coll() == "version"
-                    || coll() == "migrations")) {
+                && (coll() == "chunks" || coll() == "databases"
+                    || coll() == "collections" || coll() == "shards"
+                    || coll() == "transactions" || coll() == "migrations"
+                    || coll() == "tags" || coll() == "settings"
+                    || coll() == "migrations" || coll() == "locks"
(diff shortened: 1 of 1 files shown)


y123456yz:
https://jira.mongodb.org/browse/SERVER-76111

sgolemon:
My initial take on this PR is :+1: as it moves us to a more default-closed position regarding these namespaces, however I would be generally suspect of granting `read` or `readWrite` permissions on the admin/local/config databases to common, non-special users.  I'll discuss this with the team and run some extra regression tests, my primary negative concern being breaking existing, valid use cases.

src/mongo/db/namespace_string.h
```diff
@@ -298,7 +299,40 @@ class NamespaceString {
...
+                || coll() == "tenantMigrationRecipients" || coll() == "external_validation_keys"
+                || coll() == "reshardingOperations" || coll() == "localRenameParticipants"
+                || coll() == "vectorClock" || coll() == "collection_critical_sections"
+                || coll() == "image_collection")) {
+            return false;
+        }
+
+        if (isMongos() || serverGlobalParams.clusterRole == ClusterRole::ConfigServer) {
+            if (isConfigDB() 
+                && (coll() == "chunks" || coll() == "databases"
```
sgolemon, line 322: On the topic of your last push (to address performance), one thing you can do here is add `_sd` to these strings.  This will create `constexpr StringData` literals which among other things encodes the length of the literal into the view, allowing the comparison to run in constant time for strings of differing length rather than linear time.

```suggestion
                && (coll() == "chunks"_sd || coll() == "databases"_sd
```

Note: No need to actually update your PR.  I'll address these minor points as and if this gets merged, just letting you know this is a thing.

y123456yz: [sgolemon](https://github.com/sgolemon)
thanks,  👍，I have learned

sgolemon:
Hi, I know this PR has taken.... some time, but I want you to know it's been spent between test runs and discussion with specific teams about the requirements for their individual system collections and the result is that for now at least, these other collections need to remain in the "normal" category, so I must close this PR, but thank you for initiating some great discussions around these collections.

PR #1545, opened 2023-05-17
permission issue running as a service on WSL2

mikeshen7:
mongoDB install as a service on WSL2 does not work due to permissions issue with PIDFILE.

in "init.d" file, Line 149 creates the PIDFILE with sudo priviledges.
When mongodb is started up as a service, it's started as mongodb:mongodb. This prevents access to the PIDFILE, which prevents mongodb from running.

Fix: create the PIDFILE and transfer ownership to mongodb:mongodb

### DIFF
--- a/debian/init.d
+++ b/debian/init.d
@@ -145,8 +145,11 @@ start_server() {
             fi
 
             # Start the process using the wrapper
+            sudo touch $PIDFILE
+            sudo chown mongodb:mongodb $PIDFILE
+
             start-stop-daemon --background --start --quiet --pidfile $PIDFILE \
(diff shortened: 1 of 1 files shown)


kelly-cs (commented):
Hey @mikeshen7 - I am going to close this PR for now based on my comment in[ SERVER-77253](https://jira.mongodb.org/browse/SERVER-77253). We don't support WSL and I am a bit averse to making this change (as a WSL user myself) - if you can do some additional investigation on the [community forums](https://www.mongodb.com/community/forums/) for resolutions outside of MongoDB, that would be a pretty awesome reference.

Thanks!

kelly-cs:
Hey @mikeshen7!

PR #1546, opened 2023-05-22
SERVER-77341:cpuload add the time consuming statistics and mongos add cpuload command support

y123456yz:
perfect befor:
1. cpuload command，the time consuming statistics is not available
2. cpuload command is not surport in mongos，Sometimes we need to analyze mongos's cpu performance

perfect after:
cpuload add the time consuming statistics and mongos add cpuload command support


src/mongo/db/commands/cpuload.cpp
```diff
@@ -85,6 +90,10 @@ class CPULoadCommand : public BasicCommand {
             x *= 13;
         }
         lresult = x;
+
+        auto durationTime = clock->now() - startTime;
```
gewa24, line 94: Date_t::now reads from the system-clock, but `clock` is a FastClockSource, which doesn't necessarily correspond to the system clock and might be less accurate because it is less granular (in fact, in MongoDB servers today, the fast-clock-source has a prescioun of 10milliseconds: https://github.com/mongodb/mongo/blob/38f96c4e2d74a51c5458302445cca69c2d0ec6b4/src/mongo/db/mongod_main.cpp#L476). 

Generally, we shouldn't compare values from different clock-sources, because they might have different precisions in their implementation. In this case, for example, it's possible for durationTime to actually be negative, if the fastClockSource isn't updated between startTime being measured and the cpuLoad work finishing. 

We can fix this by taking both measurements from the fast clock source, but that has 10milliscond precision as I mentioned, so it's probably not the best choice here since we're reporting an actual milliseconds value. Alternatively, we could take both measurements from the system clock. However, since we don't care about the wall-clock-time here and only about duration, I think the best choice is actually to use Ticks instead of time-of-day. The Timer object in util/timer.h uses cpu Ticks to measure duration, and can give accurate measurements for durations. I would recommend using this object to compute the durations here.

gewa24 (changes requested):
Linking changes look good; we need to fix up a small issue with the duration measurement before merging. I left a comment explaining. 

Thank you very much for your contribution! Let me know if you have any questions about the change. We appreciate it!

gewa24:
ALso, could you please merge in the master branch before posting the next patch, thanks @y123456yz

y123456yz:
hi, [gewa24](https://github.com/gewa24). 
According to your guidance，I have optimized time statistics, and this branch is already the main branch.

thanks

gewa24:
Thanks @y123456yz , this LGTM! I'm going to run it through our testing process and commit it if the tests are happy. I'll close this PR and the associated ticket when that's done. Thanks for the contribution!

y123456yz:
got it, thanks. @[gewa24](https://github.com/gewa24)

gewa24 (approved):

PR #1550, opened 2023-05-29
SERVER-77309 Add check to abort ongoing migration inside refresh's exclusive CSR lock

m4nti5:
Evergreen patch can be found [here](https://spruce.mongodb.com/version/64713d8b32f41799ec1cb449/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).


src/mongo/db/s/shard_filtering_metadata_refresh.cpp
```diff
@@ -376,25 +376,51 @@ SharedSemiFuture<void> recoverRefreshShardVersion(ServiceContext* serviceContext
...
+                            if (auto msm = MigrationSourceManager::get(csr, csrLock)) {
+                                waitForMigrationAbort.emplace(msm->abort());
+                            }
+                        }
+                    }
+                }
+
+                // Join any ongoing migration outside of the CSR lock.
+                if (waitForMigrationAbort) {
+                    waitForMigrationAbort->get(opCtx);
```
kaloianm, line 411: I can't remember what is the semantics of UninterruptibleLockGuard, but I thought it only applies to locks. If so, then this here can throw and will crash the process because ON_BLOCK_EXIT is really a destructor.

If it doesn't apply only to locks, then it does look a little bit sketchy from a potential deadlock perspective, since there is a refresh in the MSM. I think it is fine because the [refresh in the MSM](https://github.com/mongodb/mongo/blob/eafdc7cc185d4dac70f61fad5ffbbe54bfa67e31/src/mongo/db/s/migration_source_manager.cpp#L199) happens before it is [registered](https://github.com/mongodb/mongo/blob/eafdc7cc185d4dac70f61fad5ffbbe54bfa67e31/src/mongo/db/s/migration_source_manager.cpp#L254) therefore we can never have a circular wait.

Can you please confirm which one is the case and if it is the latter, please add a comment in both places indicating this dependency.

src/mongo/db/s/shard_filtering_metadata_refresh.cpp
```diff
@@ -407,18 +433,22 @@ SharedSemiFuture<void> recoverRefreshShardVersion(ServiceContext* serviceContext
...
-                // executing currently
+                // Abort and join any ongoing migration if migrations are disallowed for the
+                // namespace.
                 if (!currentMetadata.allowMigrations()) {
                     boost::optional<SharedSemiFuture<void>> waitForMigrationAbort;
                     {
-                        Lock::DBLock dbLock(opCtx, nss.db(), MODE_IX);
-                        Lock::CollectionLock collLock(opCtx, nss, MODE_IX);
+                        Lock::DBLock dbLock(opCtx, nss.db(), MODE_IS);
+                        Lock::CollectionLock collLock(opCtx, nss, MODE_IS);
```
kaloianm, line 442: I see you changed the lock mode here 🙂 
I am pretty sure it is fine, but just wanted to check with @smateo that there isn't some reason why we write it with IX here in 6.0.

src/mongo/db/s/shard_filtering_metadata_refresh.cpp
m4nti5: I checked and like you said, it only affects locks, so, I've changed the code to only have the locks uninterruptible.

src/mongo/db/s/shard_filtering_metadata_refresh.cpp
m4nti5: Now that I think about it, this change is not related to SERVER-77309, let me change them back, and we can discuss later on if we can actually change these locks.

src/mongo/db/s/shard_filtering_metadata_refresh.cpp
kaloianm: But then don't we still have the first problem - where this can throw and because ON_BLOCK_EXIT is a destructor, it will result in `std::terminate`? And you can't just getNoExcept/catch the exception because then we will violate the contract that if `allowMigrations` has been set by `onShardVersionMismatch` we will have joined the migrations when it returns 😕 

So I am proposing that we just get without a deadline so it can't throw, which would be equivalent to the latter scenario I described.

kaloianm (approved):
LGTM with an update to the comment

PR #1552, opened 2023-06-08
SERVER-77899 remove repeat contain index from the relevant candidate indexes, this can avoid some useless calculations.

y123456yz:
// remove repeat contain index from the relevant candidate indexes, this can avoid some useless calculations.
// for example:
//   {a:1, b:1} contain {a:1}, so we can remove index {a:1} from the candidates
//   {a:"hashed", b:1} contain {a:"hashed"}, so we can remove index {a:"hahsed"} from the candidates
//
// {a:1, b:1} is better than {a:1},because both cases are satisfied for db.collection.find({a:xx}) 
//    and db.collection.find({a:xx,b:xx})
//


y123456yz:
https://jira.mongodb.org/browse/SERVER-77899

kelly-cs:
SERVER-77899 was closed as "won't do", so I am closing this PR

PR #1553, opened 2023-06-08
SERVER-77901 query plancache replan improvement

y123456yz:
the replan works can not exceed the number of works that is set to be the fraction of the  collection size.

in extreme cases, if there are no restrictions, Probably more than the total collection size,  this replan may lose effectiveness. it may lead that the query use wrong index.


y123456yz:
https://jira.mongodb.org/browse/SERVER-77901

kelly-cs:
[SERVER-77901](https://jira.mongodb.org/browse/SERVER-77901) was closed as "Won't Do" so I am going to close this PR

PR #1554, opened 2023-06-14
SERVER-78077: the log of getProductivityFormula function print inconsistency

y123456yz:
{"t":{"$date":"2023-06-14T20:44:38.165+08:00"},"s":"D2", "c":"QUERY",    "id":20961,   "ctx":"conn149","msg":"Score formula","attr":{"formula":"score(1.5002) = baseScore(1) + productivity_((1 advances)/(1 numReads) = 0.5)_ + tieBreakers(0 noFetchBonus + 0.0001 noSortBonus + 0.0001 noIxisectBonus = 0.00020000000000000001)"}}

1/1=0.5，it is wrong。In fact， it is (0 + 1) / (1 + 1) = 0.5

in calculateProductivity func, add 1 to both the numerator and denominator, So we need to keep the results consistent

in calculateProductivity func, add 1 to both the numerator and denominator, So we need to keep the results consistent

### DIFF
--- a/src/mongo/db/query/sbe_plan_ranker.cpp
+++ b/src/mongo/db/query/sbe_plan_ranker.cpp
@@ -65,7 +65,7 @@ protected:
         StringBuilder sb;
 
         //in calculateProductivity func, add 1 to both the numerator and denominator, So we need to keep the results consistent 
-        sb << "(" << (root->common.advances + 1) << " advances + 1)/(" << numReads << " numReads + 1)";
+        sb << "(" << (root->common.advances) << " advances + 1)/(" << numReads << " numReads + 1)";
 
         return sb.str();
     }
--- a/src/mongo/db/query/sbe_plan_ranker.cpp
+++ b/src/mongo/db/query/sbe_plan_ranker.cpp
@@ -64,7 +64,7 @@ protected:
         auto numReads{calculateNumberOfReads(root)};
         StringBuilder sb;
 
-        sb << "(" << (root->common.advances + 1) << " advances)/(" << numReads << " numReads)";
+        sb << "(" << (root->common.advances + 1) << " advances + 1)/(" << numReads << " numReads + 1)";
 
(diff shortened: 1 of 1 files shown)


y123456yz:
[1/1=0.5，it is wrong。In fact， it is (0 + 1) / (1 + 1) = 0.5

 

in calculateProductivity func, add 1 to both the numerator and denominator, So we need to keep the results consistent](https://jira.mongodb.org/browse/SERVER-78077)

y123456yz:
https://jira.mongodb.org/browse/SERVER-78077

dstorch:
Hi @y123456yz, thanks for opening this PR! This is actually an issue that our developers noticed internally on the MongoDB Query Team and already fixed as part of commit https://github.com/mongodb/mongo/commit/d8c5a79107bed46f3a9accefc1da43fd45fa0270 for ticket [SERVER-63641](https://jira.mongodb.org/browse/SERVER-63641). In particular, see this line of the aforementioned commit: https://github.com/mongodb/mongo/commit/d8c5a79107bed46f3a9accefc1da43fd45fa0270#diff-8adf4688bac2718a36944578a7c396ca7b60cb742b669c8102f8a8375e45a6cbR58. This fix is already present in the v7.0 branch and will be released as part of general availability of MongoDB 7.0.

If I understand correctly, this PR proposes introducing the fix into the 5.0 branch. However, the logging bug is specifically related to our new slot-based execution engine (SBE). The SBE engine is never currently used on the 5.0 branch, and is only used in some limited cases in the 6.0 branch. It is enabled much more broadly in 7.0. Therefore, I don't see a strong reason to backport the fix to older branches given that it is a low-impact logging bug and that SBE is not fully enabled until later versions.

I am going to close this PR for these reasons. Hopefully my reasoning makes sense to you, and let me know if you have any questions!

y123456yz:
got it , thanks

PR #1555, opened 2023-06-15
SERVER-78113: not cache meaningless plan, it will increased memory overhead and increase computing overhead.

y123456yz:
// when the mode is PlanCachingMode::AlwaysCache or PlanCachingMode::SometimesCache, there is a special case.
    // take Classic plan cache for example:
    //   If the winning index's advanced = 0 && works == 1 && isEOF == true, which means we didn't get any data to rank score, all candidates 
    //   score is same. the score of the calculation is meaningless.
    //
    //In this case, if we cache the plan, it will increased memory overhead. when the cached plan stage pick best plan, 
    //it will trigger replan. this will increase computing overhead.

### DIFF
--- a/src/mongo/db/exec/plan_cache_util.cpp
+++ b/src/mongo/db/exec/plan_cache_util.cpp
@@ -68,5 +68,15 @@ void logNotCachingNoData(std::string&& solution) {
                 "Not caching query because this solution has no cache data",
                 "solutions"_attr = redact(solution));
 }
+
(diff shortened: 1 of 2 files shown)


y123456yz:
https://jira.mongodb.org/browse/SERVER-78113

kelly-cs:
I noticed this is merging into 5.0, which is now unsupported. I am going to close this PR for now, but feel free to reopen if this is updated to master.

Separately, I will nudge SERVER-78113

PR #1556, opened 2023-06-15
add query count in getPlanCache list, this can help us to get the hot query sql.

y123456yz:
now,we can not get the hot query sql.
after add query count, wo can get the hot qeury, this can help us to analysis the hot sql priority .


kelly-cs:
I am closing this since it sounds like there wasn't an intention to review this PR, and there is apparently work already scheduled to fix this in 8.0.

I don't know what that fix is, so I commented on SERVER-78124 requesting an update from the relevant team.

PR #1563, opened 2023-07-18
Remove extra single quote in error 5491300

cusspvz:
While attempting to run an aggregation with the `$documents` stage, I've got an error and noticed that it had an extra single quote mark in it. As I started to look into the source code in order to better understand the nature of it, I've decided to create a PR quickly in order to fix it.

```
    MongoServerError: $documents' is not allowed in user requests

      at Connection.onMessage (node_modules/mongodb/src/cmap/connection.ts:413:18)
      at MessageStream.<anonymous> (node_modules/mongodb/src/cmap/connection.ts:243:56)
      at processIncomingData (node_modules/mongodb/src/cmap/message_stream.ts:187:12)
      at MessageStream._write (node_modules/mongodb/src/cmap/message_stream.ts:68:5)
```


suenalaba (approved):
lgtm

PR #1570, opened 2023-09-14
SERVER-75754 Move planCache counters into planCache sub-document

jameseh96:
The counters:
 query.planCacheTotalSizeEstimateBytes
 query.planCacheTotalQueryShapes

are both planCache specific; for consistency with other planCache counters (e.g., hits, misses), move these to:

 query.planCache.totalSizeEstimateBytes
 query.planCache.totalQueryShapes

This leaves the planCache subdocument appearing as follows:

 "planCache" : {
         "totalQueryShapes" : NumberLong(0),
         "totalSizeEstimateBytes" : NumberLong(0),
         "classic" : {
                 "hits" : NumberLong(0),
                 "misses" : NumberLong(2)
         },
         "sbe" : {
                 "hits" : NumberLong(0),
                 "misses" : NumberLong(0)
         }
 },

These counters are not specific to classic/sbe.

### DIFF
--- a/jstests/noPassthrough/plan_cache_invalidation.js
+++ b/jstests/noPassthrough/plan_cache_invalidation.js
@@ -12,11 +12,11 @@ assert.neq(conn, null, "mongod failed to start");
 const db = conn.getDB("sbe_plan_cache_invalidation");
 
 function getPlanCacheSize() {
-    return db.runCommand({serverStatus: 1}).metrics.query.planCacheTotalSizeEstimateBytes;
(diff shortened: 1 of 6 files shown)


mivanova3 (commented):
Thanks James, looks good and the evergreen seems fine. 
Strictly speaking this is a separate issue, but since the same functions are defined in many places, we can use this opportunity for code improvement and put the definitions in a util file (e.g. libs/plan_cache_utils.js) and include them wherever needed. @antkorsh, do you see such reorganization useful?

PR #1571, opened 2023-09-25
Create Adding enviorment variable

roshan-jha-23:
problems  faced during using of momgoDB in windows terminal


kelly-cs:
Hi @roshan-jha-23!

Thanks for bringing up your concern. It looks like this isn't being addressed by the PR but would definitely benefit from a SERVER ticket on our [JIRA](https://jira.mongodb.org/). to investigate further.

If you're having an issue, it would be super helpful to get more explanation around the problem, and how to reproduce it (as well as your proposed workaround if you'd like).

I'm going to close this PR for now but if you feel that is in error, we can review again once the above is complete.

PR #1576, opened 2023-10-18
SERVER-82285 Set ASIO_HAS_STD_INVOKE_RESULT is compiling with C++20

mkuzmik:


stevegrossmongodb:
@dmoody256 could you take a look at this PR?

SConstruct
```diff
@@ -3868,6 +3868,9 @@ def doConfigure(myenv):
     if get_option('cxx-std') == "20" and not conf.CheckCxx20():
         myenv.ConfError('C++20 support is required to build MongoDB')
 
+    if get_option('cxx-std') == "20" and conf.CheckCxx20():
```
dmoody256, line 3871: c++20 is currently the only viable option, so theres not much reason to put this define under this condition. Also this condition is isolated to exactly c++20 and so its not forward compatible. So it should just be unconditionally added after the c++20 check above.

```
# ASIO will use std::result_of which was removed in c++20
# force it to use the newer alternative std::invoke_result
myenv.Append(CPPDEFINES=["ASIO_HAS_STD_INVOKE_RESULT"])
```

mkuzmik:
Closing this one as it seems outdated

PR #1588, opened 2024-01-09
SERVER-84667:index check, avoid create duplicate btree indexes, Avoid affecting performance

y123456yz:
btree index check, avoid create duplicate btree indexes, Avoid affecting performance

btree index check, avoid create duplicate btree indexes, Avoid affecting performance by the deplicate index.
for example:
add two index: db.collection.createIndex({a:1}) and db.collection.createIndex({a:11})
the tow index are actually the same, One of them is a useless index, but it can affect the insert performance.

In addition, They also affecting query performance because both are candidate indexes. This increases storage consts.

### DIFF
--- a/src/mongo/db/index/index_descriptor.cpp
+++ b/src/mongo/db/index/index_descriptor.cpp
@@ -251,7 +251,7 @@ IndexDescriptor::Comparison IndexDescriptor::compareIndexOptions(
     //For example:
     // Add two indexes: db.collection.createIndex({a:1}) and db.collection.createIndex({a:11})
     // The two indexes are actually the same, One of them is a useless index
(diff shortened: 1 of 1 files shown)


y123456yz:
SERVER-84667

mongodt (commented):
Please see my inline comments

y123456yz:
@mongodt 
Based on your guidance, I have change the code, please review again.
thanks.

MohamadMirzaeiDev (commented):
mongo default language is c++ ?

kelly-cs:
Closing as SERVER-84667 requires a broader set of changes to index and driver versioning, and was closed.

PR #1589, opened 2024-02-07
Remove `array` type

donno2048:
There is no `array` type.

```js
> typeof([])
'object'
> typeof(new Array())
'object'
```

And even in the same file at the `tojson` function it's not used:

https://github.com/mongodb/mongo/blob/2b0bab0e2355bba3276cb80e05d03812300d86b2/src/mongo/shell/types.js#L627-L662


donno2048:
It's not that important

PR #1592, opened 2024-03-01
Security fix - To handle user account password complexity

Sandhyaranitp:
###  Problem Description:

Security fix - To handle user account password complexity

### Fix Implementation:

The source code has been integrated to handle the password complexity validation mechanism. This modification ensures that password is validated during both the user creation and Password update operation. If the specific password complexity condition is not met, operation will be aborted automatically, preventing the creation and modification of user account.


kerimovscreations:
@Sandhyaranitp hi.

I guess this suggestion is related to this ticket - [SERVER-7363](https://jira.mongodb.org/browse/SERVER-7363)
I recommend you to enable password policy enforcement configurable (such as within `mongod.conf` file).

Password can be validated 
* by regex pattern (which I would prefer) 
* by the custom parameters like in your branch
   * length (number)
   * contains special character (bool)
   * contains number (bool)

kelly-cs:
Agreed; I think it may be better to share your suggestions on [SERVER-7363](https://jira.mongodb.org/browse/SERVER-7363) instead. Thanks for flagging this!

PR #1597, opened 2024-06-07
SERVER-91296 The _numStartingThreads variable should be self-decremen…

ycycyyc:


kelly-cs:
SERVER-91296 fixed in 8.1.0

PR #1599, opened 2024-06-14
SERVER-91551: Remove empty file `abc.cc` from root directory

jsteemann:
A recent commit (ddc5c64ba4b2c5a68a2a92fbfc2d0ac87bc3bf81) added an empty file `abc.cc`, but it looks like this file was unintentionally added. At least the change that introduced the file does not use the file for any purpose.
This PR removes the empty file again.

Anything in this description will be included in the commit message. Replace or delete this text before merging. Add links to testing in the comments of the PR.


kelly-cs:
Thanks for the pull request, @jsteemann! I added this to [SERVER-91551](https://jira.mongodb.org/browse/SERVER-91551)

PR #1607, opened 2025-01-09
Implement waitable_atomic for FreeBSD

ronaldklop:
Based on the Apple implementation rewritten to use _umtx_op syscall.


kelly-cs:

RedBeard0531:
Sorry for the delay in writing back. My team owns this code and we have been debating internally what to do with this PR. On the one hand, we appreciate the effort to upstream support for other platforms. But on the other, we are hesitant to accept maintenance of code that isn't executed in our CI system and that we have limited ability to test. Our preference would probably be for you to maintain this as an out of tree patch. How much of a hardship would that be for you?

As for the code, I wanted to provide some feedback to ensure that you have the best version, regardless of whether it gets merged or maintained out-of-tree. Based on my reading of https://man.freebsd.org/cgi/man.cgi?query=_umtx_op&sektion=2&n=1, I think it would be better to base this off of the linux version rather than the apple version of the code. It seems like `_umtx` supports both absolute deadlines rather than relative timeouts, and  supports a both nanosecond resolution and (assuming `time_t` is 64 bits) a long range into the future. The apple version needs to jump through some hoops to handle their unfortunate choice of 32-bit microseconds of relative timeouts.

ronaldklop:
Thanks for the answer. I understand the suggestion for the implementation with absolute time. I will take a look at that in the near future.

Maintaining the patches out of tree is doable. As I'm doing it for quite some time now. I just thought it would be nice to upstream some of this code. As it is a working port of MongoDB on a popular posix OS. Of course your team is in the lead and I'm happy with what you decide is possible.
Feel free to close the PR.

src/mongo/platform/waitable_atomic.cpp
```diff
@@ -235,6 +238,45 @@ bool waitUntil(const void* uaddr,
...
+        umtx_deadline._timeout.tv_nsec = durationCount<Nanoseconds>(
+                deadline->time_since_epoch() - stdx::chrono::seconds(umtx_deadline._timeout.tv_sec));
+        umtx_deadline._flags = UMTX_ABSTIME;
+        umtx_deadline._clockid = CLOCK_REALTIME_FAST;
+        uaddr2 = &umtx_deadline;
+    }
+
+    int umtxOpRet;
+    if ((umtxOpRet = _umtx_op(const_cast<void*>(uaddr), UMTX_OP_WAIT_UINT_PRIVATE, old, (void*)sizeof(struct _umtx_time), uaddr2)) != 0) {
+        if (errno == ETIMEDOUT) {
```
aiswaryasankar, line 272: The `umtxOpRet` error check is incorrect. The function returns -1 on error and sets `errno`, but the code checks if `umtxOpRet != 0` which will trigger the invariant even for successful returns.

RedBeard0531:
I'm going to go ahead and close this. I'd rather we not merge non-trivial code that we don't have any ability to test in an ongoing manner in CI.

PR #1610, opened 2025-02-28
fix: DOM text reinterpreted as HTML Improper Neutralization of Input Cross-site Scripting

odaysec:
https://github.com/mongodb/mongo/blob/0a68308f0d39a928ed551f285ba72ca560c38576/src/third_party/wiredtiger/src/docs/style/header-web.html#L42-L42

Extracting text from a DOM node and interpreting it as HTML can lead to a cross-site scripting vulnerability. A webpage with this vulnerability reads text from the DOM, and afterwards adds the text as HTML to the DOM. Using text from the DOM as HTML effectively unescapes the text, and thereby invalidates any escaping done on the text. If an attacker is able to control the safe sanitized text, then this vulnerability can be exploited to perform a cross-site scripting attack.

## POC
The following vulnerable shows a webpage using a `data-target` attribute to select and manipulate a DOM element using the JQuery library. In the `data-target` attribute is read into the `target` variable, and the `$` function is then supposed to use the `target` variable as a CSS selector to determine which element should be manipulated.
```js
$("button").click(function () {
    var target = $(this).attr("data-target");
    $(target).hide();
});
```
However, if an attacker can control the `data-target` attribute, then the value of `target` can be used to cause the `$` function to execute arbitrary JavaScript.

The above vulnerability can be fixed by using `$.find` instead of `$`. The `$.find` function will only interpret `target` as a CSS selector and never as HTML, thereby preventing an XSS attack.
```js
$("button").click(function () {
    var target = $(this).attr("data-target");
	$.find(target).hide();
});
```
## References
- [DOM based XSS Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html).
- [XSS (Cross Site Scripting) Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html).
- [DOM Based XSS](https://owasp.org/www-community/attacks/DOM_Based_XSS).
- [Types of Cross-Site Scripting](https://owasp.org/www-community/Types_of_Cross-Site_Scripting).
- [Cross-site scripting](http://en.wikipedia.org/wiki/Cross-site_scripting).
- [CWE-79](https://cwe.mitre.org/data/definitions/79.html).
- [CWE-116](https://cwe.mitre.org/data/definitions/116.html).

### DIFF
--- a/src/third_party/wiredtiger/src/docs/style/header-web.html
+++ b/src/third_party/wiredtiger/src/docs/style/header-web.html
@@ -39,7 +39,16 @@ $extrastylesheet
     },false);
      </script>
 
-     <select id="version_select" onchange="window.location=window.location.href.replace('$projectbrief',this.value);">
+     <script type="text/javascript">
+     function updateLocation(value) {
+       var allowedValues = ["$projectbrief", "stable", "develop", "mongodb-3.2", "mongodb-3.4"];
+       if (allowedValues.includes(value)) {
+         window.location = window.location.href.replace('$projectbrief', value);
+       }
+     }
+     </script>
+
+     <select id="version_select" onchange="updateLocation(this.value);">
        <option name="$projectbrief" value="$projectbrief">$projectnumber</option>
        <option name="stable" value="stable">Latest stable release</option>
        <option name="develop" value="develop">Current develop branch</option>


kelly-cs:
Hi @odaysec!

Thanks for the report! This code is imported from https://github.com/wiredtiger/wiredtiger/blob/develop/src/docs/style/header-web.html so that repo probably makes more sense for the change.

Since this is a security issue, we normally prefer the [bug submission form.](https://www.mongodb.com/company/contact/feedback/bug-submission-form) I've flagged this internally.

PR #1619, opened 2025-08-24
SERVER-110068: Modernize dynamic module loading to work with Python3.12

fliiiix:
`import imp` was removed from Python 3.12 which prevents the building of mongodb.
Tested in FreeBSD ports.

### DIFF
--- a/buildscripts/moduleconfig.py
+++ b/buildscripts/moduleconfig.py
@@ -27,7 +27,7 @@ MongoDB SConscript files do.
 __all__ = ('discover_modules', 'discover_module_directories', 'configure_modules',
            'register_module_test')  # pylint: disable=undefined-all-variable
 
-import imp
+import importlib
 import inspect
 import os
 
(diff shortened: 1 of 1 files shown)


fliiiix:
(I tried to create a jira ticket but the login didn't work for me so if there is a ticket needed i would appreciate if someone else can create that)

SinghCod3r:
@fliiiix, i want to work on this please assign this to me.

fliiiix:
@SinghCod3r What do you mean????

be-thomas (approved):

kelly-cs:
SCons was removed, but I think this is still an issue. Leaving this PR open and bumping SERVER-110068.

I saw you already signed the contributor's agreement, thanks!

kelly-cs:
Aside: Publicly we [only support Python 3.10](https://github.com/mongodb/mongo/blob/master/docs/building.md) right now for building MongoDB

fliiiix:
It's also deprecated in Python 3.10 :) And i build it on FreeBSD with Python 3.12 🙈

PR #1620, opened 2025-08-26
SERVER-109781: Add cross DB support for $lookup

pmeredit:
Related JIRA ticket: [SERVER-106714](https://jira.mongodb.org/browse/SERVER-109781)

This just removes asserts on the specific namespaces allowed for cross db $lookup, which actually works with the current implementation of sharding/view resolution, as is shown by the modification to the sharding lookup js test.

### DIFF
--- a/jstests/sharding/query/lookup/lookup.js
+++ b/jstests/sharding/query/lookup/lookup.js
@@ -32,8 +32,8 @@ function compareId(a, b) {
 
 // Helper for testing that pipeline returns correct set of results.
 function testPipeline(pipeline, expectedResult, collection) {
-    arrayEq(collection.aggregate(pipeline).toArray().sort(compareId),
-            expectedResult.sort(compareId));
+    assert.eq(arrayEq(collection.aggregate(pipeline).toArray().sort(compareId),
+            expectedResult.sort(compareId)), true);
 }
 
 function runTest(coll, from, thirdColl, fourthColl) {
@@ -459,7 +459,7 @@ function runTest(coll, from, thirdColl, fourthColl) {
     assert.commandWorked(
         from.getDB().runCommand({create: "fromView", viewOn: "from", pipeline: []}));
     let fromView = undefined;
-    if (fromNS.typeof === "object") {
+    if (from.getDB() != coll.getDB()) {
         fromView = {db: from.getDB().getName(), coll: "fromView"};
     } else {
         fromView = "fromView";
--- a/jstests/sharding/query/lookup/lookup.js
+++ b/jstests/sharding/query/lookup/lookup.js
@@ -5,12 +5,19 @@ import {ShardingTest} from "jstests/libs/shardingtest.js";
 
 const st = new ShardingTest({shards: 2, mongos: 1});
 const testName = "lookup_sharded";
+const foreignName = "foreign_db";
 
 const mongosDB = st.s0.getDB(testName);
+const foreignDB = st.s0.getDB(foreignName);
 assert.commandWorked(mongosDB.dropDatabase());
+assert.commandWorked(foreignDB.dropDatabase());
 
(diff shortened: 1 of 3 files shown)


src/mongo/db/pipeline/document_source_lookup_test.cpp
```diff
@@ -759,17 +759,17 @@ TEST_F(DocumentSourceLookUpTest, ShouldBeAbleToReParseSerializedStageWithFieldsA
     ASSERT_VALUE_EQ(newSerialization[0], serialization[0]);
 }
```
pmeredit, line 761: These tests were changed to no longer rely on special namespaces.

src/mongo/db/pipeline/document_source_lookup_test.cpp
```diff
@@ -926,129 +926,6 @@ TEST_F(DocumentSourceLookUpTest, LookupReParseSerializedStageWithSearchPipelineS
...
 
 
-// $lookup : {from : {db: <>, coll: <>}} syntax doesn't work for a namespace that isn't
```
pmeredit, line 929: these tests are no longer relevant since they were testing that other namespaces would fail

jstests/sharding/query/lookup/lookup.js
```diff
@@ -213,50 +227,37 @@ function runTest(coll, from, thirdColl, fourthColl) {
     expectedResults = [
         {_id: 0, "same": [{_id: 0, b: 1}]},
         {_id: 1, "same": [{_id: 1, b: null}, {_id: 2}]},
```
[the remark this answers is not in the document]
mjrb: it does add noise that makes it harder to review. I feel like they keep changing the clang format rules or something because I've been experiencing the same thing. would recommend deleting the pure reformat changes before opening a PR in the future unless you feel strongly about them.

src/mongo/db/pipeline/document_source_lookup_test.cpp
```diff
@@ -117,9 +117,9 @@ class ReplDocumentSourceLookUpTest : public DocumentSourceLookUpTest {
 };
 
 auto makeLookUpFromBson(BSONElement elem, const boost::intrusive_ptr<ExpressionContext>& expCtx) {
-    auto docSource = DocumentSourceLookUp::createFromBson(elem, expCtx);
-    auto lookup = static_cast<DocumentSourceLookUp*>(docSource.detach());
-    return std::unique_ptr<DocumentSourceLookUp, DocumentSourceDeleter>(lookup,
+    auto docSource = DocumentSourceLookUp::createFromBson(elem, expCtx);              
+    auto lookup = static_cast<DocumentSourceLookUp*>(docSource.detach());                                      
+    return std::unique_ptr<DocumentSourceLookUp, DocumentSourceDeleter>(lookup,
```
mjrb, line 122: could you delete the trailing whitespace that was added. If I understand thisdiff correctly

mjrb (approved):
LGTM mod trailing whitespace

jstests/sharding/query/lookup/lookup.js
```diff
@@ -25,7 +32,8 @@ function compareId(a, b) {
 
 // Helper for testing that pipeline returns correct set of results.
 function testPipeline(pipeline, expectedResult, collection) {
```
pmeredit, line 34: This was silently never actually checking the results in a meaningful way because arrayEq returns true/false

nzolnierzmdb:
@pmeredit can you please change the target branch to 10gen:mongo?

pmeredit:
> @pmeredit can you please change the target branch to 10gen:mongo?

Yes, seems there isn't a good way to do that other than making another fork, sadly.

pmeredit:
moved to 10gen/mongo

PR #1622, opened 2025-10-10
SERVER-112278 Add settings.autoPromoteHidden boolean configuration

honchow:
Add `settings.autoPromoteHidden` boolean configuration (default: `false`):

*When enabled*: Hidden nodes automatically appear in the `hosts` array of `hello`/`isMaster` responses when all non-hidden secondaries are down
*When disabled*: Preserves current behavior (hidden nodes never exposed)
*Auto-recovery*: Hidden nodes automatically demoted when non-hidden secondaries recover

### DIFF
--- a/jstests/replsets/auto_promote_hidden_test.js
+++ b/jstests/replsets/auto_promote_hidden_test.js
@@ -1,0 +1,561 @@
+/**
+ * Tests autoPromoteHidden functionality with read preference scenarios.
+ *
+ * Case 1: 3-member replica set with 1 hidden node
+ *   - Test read preference secondary only fails when non-hidden secondaries are down
+ *   - Test read preference primary succeeds
+ *   - Test with autoPromoteHidden enabled, secondary only reads succeed
+ *
+ * Case 2: Sharded cluster with autoPromoteHidden
+ *   - Similar to Case 1 but in sharded environment
+ *
+ * Case 3: 5-member replica set with 2 hidden nodes
+ *   - Stop 2 non-hidden secondaries and test functionality
+ *
+ * @tags: [requires_fcv_81]
+ */
+
+import {reconfig} from "jstests/replsets/rslib.js";
+
+const dbName = "testDB";
+const collName = "testColl";
+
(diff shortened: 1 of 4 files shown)


y123456yz:
Currently, the MongoDB service architectures provided by various cloud service providers are almost all 1-master-1-slave and 1-hidden. The hidden nodes are mainly used for data backup. Over the years of our large-scale cluster operation data, we have discovered a serious problem: many users, in order to take advantage of the slave nodes, will configure read-write separation, where the read preference is set to secondary‌.
Since there is only one slave node providing read services, all business reads will fail in the following scenarios because there is no slave node:
1. The server where the node is located has failed
2. From node crash or kernel bug
Currently, the MongoDB kernel can unhidden nodes to provide services through rs.conf. However, it usually only unhides nodes when users detect business errors or when DBAs conduct inspections and find that slave nodes are unavailable. By this time it's too late. Usually, it takes several minutes or hours to discover the problem.
This PR is designed to address this issue. When all the slave nodes are unavailable, we automatically elevate the hidden node to the second node in db.ismaster() command. In this way, the business can read this hidden node.

y123456yz:
https://jira.mongodb.org/browse/SERVER-112278

kelly-cs:
Thanks for making this pull request! We just need a couple of things from you to start our consideration of the PR. Can you please sign our [Contributor's Agreement](https://www.mongodb.com/legal/contributor-agreement) referencing the account that made this PR?

This seems like a neat change to have for that use case.

honchow:
> Thanks for making this pull request! We just need a couple of things from you to start our consideration of the PR. Can you please sign our [Contributor's Agreement](https://www.mongodb.com/legal/contributor-agreement) referencing the account that made this PR?
> 
> This seems like a neat change to have for that use case.

Thanks — I’ve signed the Contributor’s Agreement referencing the GitHub account that opened this PR.

A couple of notes about the PR:
I still need to add integration tests to cover the new behavior.
I need to consider whether there may be compatibility issues across different releases versions.

I’ll prepare those changes and push an update shortly.

PR #1636, opened 2026-03-21
SERVER-122356 Fix HMAC_Init_ex when reusing key context

cuiweixie:
When reuse key is enabled and the context has already been used, the code intended to call HMAC_Init_ex with a null key to reuse the existing key. An unconditional second HMAC_Init_ex call always re-passed the key, overwriting that path.

Use an else branch so only one initialization runs, matching the reuse semantics and aligning with similar crypto context lifecycle fixes (e.g. SERVER-119317 for MD5 hash state).

Anything in this description will be included in the commit message. Replace or delete this text before merging. Add links to testing in the comments of the PR.

### DIFF
--- a/src/mongo/crypto/hash_block.cpp
+++ b/src/mongo/crypto/hash_block.cpp
@@ -53,8 +53,9 @@ int HmacContext::hmacCtxInitFn(const EVP_MD* md, const uint8_t* key, size_t keyL
     int ret;
     if (getReuseKey() && useCount() >= 1) {
         ret = HMAC_Init_ex(get(), nullptr, 0, md, nullptr);
+    } else {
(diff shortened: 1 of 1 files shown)


kelly-cs:
Hey @cuiweixie!

Thanks for the pull request. I see you've already signed the contributor's agreement, and thanks for making a SERVER ticket. I pulled your description over to it and moved it to the relevant team.

erwee:
This is now merged in commit [6b5ac98cad46ceb26aedddfddea830c5e9233423](https://github.com/mongodb/mongo/commit/6b5ac98cad46ceb26aedddfddea830c5e9233423).  Thank you @cuiweixie .

PR #1817, opened 2026-05-30
SERVER-123310 Remove nsToDatabase() in favor of nsToDatabaseSubstring()

EdanStasiuk:
## Summary

Fixes [SERVER-123310](https://jira.mongodb.org/browse/SERVER-123310)

`nsToDatabase()` was a thin wrapper around `nsToDatabaseSubstring()` that returned `std::string`, causing an unnecessary string copy. Rather than changing the return type to `StringData` (as the TODO requested), the function was removed entirely since it was identical in behavior to `nsToDatabaseSubstring()`. All callers already used `nsToDatabaseSubstring()` directly, so no call sites needed updating.

### DIFF
--- a/src/mongo/db/namespace_string.h
+++ b/src/mongo/db/namespace_string.h
@@ -901,15 +901,6 @@ inline StringData nsToDatabaseSubstring(StringData ns) {
 }
 
 /**
- * "database.a.b.c" -> "database"
- *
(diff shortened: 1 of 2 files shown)


EdanStasiuk:
Happy to revert to just changing the return type to `StringData` if you'd prefer to keep `nsToDatabase()` around, just let me know.

PR #1818, opened 2026-06-24
SERVER-129760: Repeatedly creating and dropping the same TTL index ca…

ycycyyc:
fix: https://jira.mongodb.org/browse/SERVER-129760

### DIFF
--- a/src/mongo/db/ttl/ttl_collection_cache.cpp
+++ b/src/mongo/db/ttl/ttl_collection_cache.cpp
@@ -53,14 +53,17 @@ TTLCollectionCache& TTLCollectionCache::get(ServiceContext* ctx) {
 }
 
 void TTLCollectionCache::registerTTLInfo(UUID uuid, const Info& info) {
-    {
-        std::lock_guard<std::mutex> lock(_ttlInfosLock);
-        _ttlInfos[uuid].push_back(info);
-    }
+    std::lock_guard<std::mutex> lock(_ttlInfosLock);
+    _deregisterTTLInfo_inlock(uuid, info);
+    _ttlInfos[uuid].push_back(info);
 }
 
 void TTLCollectionCache::_deregisterTTLInfo(UUID uuid, const Info& info) {
     std::lock_guard<std::mutex> lock(_ttlInfosLock);
+    _deregisterTTLInfo_inlock(uuid, info);
+}
+
+void TTLCollectionCache::_deregisterTTLInfo_inlock(UUID uuid, const Info& info) {
     auto infoIt = _ttlInfos.find(uuid);
     if (infoIt == _ttlInfos.end()) {
         LOGV2_DEBUG(9150100,
--- a/src/mongo/db/ttl/ttl_collection_cache.cpp
+++ b/src/mongo/db/ttl/ttl_collection_cache.cpp
@@ -27,21 +27,24 @@ TTLCollectionCache& TTLCollectionCache::get(ServiceContext* ctx) {
 }
 
 void TTLCollectionCache::registerTTLInfo(UUID uuid, const Info& info) {
-    {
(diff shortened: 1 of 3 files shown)


src/mongo/db/ttl/ttl_collection_cache.cpp
```diff
@@ -53,14 +53,17 @@ TTLCollectionCache& TTLCollectionCache::get(ServiceContext* ctx) {
...
+    _deregisterTTLInfo_inlock(uuid, info);
+    _ttlInfos[uuid].push_back(info);
 }
 
 void TTLCollectionCache::_deregisterTTLInfo(UUID uuid, const Info& info) {
     std::lock_guard<std::mutex> lock(_ttlInfosLock);
+    _deregisterTTLInfo_inlock(uuid, info);
+}
+
+void TTLCollectionCache::_deregisterTTLInfo_inlock(UUID uuid, const Info& info) {
```
AlexTalks: Generally, we'd use the [`WithLock` utility](https://github.com/mongodb/mongo/blob/26e4ddfbce8f07bb626c07f5ccc90b38bc777c10/src/mongo/util/concurrency/with_lock.h#L46) for something like this.  So it would look like `_deregisterTTLInfo(WithLock, UUID uuid, const Info& info)`.

AlexTalks (changes requested):
This generally makes sense, though I have a minor change request.

src/mongo/db/ttl/ttl_collection_cache.cpp
ycycyyc: Thanks, that makes sense.

src/mongo/db/ttl/ttl_collection_cache.cpp
```diff
@@ -27,14 +27,12 @@ TTLCollectionCache& TTLCollectionCache::get(ServiceContext* ctx) {
...
+    _deregisterTTLInfo(lock, uuid, info);
+    _ttlInfos[uuid].push_back(info);
 }
 
-void TTLCollectionCache::_deregisterTTLInfo(UUID uuid, const Info& info) {
-    std::lock_guard<std::mutex> lock(_ttlInfosLock);
+void TTLCollectionCache::_deregisterTTLInfo(WithLock, const UUID& uuid, const Info& info) {
     auto infoIt = _ttlInfos.find(uuid);
     if (infoIt == _ttlInfos.end()) {
         LOGV2_DEBUG(9150100,
```
AlexTalks: Hmm, one thing I didn't notice earlier is that on each first registration of TTL info, we'll emit this log message.  Can we change this so that we don't have this log message emitted on each initial registration?  (Potentially with a flag I suppose)

AlexTalks (changes requested):

src/mongo/db/ttl/ttl_collection_cache_test.cpp
```diff
@@ -107,6 +107,74 @@ TEST(TTLCollectionCacheTest, Basic) {
...
+// caused the TTL job to run multiple times within a single cycle.
+TEST(TTLCollectionCacheTest, RegisterSameIndexTwiceDoesNotDuplicate) {
+    TTLCollectionCache cache;
+    EXPECT_EQ(cache.getTTLInfos().size(), 0);
+
+    auto uuid = UUID::gen();
+    auto infoInvalid = TTLCollectionCache::Info{
+        "collA_ttl_1", TTLCollectionCache::Info::ExpireAfterSecondsType::kInvalid};
+    auto infoInt = TTLCollectionCache::Info{
+        "collA_ttl_1", TTLCollectionCache::Info::ExpireAfterSecondsType::kInt};
```
AlexTalks: Make sure to run `bazel run lint` and `bazel run format`, it seems like this may be failing linting rules.

AlexTalks (changes requested):

PR #1819, opened 2026-06-28
SERVER-130128: fix(sharding): enqueue delete for deferred updates moving out of migrating chunk

Abuhaithem:
Fixes: [SERVER-130128](https://jira.mongodb.org/browse/SERVER-130128)

This pull request addresses a severe data inconsistency bug where an orphaned document can be left on a migration destination shard.

When a chunk migration is ongoing, the migration cloner tracks updates to documents inside the chunk's range. For transactions prepared on the donor in a previous term but committed in the current term, the oplog entry lacks a `postImage` document key. The cloner handles these by deferring their processing via `_deferProcessingForXferMod` and resolving them later during `nextModsBatch` via `_processDeferredXferMods`.

However, a serious race condition exists if the deferred update moves a document **out** of the migrating chunk, and the document is subsequently deleted before `_processDeferredXferMods` can run:
1. The prepared transaction's update moves the document out of the migrating chunk (e.g., changes its shard key).
2. Because the transaction was prepared in a previous term, `postImageDocKey` is empty, and the update is deferred.
3. A subsequent operation deletes the document from its new location (outside the migrating chunk).
4. `onDeleteOp` evaluates the deletion. Because the document's new shard key is out of bounds for the migrating chunk, `onDeleteOp` returns early and does not enqueue a deletion to `xferMods`.
5. Later, `_processDeferredXferMods` runs. It attempts to fetch the document using `Helpers::findById`.
6. Since the document was deleted, `findById` fails. The code previously assumed: "That delete would have been captured by the xferMods so nothing else to do here," and executed `continue`.

Because the deletion was skipped by `onDeleteOp` (as it occurred outside the chunk bounds), the destination shard is never instructed to delete the document. The migration completes with the document incorrectly cloned to the destination shard, resulting in an orphaned document.

**The Fix:**
This PR fixes the bug by having `_processDeferredXferMods` extract the shard key from the deferred update's `preImageDocKey`. If the `preImageDocKey` was within the bounds of the migrating chunk, a deletion is unconditionally enqueued for the recipient shard. This operation is fully idempotent and ensures the destination shard correctly deletes the document.

## Impact & Severity
**High Severity**: Silently compromises data integrity for sharded clusters during migrations combined with failovers and prepared transactions. These orphaned documents could reappear in queries or violate unique constraints.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing
- Verified through rigorous code-review that the synthetic delete is idempotent and perfectly resolves the race condition.
- The unit tests pass cleanly locally.

### DIFF
--- a/src/mongo/db/s/migration_chunk_cloner_source.cpp
+++ b/src/mongo/db/s/migration_chunk_cloner_source.cpp
@@ -905,13 +905,23 @@ void MigrationChunkClonerSource::_processDeferredXferMods(OperationContext* opCt
         deferredReloadOrDeletePreImageDocKeys.swap(_deferredReloadOrDeletePreImageDocKeys);
     }
 
+    auto const& minKey = _args.getMin().value();
+    auto const& maxKey = _args.getMax().value();
+
     for (const auto& preImageDocKey : deferredReloadOrDeletePreImageDocKeys) {
         auto idElement = preImageDocKey["_id"];
         BSONObj newerVersionDoc;
         if (!Helpers::findById(opCtx, this->nss(), BSON("_id" << idElement), newerVersionDoc)) {
             // If the document can no longer be found, this means that another later op must have
-            // deleted it. That delete would have been captured by the xferMods so nothing else to
-            // do here.
+            // deleted it. If the document was moved out of the chunk by this transaction and then
+            // deleted, the later delete would not have been captured by the xferMods. We must 
+            // explicitly model a delete if the pre-image was in the chunk.
+            auto preImageShardKeyValues =
+                _shardKeyPattern.extractShardKeyFromDocumentKey(preImageDocKey);
+
+            if (isKeyInRange(preImageShardKeyValues, minKey, maxKey)) {
+                _addToTransferModsQueue(idElement.wrap(), 'd', {});
+            }
             continue;
         }
 


Abuhaithem:
image


what happend to my issue in jira @kelly-cs ?
I couldn't see it

kelly-cs:
Hey @Abuhaithem you should be able to see this one again. Our teams are taking a look still

renctan:
I don't think this is an issue today. This is because for the issue to happen, it needs the following condition:

1. Prepared transaction from previous term (so it will be sent to deferred xferMods) that modifies a document that was originally within the range of the chunk that is being moved.
2. Destination shard copied that document before deferred xferMod is processed (this will cause it to stay in the destination if deferred xferMod don't send a delete).
3. By the time xferMod is being processed, the document can no longer be accessed locally (either because it was moved out or deleted).

There are 2 things that saves this from happening:
1. [SERVER-71028](https://jira.mongodb.org/browse/SERVER-71028) - this means that cloner cannot fetch the document while document is in prepared. The cloner is basically blocked and waits for the prepared transaction to resolve.
2. If the transaction gets committed and the cloner gets unblocked, it can now copy the document to the destination. We can still potentially perform the modifications on the document so that it becomes out of range and gets deleted before the deferred xferMods is processed. However, the modifications will now under go the commit handler that is [not deferred](https://github.com/10gen/mongo/blob/f9726846e9c29e9378b192bd825354fdd400cc9c/src/mongo/db/s/migration_chunk_cloner_source.cpp#L273) and will generate a delete xferMods entry. So even if the processing of the defered xferMods skips it, the delete xferMods will be processed later.

I still think that the scenario is interesting, so I'll update our tests.

Abuhaithem:
> I don't think this is an issue today. This is because for the issue to happen, it needs the following condition:
> 
> 1. Prepared transaction from previous term (so it will be sent to deferred xferMods) that modifies a document that was originally within the range of the chunk that is being moved.
> 2. Destination shard copied that document before deferred xferMod is processed (this will cause it to stay in the destination if deferred xferMod don't send a delete).
> 3. By the time xferMod is being processed, the document can no longer be accessed locally (either because it was moved out or deleted).
> 
> There are 2 things that saves this from happening:
> 
> 1. [SERVER-71028](https://jira.mongodb.org/browse/SERVER-71028) - this means that cloner cannot fetch the document while document is in prepared. The cloner is basically blocked and waits for the prepared transaction to resolve.
> 2. If the transaction gets committed and the cloner gets unblocked, it can now copy the document to the destination. We can still potentially perform the modifications on the document so that it becomes out of range and gets deleted before the deferred xferMods is processed. However, the modifications will now under go the commit handler that is [not deferred](https://github.com/10gen/mongo/blob/f9726846e9c29e9378b192bd825354fdd400cc9c/src/mongo/db/s/migration_chunk_cloner_source.cpp#L273) and will generate a delete xferMods entry. So even if the processing of the defered xferMods skips it, the delete xferMods will be processed later.
> 
> I still think that the scenario is interesting, so I'll update our tests.

Agreed @renctan  ! with [SERVER-71028](https://jira.mongodb.org/browse/SERVER-71028) blocking the cloner and current-term ops always carrying post-image doc keys, I don't see a live path to the orphan either. But the correctness here depends on those two non-local invariants, and the existing comment states a guarantee ('the delete would have been captured') that isn't actually true ; the delete is range-filtered and only covered via the update's modeled delete. I'd suggest keeping the range-check-and-delete as cheap defense-in-depth (it's a no-op on the recipient when the doc was never transferred), or at minimum correcting the comment to document the real invariant. 

Happy to help with the test updates : )

renctan:
> Agreed @renctan ! with [SERVER-71028](https://jira.mongodb.org/browse/SERVER-71028) blocking the cloner and current-term ops always carrying post-image doc keys, I don't see a live path to the orphan either. But the correctness here depends on those two non-local invariants, and the existing comment states a guarantee ('the delete would have been captured') that isn't actually true ; the delete is range-filtered and only covered via the update's modeled delete. I'd suggest keeping the range-check-and-delete as cheap defense-in-depth (it's a no-op on the recipient when the doc was never transferred), or at minimum correcting the comment to document the real invariant.
> 
> Happy to help with the test updates : )

I updated the comment to clarify the implicit guarantee that it was relying on:

https://github.com/mongodb/mongo/commit/ad84999c41b73a7a548bdc30ac3772cc4d463525

I don't want to put the current changes in this patch because it is not a reachable path and so it cannot be tested.

